PR #35, opened 2015-09-14
docs: cleanup and fix a few bugs

robszumski:


Documentation/configure-kubectl.md
```diff
@@ -26,7 +26,7 @@ Configure your local Kubernetes client using the following commands:
 * Replace `${ADMIN_CERT}` with the path to the `admin.pem` created in previous steps
 
 ```sh
-$ kubectl config set-cluster vagrant --server=${MASTER_IP} --certificate-authority=${CA_CERT}
+$ kubectl config set-cluster vagrant --server=https://${MASTER_IP}:443 --certificate-authority=${CA_CERT}
```
robszumski: This is required to connect to the API server correctly

Documentation/deploy-master-single.md
```diff
@@ -112,7 +112,7 @@ spec:
     - --service-cluster-ip-range=${SERVICE_IP_RANGE}
     - --secure_port=443
     - --advertise-address=${ADVERTISE_IP}
-    - --admission-control=NamespaceLifecycle,NamespaceExists,LimitRanger,SecurityContextDeny,ServiceAccount,ResourceQuota
+    - --admission-control=NamespaceLifecycle,LimitRanger,SecurityContextDeny,ServiceAccount,ResourceQuota
```
robszumski: Previously, our API pod manifest wasn't accepted because the namespace didn't exist.

robszumski: cc: @bcwaldon @aaronlevy

polvi: http://kubernetes.io/v1.0/docs/admin/admission-controllers.html recommends we run with NamespaceExists. @kelseyhightower has previously suggested we turn it off. @kelseyhightower can you settle the debate?

bcwaldon: I would like to keep the NamespaceExists admission controller given it is suggested for all 1.0 deployments and GKE also uses it. 
i haven't had to create the kube-system namespace manually before, so there is likely something else going on if you're having issues creating the apiserver.

kelseyhightower: This one is mainly a bootstrapping problem. The kube-system namespace does not exist by default, unless something has changed. So we have a decision to make. If you bootstrap the api server using a pod under the kubelet in the kube-system namespace, then you need to handle the race. I think this issue came up because we are trying to deploy the api server in the kube-system namespace before it exists.

kelseyhightower: So either don't put the kube-apiserver in the kube-system namespace -- recommended so that the service for kube-apiserver exists in the default namespace. kube-system is mainly used for add-ons, which I don't think the apiserver falls in that group. So leave NamespaceExists in place and don't put the kube-apiserver in the kube-system namespace.

bcwaldon: The kubernetes service in the default namespace doesn't use a selector, so it can be in a different namespace than the kube-apiserver pods.

aaronlevy: This doesn't sound right. The apiserver is what would enforce this. I think something else was causing an issue here.

aaronlevy: Also, only sort of related, but all "master" components use the insecure port. Which bypasses all auth checks (and I'm pretty sure admission controller checks, but I'd have to trace code to make sure).

aaronlevy: Sorry, missed the slack discussion. But I just tested and it looks like it's failing on the creation of the mirror pods. We've just been lucky with timing in automated deployments (create the namespace before the mirror pods are being created).

So if we leave namespace as is, the instruction order might be: setup master, setup kubectl, create namespace, create worker. Or change it back to default.

aaronlevy:
@robszumski https://github.com/coreos/coreos-kubernetes/pull/40 adds instructions for adding the namespace. After adding the NamespaceExists admission controller back in, lgtm

PR #39, opened 2015-09-14
single-node: use 1024 MB of RAM

philips:
tectonic-manager is unable to deploy with only 512MB b/c it ran out of
memory. Fix by increasing RAM.


aaronlevy:
LGTM

bcwaldon:
LGTM - bonus points for pulling this out into a variable

bcwaldon:
@philips merging for ya

PR #40, opened 2015-09-14
Boot fixes

aaronlevy:


robszumski:
LGTM

robszumski:
But we should investigate the dropin ExecStartPre

PR #42, opened 2015-09-15
docs: flannel config from location that survives restart

aaronlevy:
/cc @bcwaldon @philips 

This feels kinda hackish. Before updating this in the rest of the vagrant scripts, any thoughts on this approach?

### DIFF
--- a/Documentation/deploy-master-single.md
+++ b/Documentation/deploy-master-single.md
@@ -20,16 +20,25 @@ Place the keys generated previously in the following locations:
 
 [flannel][flannel-docs] provides a key Kubernetes networking capability &mdash; a software-defined overlay network to manage routing of the [Pod][pod-overview] network.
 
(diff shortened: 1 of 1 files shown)


Documentation/deploy-master-single.md
```diff
@@ -20,18 +20,27 @@ Place the keys generated previously in the following locations:
...
+**/etc/flannel/options.env**
 
 ```sh
 FLANNELD_IFACE=${ADVERTISE_IP}
 FLANNELD_ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
 ```
 
+Next create a [systemd drop-in][dropins], which will use the above configuration when flannel starts
+
+**/etc/systemd/system/flanneld.service.d/40-options.conf**
```
bcwaldon: call this 40-ExecStartPre-options.conf

bcwaldon:
@aaronlevy LGTM - could you file a bug in coreos-overlay explaining why we have to do this?

bcwaldon:
@aaronlevy rebase required

PR #45, opened 2015-09-15
docs: kube-system namespace created prior to addon

aaronlevy:
kube-system namespace creation was moved into the deploy-master step. Will get an error here if we try and create it again.


robszumski:
LGTM

PR #47, opened 2015-09-15
Documentation: Fix IP addrs used in kubectl config

### LINKED ISSUE #46: kubernetes/docs: Incorrect IP used for example kubectl config
In the Kubernetes single node vagrant deploy docs, the examples show two different IP addresses (`172.17.4.101` & `172.17.4.201`). The Vagrantfile which is used to stand up the single node cluster uses a specific IP address (`172.17.4.99`) for the node, so the docs should reflect that. This could cause confusion for folks spinning up the single node cluster for the first time.

derekparker:
Fixes #46


robszumski:
LGTM

PR #51, opened 2015-09-16
*: remove unnecessary container port

bcwaldon:
This was likely added by accident, and we don't see it
being necessary for the cluster to operate correctly.


aaronlevy:
LGTM

PR #54, opened 2015-09-17
Documentation: clarify which dir to configure kubectl from

sym3tri:
Since the `cd foo` is in the previous section it was a bit unclear.


aaronlevy:
LGTM

PR #69, opened 2015-09-21
Fixed Kubernetes docs

kayrus:
* Moved up 'hostNetwork: true' inside kube-controller-manager yaml
* Added note for keys' permissions
* Added info about VIP addreses
* Added CoreOS release requirement into the top of README.md
* Updated curl api response to v1.0.6

/cc @robszumski

### DIFF
--- a/Documentation/configure-kubectl.md
+++ b/Documentation/configure-kubectl.md
@@ -21,9 +21,9 @@ $ mv kubectl /usr/local/bin/kubectl
 Configure your local Kubernetes client using the following commands:
 
 * Replace `${MASTER_HOST}`
-* Replace `${CA_CERT}` with the path to the `ca.pem` created in previous steps
-* Replace `${ADMIN_KEY}` with the path to the `admin-key.pem` created in previous steps
-* Replace `${ADMIN_CERT}` with the path to the `admin.pem` created in previous steps
+* Replace `${CA_CERT}` with the absolute path to the `ca.pem` created in previous steps
+* Replace `${ADMIN_KEY}` with the absolute path to the `admin-key.pem` created in previous steps
+* Replace `${ADMIN_CERT}` with the absolute path to the `admin.pem` created in previous steps
 
 ```sh
 $ kubectl config set-cluster vagrant --server=https://${MASTER_HOST} --certificate-authority=${CA_CERT}
--- a/Documentation/getting-started.md
+++ b/Documentation/getting-started.md
@@ -39,19 +39,19 @@ This network must be routable between all nodes in the cluster. In a default ins
 
 **SERVICE_IP_RANGE**=10.3.0.0/24
 
-The CIDR network to use for service cluster IPs. Each service will be assigned a cluster IP out of this range. This must not overlap with any IP ranges assigned to the POD_NETWORK, or other existing network infrastructure. Routing to these IPs is handled by a kube-proxy service local to each node, and are not required to be routable between nodes.
+The CIDR network to use for service cluster VIPs (Virtual IPs). Each service will be assigned a cluster IP out of this range. This must not overlap with any IP ranges assigned to the `POD_NETWORK`, or other existing network infrastructure. Routing to these VIPs is handled by a local kube-proxy service to each node, and are not required to be routable between nodes.
 
 <hr/>
 
 **K8S_SERVICE_IP**=10.3.0.1
 
-The IP address of the Kubernetes API Service. If the SERVICE_IP_RANGE is changed above, this must be set to the first IP in that range.
+The VIP (Virtual IP) address of the Kubernetes API Service. If the `SERVICE_IP_RANGE` is changed above, this must be set to the first IP in that range.
 
 <hr/>
 
 **DNS_SERVICE_IP**=10.3.0.10
 
-The IP address of the cluster DNS service. This IP must be in the range of the SERVICE_IP_RANGE and cannot be the first IP in the range. This same IP must be configured on all worker nodes to enable DNS service discovery.
+The VIP (Virtual IP) address of the cluster DNS service. This IP must be in the range of the `SERVICE_IP_RANGE` and cannot be the first IP in the range. This same IP must be configured on all worker nodes to enable DNS service discovery.
 
 ## Deploy etcd Cluster
 
@@ -65,7 +65,7 @@ If you are starting etcd manually, we need to first configure it to listen on al
 
 * Replace `${PUBLIC_IP}` with the etcd machines publicly routable IP address.
 
-** /etc/systemd/system/etcd2.service.d/40-listen-address.conf
+**/etc/systemd/system/etcd2.service.d/40-listen-address.conf**
 
 ```
 [Service]
--- a/Documentation/openssl.md
+++ b/Documentation/openssl.md
@@ -18,7 +18,7 @@ If you will be running a highly-available control-plane consisting of multiple m
 
 **K8S_SERVICE_IP**=10.3.0.1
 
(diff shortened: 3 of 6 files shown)


aaronlevy:
> Removed redundant etcd container from kube-dns pod

We do need to keep the etcd container. Right now worker nodes have access to control-plane etcd cluster due to flannel. However, ideally in the future this won't be a necessity (and already is not in some environments where flannel isn't used for the pod-network). When the DNS pod gets scheduled to a worker, we shouldn't require that it will have access to the control-plane etcd cluster.

> Fixed flannel env configuration for worker node (/run directory is not persistent)

Can you break this out into another PR we can get this merged as soon as possible (in case there are unrelated changes still needed in this PR). This was an oversight on my part, as it was fixed in docs & master already.

> Added instruction on how to download kubelet on earlier CoreOS versions

We made the decision *not* to encourage users to download individual binaries to the host. The reason being that it complicates issues around upgrades / consistency in docs / etc. An option that was floated is running certain tools inside of containers, but we haven't fully fleshed out that story yet.

aaronlevy:
I went ahead and fixed the flannel env as it was a bug: https://github.com/coreos/coreos-kubernetes/pull/71 but thank you for catching that!

kayrus:
/cc @aaronlevy @robszumski 
updated PR

Documentation/configure-kubectl.md
```diff
@@ -15,15 +15,15 @@ After downloading the binary, ensure it is executable and move it into your `PAT
 
 ```sh
 $ chmod +x kubectl
-$ mv kubectl /usr/local/bin/kubectl
+$ sudo mv kubectl /usr/local/bin/kubectl
```
aaronlevy: Can we drop the sudo? When it is specifically on a CoreOS machine, it's safer to make assumptions about underlying system (and when sudo is necessary). But in this case it's their local machine, and I'd rather leave this up to user.

Documentation/openssl.md
```diff
@@ -29,7 +29,7 @@ $ openssl genrsa -out ca-key.pem 2048
 $ openssl req -x509 -new -nodes -key ca-key.pem -days 10000 -out ca.pem -subj "/CN=kube-ca"
 ```
 
-You need to store the CA keypair in a secure location for future use.
+**It is highly recommended to generate and store keypairs in a secure location for future use, i.e. local machine.**
```
aaronlevy: Nit: drop "i.e. local machine"

README.md
```diff
@@ -8,6 +8,7 @@ It is officially maintained by the CoreOS team and meant to be a set of introduc
 
 When designing these guides and tools, the following considerations are made:
 
+* You must use a CoreOS version 773.1.0+ for the `kubelet` to be present in the image
```
aaronlevy: Can we drop this? This section was to call out the the motivations of the guides, not necessarily prerequisites. We've added the version requirement to the actual deploy docs.

aaronlevy:
One other change: Can you remove the UI-addon from this commit?

The initial goal of the guides is to demonstrate / document a fully-functioning Kubernetes cluster, of which the DNS addon is pretty much a requirement. However, other addons are (so far) somewhat optional. 

In this case it may be better to simply point to the upstream documentation, given that there is nothing particular to the CoreOS guide that needs to change to deploy the UI-addon.

This may change as upstream requirements / conventions change, but at this stage, the primary goal of the guides is to get a base Kubernetes cluster running, from which you can choose to deploy applications / addons.

Documentation/openssl.md
kayrus: IMHO, local machine is a good place to store private keys, especially CA private key.

README.md
kayrus: I've put version requirement at the beginning because I've faced situation when I have deployed stable cluster and tried to follow this doc. I've generated keys, set environment variables and at the middle of that doc it tells me that I have to install new CoreOS version. And I have to upgrade/reinstall my cluster. It is quite annoying.

kayrus:
Upstream documentation doesn't work sometimes. For example guestbook. It is hardcoded to work on GCE/AWS (because of loadbalancer port type) and it is not possible to access GUI until you define proper nodeport. I know that we have to contribute to kubernetes and fix that, but previously guestbook doc worked, now it doesn't, it seems someone has focused on cloud platforms only.

Removed UI addon. But why do we call this doc "Deploy Add-onS", not "Deploy DNS Add-on"?

Documentation/configure-kubectl.md
kayrus: would make sense. fixed.

Documentation/openssl.md
aaronlevy: I would probably argue the opposite, that the deployer's local machine is not the best place to store this data. Ultimately this is provided as a proof-of-concept, but I don't think we should be calling out the deployer's local machine as the "highly recommended secure location" for key material.

README.md
aaronlevy: Instead maybe move this to a "prerequisites" section at top of getting started guides then (https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/getting-started.md)? e.g. maybe prerequisites are: CoreOS version x.y.z, deployment options, generating tls certificates.

aaronlevy:
We could change the link to "Deploy Required Addons" if that makes it more clear. 

As far as the guestbook example, I opened an issue to document nodePort use: https://github.com/coreos/coreos-kubernetes/issues/75

Documentation/openssl.md
kayrus: Ok, can we just bold old text?

aaronlevy: Sure.

aaronlevy:
Also, need to move the deploy-addons.md file back to its original location. Couple reasons for this: The coreos.com docs would need to handle redirect links for this, so we're trying to minimize any file moves unless really necessary. Also, DNS is the only addon right now, but that could change - and if it does, and we move the file back, then we have more redirects.

Same thing for the link events, but I'll add a comment in-line

Documentation/configure-kubectl.md
```diff
@@ -42,5 +42,5 @@ X.X.X.X       kubernetes.io/hostname=X.X.X.X       Ready
 
 <div class="co-m-docs-next-step">
   <p><strong>Is kubectl working from your local machine?</strong> We're going to install an add-on with it next.</p>
-  <a href="deploy-addons.md" class="btn btn-primary btn-icon-right" data-category="Docs Next" data-event="Kubernetes: Addons">Yes, ready to deploy add-ons</a>
+  <a href="deploy-dns-add-on.md" class="btn btn-primary btn-icon-right" data-category="Docs Next" data-event="Kubernetes: DNS add-on">Yes, ready to deploy DNS add-on</a>
```
aaronlevy: File location and data-event need to stay the same. The link text is okay to change.

bcwaldon:
@kayrus were you able to get back to this?

kayrus:
Sorry, didn't have time yet.

kayrus:
@bcwaldon done

kayrus:
@bcwaldon any news?

Documentation/getting-started.md
```diff
@@ -2,6 +2,7 @@
 
 This guide will walk you through a deployment of a single-master/multi-worker Kubernetes cluster on CoreOS. We're going to:
 
+- You must use a CoreOS release 773.1.0+ for the `kubelet` to be present in the image
```
aaronlevy: Nit: This seems out of place. The heading is "We're going to", followed by "You must use...".

This also isn't true of etcd dedicated etcd nodes (can be run from stable). We note the required version in both the master installation and the worker installation steps. Can we just drop this line for now to get this merged (unless you have another placement suggestion)?

kayrus:
Removed. There is almost nothing left from my initial commit (

aaronlevy:
These changes are helpful though! LGTM

PR #77, opened 2015-09-23
Added http proxy support

ndobson:
Added http proxy support via the vagrant-proxyconf plugin

### DIFF
--- a/Documentation/kubernetes-on-vagrant-single.md
+++ b/Documentation/kubernetes-on-vagrant-single.md
@@ -12,6 +12,16 @@ Navigate to the [Vagrant downloads page][vagrant-downloads] and grab the appropr
 
 [vagrant-downloads]: https://www.vagrantup.com/downloads.html
 
+##### Proxy Support
+
+For http proxy support, install the vagrant-proxyconf plugin.
+
+```sh
+vagrant plugin install vagrant-proxyconf
+```
+
+The provisioning scripts will detect if an HTTP_PROXY environment variable is set and will configure appropriately.
+
 ### kubectl
 
 The primary CLI tool used to interact with the Kubernetes API is called `kubectl`.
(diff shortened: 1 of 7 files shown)


aaronlevy:
I'm not familiar with this plugin, can you describe what functionality you need that this provides?

ndobson:
@aaronlevy the vagrant-proxyconf plugin enables vagrant to automatically set proxy environment variables in /etc/enivonment and /etc/profile. 

For those of us lucky enough to have to work behind a corporate http proxy, this with my other changes enables us to pull down docker images from the internet via an http proxy.

aaronlevy:
Ideally we minimize the amount of use-case specific configuration that is in the base bootstrap scripts.

One option would be for the initialization scripts to source a pre-defined "user-templates" file that is ignored in base repo (but allows user-configuration to live alongside). Similar to the `config.rb.sample` for user-defined Vagrant overrides.

Do you think something like that would work for your use-case?

ndobson:
I suppose so, but i think it would be nice to have http_proxy support baked in as many enterprise users have to deal with it. If you look at the way I've implemented this, it only does something new if:

1.  the vagrant-proxyconf plugin is installed
2.  there is an http_proxy veriable set in the shell environment.

Otherwise nothing new happens.

aasimid:
I am also one of those unlucky ones behind a corporate proxy and the funny thing is that I implemented the same concept before stumbling into this (albeit with a lot more hardcoded values).
 The concept actually follows the guideline suggested here: https://coreos.com/os/docs/latest/customizing-docker.html

One comment I have is that in my case you cannot even download the plugin (as you are already behind the proxy) via:

$vagrant plugin install vagrant-proxyconf

Instead the plugin has to be downloaded manually and then installed like this:

$sudo vagrant plugin install vagrant-proxyconf-1.5.2.gem

ndobson:
@aasimid You should be able to install vagrant plugins from behind a proxy by setting your `http_proxy` and `https_proxy` environment variables.

aasimid:
@ndobson Thanks you are right. I was pretty sure I had them set, so I had another go on a fresh machine just to try this out and it's like you say; it turned out I set them at some point after installing the plugin.

bcwaldon:
1. Is there a less-invasive way we can support this? Having to maintain this feature across several scripts is asking for it to break.
2. How can we solve this problem for *all* traffic, rather than just docker?

aaronlevy:
Closing due to inactivity. Please let us know if this should be reopened.

PR #80, opened 2015-09-30
Multi-node AWS Installer

bcwaldon:
This tool is intended to make it incredibly easy to deploy a full-featured Kubernetes cluster on AWS. Since we have not yet cut a release of this, you will have to set `artifactURL: https://coreos-kubernetes.s3.amazonaws.com/latest` in your cluster config file for this to work.

I broke apart the Godeps commit for reviewability here.


aaronlevy:
The `multi-node/aws/README.md` docs will likely need to move under Documentation so they get synced properly. /cc @robszumski

silas:
We should change `"InstanceTenancy":    "dedicated",` to `"InstanceTenancy":    "default",`, otherwise we'll add an additional $2/hour for any user that doesn't already have this enabled.

* https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/dedicated-instance.html
* https://aws.amazon.com/ec2/purchasing-options/dedicated-instances/

robszumski:
Here's my thinking for how the docs for this fit together.

Separate this out into 2 docs:
  - keep everything under "Development" in the readme. link to guides on coreos.com similiar to the [overall readme](https://github.com/coreos/coreos-kubernetes/tree/64de7c3d85ed3731e99b7378e14aaf8d90d1c33f#deploying-on-coreos)
  - move the actual guide into `Documentation/kubernetes-on-aws.md` so it gets synced and versioned

Adding some more stuff inline.

robszumski:
Should we call out the config options with some more detail, similar to https://coreos.com/kubernetes/docs/latest/getting-started.html#deployment-options?

bcwaldon:
@robszumski :+1: on breaking apart the docs like you suggested. I will document each template parameter, too.

aaronlevy:
One minor nit, but otherwise LGTM

bcwaldon:
thanks @aaronlevy and @robszumski!

PR #85, opened 2015-10-02
docs: fix headings

robszumski:


bcwaldon:
@robszumski thanks, will stop at H4 next time :)

PR #93, opened 2015-10-03
Require confirmation for destroying a cluster

jimmycuadra:
Help prevent accidental cluster destruction by requiring either interactive confirmation or --force=true.

### DIFF
--- a/multi-node/aws/cmd/kube-aws/command_destroy.go
+++ b/multi-node/aws/cmd/kube-aws/command_destroy.go
@@ -17,10 +17,13 @@ var (
 		Long:  ``,
 		Run:   runCmdDestroy,
 	}
+
+	force bool
 )
 
 func init() {
 	cmdRoot.AddCommand(cmdDestroy)
+	cmdDestroy.Flags().BoolVar(&force, "force", false, "Destroy the cluster without interactive confirmation")
 }
 
 func runCmdDestroy(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


xied75:
:+1:

multi-node/aws/cmd/kube-aws/command_destroy.go
```diff
@@ -33,6 +36,20 @@ func runCmdDestroy(cmd *cobra.Command, args []string) {
 
 	c := cluster.New(cfg, newAWSConfig(cfg))
 
+	if !force {
+		var confirmation string
+
+		fmt.Print("Are you sure you want to destroy the cluster? This action cannot be undone. Type \"yes\" to proceed: ")
```
bcwaldon: Simplify this to `Are you sure you want to permanently destroy the cluster? [y/N] `

multi-node/aws/cmd/kube-aws/command_destroy.go
```diff
@@ -33,6 +36,20 @@ func runCmdDestroy(cmd *cobra.Command, args []string) {
...
+		if _, err = fmt.Scanln(&confirmation); err != nil {
+			stderr("Failed to read user input: %v", err)
+			os.Exit(1)
+		}
+
+		if confirmation != "yes" {
```
bcwaldon: Look for a "y" or "Y", typing "yes" is annoying

bcwaldon:
Agree w/ this conceptually, I just want to see the UX improved a bit.

multi-node/aws/cmd/kube-aws/command_destroy.go
jimmycuadra: My reason for spelling it out completely is to make it slightly safer by making it less likely you'll just quickly type "y<return>" without thinking. While you may create and destroy clusters a lot while using this as a development tool, if kube-aws is intended to be used to manage production clusters, I'd want the program to help protect me from accidentally destroying a cluster.

That said, I'm happy to use y/Y if that's really what you'd prefer.

aaronlevy:
@jimmycuadra there has been a major refactoring of the kube-aws tool. If this is still a feature you would like to see, would you be able to re-apply these changes on top of the new codebase? Alternatively, I can close this PR and open an issue to track the feature request.

jimmycuadra:
I don't actually use this tool, so I will just close it. Thanks!

PR #96, opened 2015-10-04
pkg/cluster: mark AZ choice as optional

robszumski:
When using the AWS CloudFormation interface, the Availability Zone field is not marked optional.

> If not set, a zone will be chosen for you in the region being used to deploy the CloudFormation stack.


aaronlevy:
lgtm

PR #101, opened 2015-10-07
multi-node/aws: increase default volume size

silas:
The current template defaults to 8 GiB root volumes with ~5 GiB of usable space with system overhead.


bcwaldon:
@silas can you make sure this is configurable via cluster.yaml, and documented in the example?

silas:
Updated

bcwaldon:
@silas only nit remaining is to call it workerRootVolumeSize everywhere, just to be more specific. LGTM otherwise

Issue #73, opened 2015-09-22
kubectl exec - worker cert issue


ndobson:
When using the OpenSSL documentation running kubectl exec produces the following error:
```
kubectl --insecure-skip-tls-verify=false exec busybox -- nslookup kubernetes
error: Unable to upgrade connection: {
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "x509: certificate is valid for kube-worker, not coreos03",
  "code": 500
}
```
As an insecure workaround I generated the worker csr like so:
```
openssl req -new -key worker-key.pem -out worker.csr -subj "/CN=*"
```

hunter:
We're running into a similar issue with `exec`

```
error: Unable to upgrade connection: {
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "x509: cannot validate certificate for 172.31.129.X because it doesn't contain any IP SANs",
  "code": 500
}
```

polvi:
We are planning to upgrade the doc to show the `openssl req -new -key worker-key.pem -out worker.csr -subj "/CN=*"` work-around as well as document the "right" way to do it by generating a key for every worker. @hunter @ndobson does that sound like a reasonable approach to you? Any other ideas?

ndobson:
@polvi It could, but also note that the `subj "/CN=*"` approach only works if you change the `--hostname-override=${ADVERTISE_IP}` option from the kubelet of each worker and let the kubelet register itself as a resolvable hostname.

Otherwise, as @hunter demonstrates above, if you try to connect to the worker by ip the cert has to have an IP SAN entry matching that specific ip.

polvi:
Another solution could be to use ssh. I'm pretty sure the apiserver can connect over SSH, but that would require setting up SSH infrastructure as well as x509. Any thoughts on that?

bcwaldon:
This is handled by the AWS installer (multi-node/aws) by adding *.ec2.internal and *.<region>.cluster.internal SANs

ntquyen:
@ndobson I'm having the same problem and have little experience in ssl. Can you please give me more instruction on generating new worker csr? I'm using multi-node for development.

hosh:
[...]

PR #102, opened 2015-10-07
[ssl] include IP SANS information for worker containers in the worker…

colhom:
… ssl key. allows for exec'ing into worker containers

Fixes the X509 cert validation problem when calling "kubectl exec" into a pod, where the worker is accessed by a different network address than the controller (default case).

$> kubectl exec <worker_pod_name>  -ti -- sh 

error: Unable to upgrade connection: {
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "x509: cannot validate certificate for 172.17.4.201 because it doesn't contain any IP SANs",
  "code": 500
}

### DIFF
--- a/multi-node/vagrant/.gitignore
+++ b/multi-node/vagrant/.gitignore
@@ -1,3 +1,5 @@
 .vagrant
 config.rb
 ssl/
+.dockercfg
+
--- a/multi-node/vagrant/Vagrantfile
+++ b/multi-node/vagrant/Vagrantfile
@@ -49,8 +49,9 @@ etcd_endpoints = etcdIPs.map.with_index{ |ip, i| "http://#{ip}:2379" }.join(",")
 
 # Configure SSL certificates
 if !File.exist?(SSL_CONTROLLER_PATH) || !File.exists?(SSL_WORKER_PATH) then
-    sans = "IP.1=10.3.0.1," + controllerIPs.map.with_index { |ip, i| "IP.#{i+2}=#{ip}"}.join(",")
-    system("mkdir -p ssl && ./init-ssl ssl #{sans}") or abort ("failed generating SSL artifacts")
+  controller_sans = "IP.1=10.3.0.1," + controllerIPs.map.with_index { |ip, i| "IP.#{i+2}=#{ip}"}.join(",")
+  worker_sans = "IP.1=10.3.0.1," + workerIPs.map.with_index { |ip, i| "IP.#{i+2}=#{ip}"}.join(",")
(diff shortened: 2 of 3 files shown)


aaronlevy:
Couple general discussion points:

The ideal approach is to have a cert per worker. Adding a bunch of SANs is a little less than ideal (and also won't give us actual node identity, as that comes from the CN). We should also be smarter in our ssl asset generation to not just check for a tarball - but instead that the cert for that node has actually been generated (otherwise changing the worker count on an existing cluster then doing `vagrant up` will not work because the new node IP will not be part of the cert SANs).

So in general, this change allows exec to work, but I'm hesitant to go further down this route. Just to put out another option, we could use ssh keys to allow for the exec/proxy path (but same issue that we still don't get real node identity). See: https://github.com/kubernetes/kubernetes/blob/v1.0.6/cmd/kube-apiserver/app/server.go#L208-L209. 

Also, it seems like the dockercfg change is somewhat unrelated to this?

Related: https://github.com/coreos/coreos-kubernetes/issues/73

/cc @bcwaldon

colhom:
So, I agree with your concerns. Going forward, all we're really accomplishing with this approach is getting exec to work.. without any real progress towards ensuring node identity.

I'm pretty new to k8s, so excuse any unforeseen stupidity in the following proposition.

How about (as vagrant provisioning steps)
```
  - Each node get's a unique+random subdomain (eg. worker-rx642y.cluster.local )

  - Vagrant expects to be provided with a root CA for use within the cluster, 
     Anything talking to the cluster will use this CA to validate 
     the nodes' ssl certs (via --root-ca-file)

  - for all existing nodes
       * nuke any existing certs
       * use the internal root CA to generate a valid cert with 
          the Common Name set to the internal DNS address for 
          that node (eg. worker-rx642y.cluster.local ) 
```
WRT the dockercfg changes... I forgot how pull requests work. I pushed that to master without realizing it would stack on top of my existing pull request. The dark days of github noobdom

~Colin

aaronlevy:
What would be managing the "internal DNS" records for those nodes?

I think we can go with the cert per-worker approach without too much changing:

In init-ssl:
- If CA does not exist
   - Delete all existing ssl assets
   - Create CA keypair
- If "server" cert does not exist:
    - Create keypair with SANS=control node IPs
- If "worker" cert for each worker does not exist:
    - Create keypair for node with CN=IP

In Vagrantfile
- Transfer server keypair+CAcert to all control nodes
- Transfer each worker keypair+CAcert only to the applicable node

This way all certs can generally stay static, but if a new worker node is launched, we generate a keypair for that node, and it is copied only to that node. Little more tedious from Vagrantfile perspective, because you're not just checking for existence of a tarball, but instead individual certs.

PR #105, opened 2015-10-09
Clarified how to modify default config

tedyoung:
As a non-Vagrant expert, I found it confusing to have the config.rb.sample file to be completely commented out, when it's not included by default anyway.

### DIFF
--- a/Documentation/kubernetes-on-vagrant.md
+++ b/Documentation/kubernetes-on-vagrant.md
@@ -41,22 +41,25 @@ $ cd coreos-kubernetes/multi-node/vagrant
 
 ## Start the Machines
 
-The default cluster configuration is to start a virtual machine for each role &mdash; controller, worker, and etcd server. However, you can modify the default cluster settings by copying `config.rb.sample` to `config.rb` and modifying configuration values.
+The default cluster configuration is to start a virtual machine for each role &mdash; controller, worker, and etcd server. However, you can modify the default cluster settings by copying `config.rb.sample` to `config.rb` and modifying the configuration values, e.g.:
 
+```ruby
+$update_channel="beta"
+
+# Leave defaults for the controller
+# $controller_count=1
+# $controller_vm_memory=512
+
+# More workers, each with more memory
+$worker_count=3
+$worker_vm_memory=1024
+
+# More etcd instances, and increase memory for each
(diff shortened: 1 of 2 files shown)


aaronlevy:
In general, I think it would be better to keep the values provided in `config.rb.sample` match the actual defaults. Copying config.rb.sample to config.rb should essentially be a no-op unless specific values are modified (same idea for leaving all values commented).

Also, general stylistic nit:  remove extra spaces after `#` in commented out commands (makes actual diff more concise too).

Documentation/kubernetes-on-vagrant.md
```diff
@@ -41,22 +41,25 @@ $ cd coreos-kubernetes/multi-node/vagrant
...
 
-#$etcd_count=1
-#$etcd_vm_memory=512
+# More etcd instances, and increase memory for each
+$etcd_count=3
+$etcd_vm_memory=1024
 ```
 
-Next, simply run `vagrant up` and wait for the command to succeed.
+Next, run `vagrant up` and wait for the command to succeed. (If you get `No usable default provider could be found for your system.`, then you need to install/specify a provider as described [here](https://docs.vagrantup.com/v2/getting-started/providers.html)).
```
aaronlevy: I don't think we should assume / document this type of failure case here. This is generally outside the scope of this documentation (configuring vagrant).

multi-node/vagrant/config.rb.sample
```diff
@@ -1,10 +1,26 @@
...
+# configuration options (global ruby variables) specified below. To override the
+# defaults that can be found in the Vagrantfile, copy this file to "config.rb", then
+# for each option that you want to change, uncomment the line (i.e., remove the
+# leading `#`) leaving just the `$`-prefixed variable name, and replace everything
+# after the equals sign with your desired setting.
+#
 
-#$controller_count=1
-#$controller_vm_memory=512
+# Official CoreOS channel from which updates should be downloaded: "alpha", "beta", "stable"
```
aaronlevy: Might be good to also mention minimum version (773.1.0+ ) here too (first version containing the kubelet binary)

aaronlevy:
Closing due to inactivity. Please reopen if still necessary.

PR #108, opened 2015-10-16
switch to 172.20.0.0/14 for networking

dysinger:
10.0.0.0/14 is really really popular for folks to already have in their infrastructure.  If one wanted to vpc peer to the kubernetes cluster, it would be difficult with 10.0.0.0/14. 172.20.0.0/14 is probably less popular & more likely to just work.  

Creating this patch shows that we need a way to have these network IP addresses & CIDRs as cluster.yaml settings & not hard coded into go & misc files.  I will make sure there's an issue for this concern.


dysinger:
http://blog.thestateofme.com/2014/05/27/beware_the_default_network/

silas:
+1

Although longer term we should enable the AWS cloudprovider routes feature (https://github.com/kubernetes/kubernetes/blob/071d21257fdfd439a9286fffbd5972c85643cd49/pkg/cloudprovider/providers/aws/aws_routes.go) and make the subnets fully configurable.

bcwaldon:
I don't want to change the defaults until all fields are configurable. This needs to land after https://github.com/coreos/coreos-kubernetes/pull/190

phemmer:
I would not recommend changing the default address anyway. At least not until kubernetes properly supports it. Right now the 10.0.0.0/8 subnet is hard-coded in some parts of kubernetes: https://github.com/kubernetes/kubernetes/blob/v1.1.2/pkg/kubelet/container_bridge.go#L127

bcwaldon:
We are going to make this configurable for folks that don't need to rely on the portions of Kubernetes hard-coded to 10.0.0.0/8: https://github.com/coreos/coreos-kubernetes/pull/190. Please interface with the upstream community to address the issues preventing us from using a range outside of 10.0.0.0/8 if you believe it is important.

PR #113, opened 2015-10-17
flannel: use vxlan backend

aaronlevy:
Fixes: https://github.com/coreos/coreos-kubernetes/issues/112

Ran k8s conformance tests. All worked well.
Leaving the aws installation as-is for now (using udp) because we are switching to use the cloud-provider/vpc integration soon anyway.

/cc @philips @eyakubovich

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -361,7 +361,7 @@ Earlier it was mentioned that flannel stores cluster-level configuration in etcd
 * Replace `$ETCD_SERVER` with one host from `$ETCD_ENDPOINTS`
 
 ```sh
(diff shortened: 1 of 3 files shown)


eyakubovich:
Do any docs have to be updated regarding the port number that should be opened? udp backend uses 8285 while vxlan uses 8472.

aaronlevy:
For the vagrant install, and the step-by-step guide (which I missed, but just added in new commit) we don't document specific port use.

The aws guide does setup firewall rules, but leaving those using udp until can swap out flannel for vpc integration.

bcwaldon:
LGTM, bonus points for documenting the port requirement in the generic multi-node guide :)

aaronlevy:
Half points for tracking it in an issue? https://github.com/coreos/coreos-kubernetes/issues/116

bcwaldon:
LGTM :)

PR #117, opened 2015-10-21
Kubernetes on OS X Apps

rimusz:


aaronlevy:
Hey @rimusz, it was great to meat you at Kubecon! I haven't had a chance to give these projects a spin yet, but hopefully soon. But I don't think we're going to be linking to outside projects right now. Also it looks like your project is linked from upstream docs already: http://kubernetes.io/v1.1/docs/getting-started-guides/coreos.html

rimusz:
@aaronlevy it was nice meeting you to there. @philips did ask about to link kube-solo App from coreos-kubernetes

PR #121, opened 2015-10-22
[ssl] individual SSL certs for workers. Each worker cert's IP SANS is…

colhom:
… provisioned only for that worker's IP address. Api server nodes get unique ssl certificates provisioned for ALL controller ip addresses plus the cluster ip. Node common names are of the form kube-{role}-{node_ip}.


colhom:
@aaronlevy pretty pretty please

aaronlevy:
Couple nits:
-break up commit message / see CONTRIBUTING doc.
-we also have an init-ssl script int he single-node dir. Can we combine these to just use your new version?

Otherwise, LGTM

colhom:
Should be good to go. Would have liked to break up the first commit in two, but the consolidation and worker cert SANS happened in parallel

PR #125, opened 2015-10-25
multi-node/aws/artifacts: Add node labels for AWS metadata.

### LINKED ISSUE #120: Tag nodes in AWS after kubelet registration with instance-type
That way if you have a cluster with mixed node types you can target node types using `nodeSelector`.  This would require changes to `kubelet.service` unit installed by `multi-node/aws/artifacts/scripts/install-worker.sh`.

Similarly, it would be a good idea to automatically add tags for other EC2 metadata such as `placement/availability-zone` and `ami-id`.  The tags could be namespaces under `as.amazon.com`.  E.g. `aws.amazon.com:instance-type`.

eliaslevy:
This demonstrates a proposal on how to tag nodes with labels derived from AWS metadata.  It makes use of a small single-shot container dropped into the /etc/kubernetes/metadata directory to fetch the metadata and apply the label.

The container I use can be found in Docker Hub as elevy/aws-node-labels.  CoreOS would copy and host the container for it to be officially integrated.

Note that this is just a stopgap until the https://github.com/kubernetes/kubernetes/pull/13524 proposal is implemented.

Fixes #120.

### DIFF
--- a/multi-node/aws/artifacts/manifests/worker/aws-node-labels.yaml
+++ b/multi-node/aws/artifacts/manifests/worker/aws-node-labels.yaml
@@ -1,0 +1,23 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: aws-node-labels
+spec:
+  hostNework: true
+  restartPolicy: OnFailure
+  containers:
+    - name: apply-labels
+      image: elevy/aws-node-labels:latest
+      securityContext:
+        privileged: true
+      env:
+        - name: CONTROLLER_ENDPOINT
+          value: "${CONTROLLER_ENDPOINT}"
+      volumeMounts:
+        - mountPath: /etc/kubernetes/ssl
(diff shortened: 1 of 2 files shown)


bcwaldon:
@eliaslevy thanks for demonstrating how to do this. I don't immediately see the value in carrying this, especially if a fix is coming upstream.

eliaslevy:
Its unclear when that proposal will drop, or even if it will be accepted.  But as you wish.

hosh:
:+1: I'm using AWS and have been looking around for different approaches.

aaronlevy:
I also don't feel like this is something we should be carrying. I think that the choice to run a side-car labelling pod should be left up to the user, as this may not be something all AWS users want. Generally the motivation of this repo is to launch the base CoreOS+Kubernetes cluster, from which users can then work from.

PR #126, opened 2015-10-26
Update deploy-master.md to fix broken manifest path.

drocsid:
kublet deploys pods from the config path /etc/kubernetes/manifests, and not /srv/kubernetes/manifests. I was not able to have a functioning cluster until setting the path properly. This was quite annoying.


aaronlevy:
The /srv/kubernetes/manifests path is intentional for the controller-manager and scheduler. In a multi-node cluster, these are managed by the "podmaster".

Each podmaster will master elect for each component. If it is the master then it will move the manifest file from /srv/kubernetes/manifests to /etc/kubernetes/manifests. This ensures that only a single copy of controller-manager / scheduler are running on the cluster.

See: http://kubernetes.io/v1.0/docs/admin/high-availability.html#master-elected-components

PR #132, opened 2015-11-02
Configure kubectl keys with apiserver

spiddy:
admin.pem and admin-key.pem no longer exists. The name seems to have changed to apiserver.pem and apiserver-key.pem


aaronlevy:
Thanks for catching this. It's been fixed in https://github.com/coreos/coreos-kubernetes/pull/133

PR #135, opened 2015-11-03
[multi-node/aws docs] update s3 artifact URL docs to recognize new s3…

colhom:
… endpoint format

Requesting https://s3.amazonaws.com/bucket/path you get an xml blob telling you
to use the url https://bucket.s3.amazonaws.com/path instead


bcwaldon:
LGTM

colhom:
ping

PR #138, opened 2015-11-08
Instantiate a VPN access server with the Kubernetes cluster

### LINKED ISSUE #137: Secure AWS Deployments with a VPN
Currently the AWS deployment allows incoming connections from the Internet to the controller on ports 22 and 443 and to the works on port 22.  It would be nice if there was the option to set up a VPN along with Kubernetes, and lock down the security group ingress rules to only allow connections from within the VPC.

[Here](https://gist.github.com/eliaslevy/bda1a3ae2588661e7121) you can find a CloudFormation template that instantiates an OpenVPN access server on an existing VPC that already has an Internet gateway.  It could be integrated into the Kubernetes CF template while tightening down the ingress security groups.

After that, one only needs to SSH to the VPN instance as `openvpnas`, change the password for the `openvpn` user, visit `https://[elastic_ip]:943/admin` to change any configuration, visit `https://[elastic_ip]` to download the VPN profile, import the profile into your VPN client (e.g. Tunnelblick), and connect to the VPN.

eliaslevy:
It also disables the controller elastic IP and SSH and HTTPS access to cluster nodes.

To configure the VPN access server after its brought up by the CloudFormation template:
- ssh openvpnas@[elastic_ip]
- passwd openvpn
- if desired, login to https://[elastic_ip]:943/admin with the openvpn user to config the VPN
- login to https://[elastic_ip], select login, not connect, from the drop down and download VPN connection profile
- import the profile into Tunnelblick or other VPN client by opening the file
- connect using your username (openvpn) and password

Closes #137.

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -12,8 +12,8 @@ import (
 )
 
 type ClusterInfo struct {
-	Name         string
-	ControllerIP string
+	Name  string
+	VpnIP string
 }
 
 func (c *ClusterInfo) String() string {
@@ -22,7 +22,8 @@ func (c *ClusterInfo) String() string {
 	w.Init(buf, 0, 8, 0, '\t', 0)
 
 	fmt.Fprintf(w, "Cluster Name:\t%s\n", c.Name)
-	fmt.Fprintf(w, "Controller IP:\t%s\n", c.ControllerIP)
+	fmt.Fprintf(w, "Controller IP:\t10.0.0.50\n")
+	fmt.Fprintf(w, "VPN IP:\t%s\n", c.VpnIP)
 
 	w.Flush()
 	return buf.String()
--- a/multi-node/aws/pkg/cluster/stack.go
+++ b/multi-node/aws/pkg/cluster/stack.go
@@ -78,14 +78,9 @@ func mapStackResourcesToClusterInfo(svc *ec2.EC2, resources []cloudformation.Sta
 	var info ClusterInfo
 	for _, r := range resources {
 		switch aws.StringValue(r.LogicalResourceId) {
-		case resNameEIPController:
-			if r.PhysicalResourceId != nil {
-				info.ControllerIP = *r.PhysicalResourceId
-			} else {
-				return nil, fmt.Errorf("unable to get public IP of controller instance")
-			}
 		}
 	}
+	return nil, fmt.Errorf("unable to get public IP of controller instance")
 
 	return &info, nil
 }
--- a/multi-node/aws/pkg/cluster/stack.go
+++ b/multi-node/aws/pkg/cluster/stack.go
@@ -78,9 +78,14 @@ func mapStackResourcesToClusterInfo(svc *ec2.EC2, resources []cloudformation.Sta
 	var info ClusterInfo
 	for _, r := range resources {
 		switch aws.StringValue(r.LogicalResourceId) {
+		case resNameEipVpn:
+			if r.PhysicalResourceId != nil {
+				info.ControllerIP = *r.PhysicalResourceId
+			} else {
+				return nil, fmt.Errorf("unable to get public IP of VPN instance")
+			}
 		}
 	}
-	return nil, fmt.Errorf("unable to get public IP of controller instance")
 
 	return &info, nil
 }
--- a/multi-node/aws/pkg/cluster/stack_template.go
+++ b/multi-node/aws/pkg/cluster/stack_template.go
@@ -385,7 +385,7 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
 			"NetworkInterfaces": []map[string]interface{}{
 				map[string]interface{}{
 					"PrivateIpAddress":         "10.0.0.50",
-					"AssociatePublicIpAddress": false,
+					"AssociatePublicIpAddress": true,
 					"DeleteOnTermination":      true,
 					"DeviceIndex":              "0",
 					"SubnetId":                 newRef(resNameSubnet),
--- a/multi-node/aws/pkg/cluster/stack_template.go
+++ b/multi-node/aws/pkg/cluster/stack_template.go
@@ -48,6 +48,8 @@ var (
 	supportedChannels    = []string{"alpha"}
 	tagKubernetesCluster = "KubernetesCluster"
 
+	vpcCidr = "10.0.0.0/16"
+
 	sgProtoTCP = "tcp"
 	sgProtoUDP = "udp"
 
@@ -141,7 +143,7 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
 	res[resNameVPC] = map[string]interface{}{
 		"Type": "AWS::EC2::VPC",
 		"Properties": map[string]interface{}{
-			"CidrBlock":          "10.0.0.0/16",
+			"CidrBlock":          vpcCidr,
 			"EnableDnsSupport":   true,
 			"EnableDnsHostnames": true,
 			"InstanceTenancy":    "default",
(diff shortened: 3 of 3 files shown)


phemmer:
In my opinion, this hurts, not helps, security.
Currently, the system is secured by SSH public key, which is a very secure authentication mechanism. This change introduces a weaker authentication scheme: username/password.
To make matters worse, this username/password are the same for all deployments. So when your stack comes up, anyone can log into it (until you get a chance to change the password).
Also, this now requires me to install a VPN client.

If the concern is exposing the controller to the outside world, a jump host would be a better solution in my opinion. Just a vanilla instance, of any OS, that lives within the VPC, but accepts connections from the outside world.

eliaslevy:
I think you misunderstand OpenVPN.  The client is authenticated not solely on a username and password, but also on a client certificate.  In addition, you can enable OTP second factor on top of client certificate authentication.

You can configure the VPN instance security group so as to not expose the web UI, accessing it via a forwarded connection via SSH for configuration purposes.  You can set the admin password to whatever you want via the CloudFormation user data field for the VPN instance.

The concern is not simply exposing the controller, but any other services you may stand up that you need to manage, including ones hosted in private subnets.  A SSH bastion host is all well and good, but it not as seamless as a VPN connection.  With Kubernetes AWS native networking (#141), that also means being able to seamlessly connect to the Pod network space if you configure OpenVPN correctly.  That means bypassing the buggy apiserver HTTP proxy and being able to connect non-HTTP services in Pods.

phemmer:
> I think you misunderstand OpenVPN. The client is authenticated not solely on a username and password.

Not true, you even went over this in your intitial description:

> if desired, login to https://[elastic_ip]:943/admin with the openvpn user
...
> login to https://[elastic_ip]

&nbsp;

> You can configure the VPN instance security group so as to not expose the web UI, accessing it via a forwarded connection via SSH for configuration purposes

You don't need a VPN server to do this. You can do the same for the coreos-kubernetes security groups right now.

&nbsp;

> With Kubernetes AWS native networking (#141), that also means being able to seamlessly connect to the Pod network space if you configure OpenVPN correctly

AWS offers other features which can do this, such as [VPC peering](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-peering.html) and [VPC endpoints](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-endpoints.html).
This solution forces me to spin up things I may not need or want.

eliaslevy:
Again, you can close off the web UI from all external access by modifying the VPN instance security group.  Then you'll have to use a forwarded connection via SSH to access the web UI.

The actual VPN connection is secured by **both** a client key, which is embedded in profile you download and import into your VPN client, and the user name and password. It can be additional secured by a one time password on top of the client secret if you so desire.

AWS VPC Peering is about connecting VPC networks. AWS VPC Endpoints is about allowing your VPC instances to access other AWS resources (although its limited to only S3 so far).  A VPN access server is about allowing an external *client* easy access to the VPC network.  Completely different use cases.

phemmer:
> Again, you can close off the web UI from all external access by modifying the VPN instance security group. Then you'll have to use a forwarded connection via SSH to access the web UI.

And you can do this with the current deployment as well. Adding a VPN does not grant anything new here.

> The actual VPN connection is secured by both a client key,
It doesn't matter. Your system is only as secure as its weakest link.

> AWS VPC Peering is about connecting VPC networks

Correct, such as connecting to an existing VPC which you already have a VPN for, or some other form of access.

> AWS VPC Endpoints is about allowing your VPC instances to access other AWS resources

Not true. A quote from the document I linked:
> A VPC endpoint enables you to create a private connection between your VPC and another AWS service without requiring access over the Internet, through a NAT instance, a **VPN connection**, or **AWS Direct Connect**

Note that it allows configuring a VPN, as well as AWS direct connect, which is a private link to a VPC.

eliaslevy:
On Thu, Nov 12, 2015 at 9:43 AM, Patrick Hemmer <notifications@github.com>
wrote:

> And you can do this with the current deployment as well. Adding a VPN does
> not grant anything new here.
>
>
It grants you network access to the VPC without the use of SSH forwarded
connections or having to expose services to the Internet.  I am not sure
why you find that difficult to understand.  If you have a service running
within k8s that you'd like to access, say Grafana, the VPN access server
will allow you access it easily.



> The actual VPN connection is secured by both a client key,
> It doesn't matter. Your system is only as secure as its weakest link.
>
> Which is not the password, since you can block the web UI and the VPN
connection is protected by a client key **in addition**to the client.


> AWS VPC Peering is about connecting VPC networks
>
> Correct, such as connecting to an existing VPC which you already have a
> VPN for, or some other form of access.
>
The whole point of the VPN access server instance is to provide that "other
form of access".


> AWS VPC Endpoints is about allowing your VPC instances to access other AWS
> resources
>
> Not true. A quote from the document I linked:
>
> A VPC endpoint enables you to create a private connection between your VPC
> and another AWS service without requiring access over the Internet, through
> a NAT instance, a *VPN connection*, or *AWS Direct Connect*
>
> Note that it allows configuring a VPN, as well as AWS direct connect,
> which is a private link to a VPC.
>

It says it allows you to connect to another AWS service **without**
requiring a VPN connection.  It does not say that it allows you connect to
your VPC with a VPN connection.

I am well aware a AWS Direct Connect is a private link to a VPC. I have a
AWS Direct Connect connection between my data center and my AWS VPC.  Once
again, the VPN access server instance is not about connection your network
to the VPC (site-to-site). It is about connecting your client computer to
the VPC (endpoint-to-site), for ease of access from a *client*.

What you are probably trying to refer to is to an AWS [Virtual Private
Gateway](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_VPN.html),
which is actually a hardware VPN concentrator.  The AWS VPG is about
setting up a site-to-site VPN tunnel.  It requires you to modify your VPC
routing table to route your traffic back through the VPN connection, or
your VPN access device must speak BGP to tell the VPG what routes to add.

By contrast, the VPN access server instance is about allow access to the
VPC from *endpoints*.  It requires no modification of the VPC routing table.

I suggest you read [Amazon VPC Connectivity Options](
https://media.amazonwebservices.com/AWS_Amazon_VPC_Connectivity_Options.pdf),
pages 26-31.

phemmer:
> If you have a service running
within k8s that you'd like to access, say Grafana, the VPN access server
will allow you access it easily.

Nowhere did I say a VPN does not let you have easy access to non-public addresses.
You seem to keep emphasizing that the ability to adjust the security is limited to a VPN server, when it is not.

> The whole point of the VPN access server instance is to provide that "other
form of access".

It is not the responsibility of coreos to provide alternate forms of access to the cluster. This is the responsibility of the user.

> It says it allows you to connect to another AWS service **without**
requiring a VPN connection.  It does not say that it allows you connect to
your VPC with a VPN connection.
You are correct. I cited the wrong reference.

> Once again, the VPN access server instance is not about connection your network to the VPC (site-to-site). It is about connecting your client computer to the VPC (endpoint-to-site), for ease of access from a *client*.

...which you can do using a things such as VPC peering & AWS Direct Connect, and other solutions. Site-to-site means that it allows traffic from separate networks to access each other. If your personal machines sits on, or routes through, one of those networks, you can utilize the site-to-site link.

The bottom line is that this PR reduces system security, and is forcing your personal solution onto others who don't want or need it.
The cloudformation template should provide the basics for building things on top of. Not provide things that users have to rip out and re-engineer.

eliaslevy:
> The specifics are also largely irrelevant. 

LOL.

> The point I was trying to make is there are other solutions available.

That is true.

> It is not the responsibility of coreos to provide alternate forms of access to the cluster. This is the responsibility of the user.

It's not the responsibility of CoreOS to provide anything.  They have kindly made their work on Kubernetes and CoreOS available.  Their work exposes the nodes using SSH and the k8s controller using HTTPS.  I choose to lock them down while simultaneously making accessing other services in the cluster from an endpoint easier to access for development and management purposes using a VPN access server.  Like them, I am kindly making my work available to others. CoreOS is not required to merge my changes.  They can decide to close the pull request if they so choose.

> The bottom line is that this PR reduces system security

That is false, but at this point I think we are past arguing about it, as "specifics are largely irrelevant."

> and is forcing your personal solution onto others who don't want or need it.

No one is forcing anyone to do anything.  CoreOS can choose not to merge the pull request.  I assure you, I won't be angry.  The pull request is merely a suggestion that works for me and I think would be useful to others.   On the other hand, you seem rather invested in this.

Maybe its time for bed?

aaronlevy:
I do think adding a VPN machine is outside the intended scope of this project, as we can't make assumptions about how a user will want to interact with their cluster (nor should we launch additional resources based on those assumptions). That being said, it's great to see these use-cases being explored.

PR #139, opened 2015-11-11
docs: fix kubelet arguments

brenix:
Cluster arguments for kubelet should use a dash instead of underscore. See https://github.com/kubernetes/kubernetes/blob/master/docs/admin/kubelet.md#options


aaronlevy:
Both should work, but I'm all for consistency. LGTM

PR #142, opened 2015-11-12
multi-node/aws: Upgrade to K8s 1.1.1 and configure kube-proxy to use iptables

eliaslevy:
The kubelet 1.1.1 binary is downloaded and used instead of the CoreOS kubelet to maintain version parity with other components.  To be reversed once CoreOS ships a v1.1.1 kubelet.

### DIFF
--- a/multi-node/aws/artifacts/scripts/install-controller.sh
+++ b/multi-node/aws/artifacts/scripts/install-controller.sh
@@ -5,7 +5,7 @@ set -e
 export ETCD_ENDPOINTS=
 
 # Specify the version (vX.Y.Z) of Kubernetes assets to deploy
-export K8S_VER=v1.0.7
+export K8S_VER=v1.1.1
 
 # The CIDR network to use for pod IPs.
 # Each pod launched in the cluster will be assigned an IP out of this range.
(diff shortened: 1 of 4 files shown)


multi-node/aws/artifacts/scripts/install-controller.sh
```diff
@@ -64,6 +64,13 @@ function init_config {
 	done
 }
 
+function install_kubelet {
+	mkdir /opt
+	cd /opt
+	curl -sLO https://storage.googleapis.com/kubernetes-release/release/v1.1.1/bin/linux/amd64/kubelet
```
iterion: We should sub `v1.1.1` in this line with ${K8S_VER} so it doesn't need updated with each version change.

eliaslevy: Good point.

aaronlevy:
Appreciate the quick turnaround on this, but we're still working through testing some upgrade issues and not quite ready to bump the version.

I am going to close this for now in favor of tracking the upgrade process for docs+aws+vagrant here: https://github.com/coreos/coreos-kubernetes/pull/144.

PR #144, opened 2015-11-13
K8S v1.1.1 upgrade

aaronlevy:


SpencerBrown:
hello Aaron, I have a task to upgrade our Kubernetes cluster to 1.1.1, how ready to go is this PR in case I want to use coreos-kubernetes?

aaronlevy:
Hi Spencer,

The upgrade path is one of the reasons for holding off on merging this, but if you're creating a new cluster, this code as-is is pretty safe with a couple caveats:

The Kubelet that is currently shipped in the CoreOS image is v1.0.7 -- so using this code will upgrade your master components, but the kubelet will not yet be updated (unless manually replaced on the host). This should be fine, as "master version >= kubelet version" is a supported upgrade path in k8s upstream.

We are still running our own internal testing of the new version (and version skews of kubelet / master components), and the main reason this has not yet merged.

philips:
LGTM. We are following the recommended upgrade path and will figure out the kubelet situation in CoreOS in a separate discussion.

See https://github.com/kubernetes/kubernetes/issues/17184 and https://github.com/kubernetes/kubernetes/issues/16961

Issue #82, opened 2015-10-01
Expose cadvisor on public interface


aaronlevy:
The kubelet cadvisor interface is currently only listening on localhost (via `--cadvisor-port=0`). We may want to expose cadvisor so metric collection tools (e.g. heapster addon) can retrieve data from the nodes.

PR #145, opened 2015-11-14
Added Addon Kube-UI to multi-node aws artifacts

ankitforcode:
1. Added Replication Controller for Kube-UI
2. Added Service for Kube-UI
3. Added Addon - Kube-UI to be launched on Controller node with 'kube-aws up' script also changed the cadvisor port from 0 to 4194
4. Changed the cadvisor port from 0 to 4194 to be used with Kube-UI to get the stats from worker nodes
5. Added Port TCP port 4194 for cAdvisor when using KubeUI in render function for kube-aw

### DIFF
--- a/multi-node/aws/artifacts/manifests/cluster/kube-ui-rc.json
+++ b/multi-node/aws/artifacts/manifests/cluster/kube-ui-rc.json
@@ -1,0 +1,59 @@
+{
+  "apiVersion": "v1",
+  "kind": "ReplicationController",
+  "metadata": {
+    "labels": {
+      "k8s-app": "kube-ui",
+      "kubernetes.io/cluster-service": "true",
+      "version": "v3"
+    },
+    "name": "kube-ui-v3",
+    "namespace": "kube-system"
+  },
+  "spec": {
+    "replicas": 1,
+    "selector": {
+      "k8s-app": "kube-ui",
+      "version": "v3"
+    },
+    "template": {
+      "metadata": {
+        "labels": {
+          "k8s-app": "kube-ui",
+          "kubernetes.io/cluster-service": "true",
+          "version": "v3"
+        }
+      },
+      "spec": {
+            "containers":[
+               {
+                  "name":"kube-ui",
+                  "image":"gcr.io/google_containers/kube-ui:v3",
+                  "resources": {
+                    "limits": {
+                      "cpu": "100m",
+                      "memory": "50Mi"
+                     }
+                   },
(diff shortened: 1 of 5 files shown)


ankitforcode:
@aaronlevy Can you please review

aaronlevy:
Hi @ankitforcode,
Would it be possible to move the cadvisor change into a separate PR? This is something that I've wondered about, and I think it makes sense to use the default behavior (could even just remove the `--cadvisor-port` flag altogether). ref: https://github.com/coreos/coreos-kubernetes/issues/82

As far as the Kube-UI addon, the motivation for this project thus far has been to provide a baseline installation of kubernetes from which users can build on. The DNS addon is part of the initial deployment because it is essentially required for a fully-functioning cluster. But other addons are optional.

In general, the Kubernetes installations based on this repo should pass the upstream conformance tests (another reason the cadvisor port should probably be left at default), but anything running in the cluster beyond that I think should be left up to the user to decide (e.g. UI, Heapster, etc.)

Maybe instead we can add a documentation section for addons which point to the upstream docs / manifests?

ankitforcode:
@aaronlevy Agree.

ankitforcode:
@aaronlevy  I strongly believe that Kube-UI should be the part of the base installation since it provides the base info for the cluster and stats for the worker nodes. As, I can already see there are lots of issues posted on users trying to get the Kube-UI working.

aaronlevy:
I am still leaning toward adding documentation for optional addons, rather than making this decision for the user. For example, we could add something along the lines of:

```
# Deploy Kube-UI addon
> git clone https://github.com/kubernetes/kubernetes
> cd kubernetes && git checkout v1.1.1
> kubectl create -f kubernetes/cluster/addons/kube-ui/.

# Authenticate & expose Kube-UI via proxy:
> kubectl proxy
Starting to serve on 127.0.0.1:8001

# Load kube-UI in browser at: http://127.0.0.1:8001/ui
```

Any thoughts on this @bcwaldon @silas ?

bcwaldon:
I'm strongly for only supporting what is necessary to get a cluster running, then providing documentation & examples for additional components. I don't believe the Kube-UI is required as kubectl provides the same information, albeit via a CLI.

aaronlevy:
Opened an issue for creating documentation for deploying the kube-ui: https://github.com/coreos/coreos-kubernetes/issues/163

Closing this in favor of documenting the steps to deploy kube-ui addon.

PR #160, opened 2015-11-23
k8s: bump to version 1.1.2

aaronlevy:


philips:
lgtm

robszumski:
Can we get a release + notes? Anything else that has changed since 2 weeks ago?

aaronlevy:
I was going to try and tag a release today. There's been some bug fixes
I'll include in the changelog too.

ddispaltro:
This doesn't actually work, the kubelet version you use is from `/usr/bin/kubelet` which may not match up, version wise.  You'll probably need to run hyperkube for the kubelets

aaronlevy:
@ddispaltro you are correct that this will not update the kubelet that ships with the OS. This will update the the master components to v1.1.2, but the underlying kubelet may be of a different version. Having the master components >= kubelet version is, however, supported.

If you want to also update the kubelet out-of-band from the OS, you could put an updated binary on the host and change the kubelet.service unit to use it.

Also see: https://github.com/coreos/bugs/issues/978#issuecomment-161391489 for some background on updating the kubelet in the OS image. We will be shipping a newer kubelet in an alpha release soon, but eventually the kubelet will likely be run via a signed image/container

PR #162, opened 2015-11-23
vagrant: add kubeconfig files to support KUBECONFIG path and vagrant-…

colhom:
…{single,multi} clusters

### DIFF
--- a/Documentation/kubernetes-on-vagrant-single.md
+++ b/Documentation/kubernetes-on-vagrant-single.md
@@ -53,12 +53,21 @@ Once Vagrant is finished booting and provisioning your machine, your cluster is 
 
 Once in the `coreos-kubernetes/single-node/` directory, configure your local Kubernetes client using the following commands:
 
-```sh
-$ kubectl config set-cluster vagrant --server=https://172.17.4.99:443 --certificate-authority=${PWD}/ssl/ca.pem
-$ kubectl config set-credentials vagrant-admin --certificate-authority=${PWD}/ssl/ca.pem --client-key=${PWD}/ssl/admin-key.pem --client-certificate=${PWD}/ssl/admin.pem
-$ kubectl config set-context vagrant --cluster=vagrant --user=vagrant-admin
-$ kubectl config use-context vagrant
-```
+You can choose from one of the two following options.
+
+1. **Use a custom KUBECONFIG path**
+   ```sh
+   export KUBECONFIG="${KUBECONFIG}:$(pwd)/kubeconfig"
+   kubectl config use-context vagrant-single
+   ```
+
+2. **Update the local-user kubeconfig**
+   ```sh
+   $ kubectl config set-cluster vagrant-single-cluster --server=https://172.17.4.99:443 --certificate-authority=${PWD}/ssl/ca.pem
+   $ kubectl config set-credentials vagrant-single-admin --certificate-authority=${PWD}/ssl/ca.pem --client-key=${PWD}/ssl/admin-key.pem --client-certificate=${PWD}/ssl/admin.pem
+   $ kubectl config set-context vagrant-single --cluster=vagrant-single-cluster --user=vagrant-single-admin
+   $ kubectl config use-context vagrant-single
+   ```
 
 Check that your client is configured properly by using `kubectl` to inspect your cluster:
 
--- a/Documentation/kubernetes-on-vagrant.md
+++ b/Documentation/kubernetes-on-vagrant.md
@@ -64,15 +64,25 @@ Next, simply run `vagrant up` and wait for the command to succeed.
 Once Vagrant is finished booting and provisioning your machine, your cluster is good to go.
 
 ## Configure kubectl
+You can choose from one of the two following options.
 
-Configure your local Kubernetes client using the following commands:
+1. **Use a custom KUBECONFIG path**
 
-```sh
-$ kubectl config set-cluster vagrant --server=https://172.17.4.101:443 --certificate-authority=${PWD}/ssl/ca.pem
-$ kubectl config set-credentials vagrant-admin --certificate-authority=${PWD}/ssl/ca.pem --client-key=${PWD}/ssl/admin-key.pem --client-certificate=${PWD}/ssl/admin.pem
-$ kubectl config set-context vagrant --cluster=vagrant --user=vagrant-admin
-$ kubectl config use-context vagrant
-```
+	```sh
+	export KUBECONFIG="${KUBECONFIG}:$(pwd)/kubeconfig"
+	kubectl config use-context vagrant-multi
+	```
+
+2. **Update the local-user kubeconfig**
+
+	Configure your local Kubernetes client using the following commands:
+
+	```sh
+	$ kubectl config set-cluster vagrant-multi-cluster --server=https://172.17.4.101:443 --certificate-authority=${PWD}/ssl/ca.pem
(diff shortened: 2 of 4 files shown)


colhom:
@aaronlevy ping.

aaronlevy:
How will you be using this / can we add some documentation on how we expect these to be used?

We currently document creating these via `kubectl config set-foo` commands: https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/kubernetes-on-vagrant-single.md#configure-kubectl

This change seems useful, but may be worth describing how they can be used vs using the commands to modify existing config.

colhom:
@aaronlevy docs added. also i have both the clusters named "vagrant" to be consistent with the `kubectl config set-*` instructions- the assumption is people will be using one or the other?

Documentation/kubernetes-on-vagrant.md
```diff
@@ -60,6 +60,16 @@ Next, simply run `vagrant up` and wait for the command to succeed.
 Once Vagrant is finished booting and provisioning your machine, your cluster is good to go.
 
 ## Configure kubectl
+You can choose from one of the two following options.
+
+1. **KUBECONFIG path**
```
aaronlevy: nit: "Use a custom KUBECONFIG path"

Documentation/kubernetes-on-vagrant.md
```diff
@@ -60,6 +60,16 @@ Next, simply run `vagrant up` and wait for the command to succeed.
...
+
+```sh
+export KUBECONFIG="${KUBECONFIG}:$(pwd)/kubeconfig"
+kubectl config use-context vagrant
+```
+
+2. **kubectl config**
```
aaronlevy: nit: "Update the local-user kubeconfig"

aaronlevy: nit: this doesn't render as "2." -- you will need to indent the code block above.

aaronlevy:
I actually get annoyed with the name being "vagrant" for both single/multi-node configs because I often find myself switching between. In the case of the custom KUBECONFIG path, it doesn't matter because there is only one user/cluster/context per file.

I'd like to see it change to something like "vagrant-multi" in the case of updating existing kubeconfig, so the settings don't stomp on eachother. But that can be left to a different PR

Documentation/kubernetes-on-vagrant-single.md
```diff
@@ -49,6 +49,15 @@ Once Vagrant is finished booting and provisioning your machine, your cluster is
 
 Once in the `coreos-kubernetes/single-node/` directory, configure your local Kubernetes client using the following commands:
 
+You can choose from one of the two following options.
```
bcwaldon: There are really three options:

1. export `KUBECONFIG`
2. run the appropriate `kubectl config` commands
3. run kubectl with the `--kubeconfig` flag

Why do we need to document all of these methods?

bcwaldon:
I agree w/ aaron regarding differentiation of multi- and single-node kubeconfigs. Please do that in this PR.

bcwaldon:
@colhom Were you able to make the requested changes?

colhom:
@bcwaldon changes made

Documentation/kubernetes-on-vagrant-single.md
```diff
@@ -49,12 +49,21 @@ Once Vagrant is finished booting and provisioning your machine, your cluster is
...
+   ```sh
+   export KUBECONFIG="${KUBECONFIG}:$(pwd)/kubeconfig"
+   kubectl config use-context vagrant-single
+   ```
+
+2. **Update the local-user kubeconfig**
+   ```sh
+   $ kubectl config set-cluster vagrant-single --server=https://172.17.4.99:443 --certificate-authority=${PWD}/ssl/ca.pem
+   $ kubectl config set-credentials vagrant-admin --certificate-authority=${PWD}/ssl/ca.pem --client-key=${PWD}/ssl/admin-key.pem --client-certificate=${PWD}/ssl/admin.pem
+   $ kubectl config set-context vagrant-single --cluster=vagrant --user=vagrant-admin
```
aaronlevy: the `--cluster=vagrant` doesn't match the `set-cluster vagrant-single` above.

Documentation/kubernetes-on-vagrant.md
```diff
@@ -60,15 +60,25 @@ Next, simply run `vagrant up` and wait for the command to succeed.
...
+	kubectl config use-context vagrant-multi
+	```
+
+2. **Update the local-user kubeconfig**
+
+	Configure your local Kubernetes client using the following commands:
+
+	```sh
+	$ kubectl config set-cluster vagrant-multi --server=https://172.17.4.101:443 --certificate-authority=${PWD}/ssl/ca.pem
+	$ kubectl config set-credentials vagrant-admin --certificate-authority=${PWD}/ssl/ca.pem --client-key=${PWD}/ssl/admin-key.pem --client-certificate=${PWD}/ssl/admin.pem
```
aaronlevy: This will overwrite the `vagrant-admin` set in the single node cluster if someone has already followed those steps. If the namespacing is meant to make this easier to switch between (e.g. `kubect use-context vagrant-[single|multi]`) then we need to namespace this too: `vagrant-multi-admin`

Documentation/kubernetes-on-vagrant.md
```diff
@@ -60,15 +60,25 @@ Next, simply run `vagrant up` and wait for the command to succeed.
...
+	$ kubectl config set-context vagrant-multi --cluster=vagrant --user=vagrant-admin
```
aaronlevy: `--cluster=vagrant` doesn't mat the `set-cluster vagrant-multi` from above

multi-node/vagrant/kubeconfig
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Config
+clusters:
+- cluster:
+    certificate-authority: ssl/ca.pem
+    server: https://172.17.4.101:443
+  name: vagrant-cluster
```
aaronlevy: Just for consistency sake, can we keep same names as given in the local-user examples. In this case "vagrant-multi"

multi-node/vagrant/kubeconfig
```diff
@@ -0,0 +1,18 @@
...
+contexts:
+- context:
+    cluster: vagrant-cluster
+    namespace: default
+    user: vagrant-admin
+  name: vagrant-multi
+users:
+- name: vagrant-admin
```
aaronlevy: Even though this is a static config and it wouldn't exactly matter like it would with modifying existing user config, can we namespace these like the local-user examples too: `vagrant-multi-admin`

single-node/kubeconfig
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Config
+clusters:
+- cluster:
+    certificate-authority: ssl/ca.pem
+    server: https://172.17.4.99:443
+  name: vagrant-cluster
+contexts:
+- context:
+    cluster: vagrant-cluster
```
aaronlevy: vagrant, or vagrant-single, or vagrant-cluster, or vagrant-single-cluster. Don't feel strongly but keep it consistent between all examples.

aaronlevy:
Made some comments RE consistency. Also once the changes are made, just sanity check that you are able to switch between contexts as expected and have it work for single/multi node clusters:

```
kubectl config use-context vagrant-single; kubectl get nodes
kubectl config use-context vagrant-multi; kubectl get nodes
KUBECONFIG=path/to/single kubectl get nodes
KUBECONFIG=path/to/multi kubectl get nodes
```

colhom:
rebased/squashed. tested KUBECONFIG and kubectl approaches, everything checks out.

aaronlevy:
lgtm

PR #165, opened 2015-11-25
Add notes on building `kube-aws`

dhoeric:
docs: update building `kube-aws`

The current readme is not enough to build the `kube-aws`, need a much complete notes to build `kube-aws`.

### DIFF
--- a/multi-node/aws/README.md
+++ b/multi-node/aws/README.md
@@ -11,6 +11,8 @@ Run the `./build` script to compile `kube-aws` locally.
 This depends on having golang available on your workstation.
 The compiled binary will be available at `./bin/kube-aws`.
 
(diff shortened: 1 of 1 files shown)


aaronlevy:
Hi @dhoeric,

Can you give some more detail on what issue you were having that this resolves? The build script explicitly sets the Godeps dir as part of the GOPATH: https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/build#L20

dhoeric:
@aaronlevy The intention of the commit is to fix the ordering problem on 
https://github.com/coreos/coreos-kubernetes/blob/acc15da705982a58b0a7f84049387100ca3f1d23/multi-node/aws/build#L20 which has been fixed.
Therefore this request can be closed.

Issue #5, opened 2015-08-31
Secure kubelets with TLS


bcwaldon:
All kubelets should be deployed with unique client TLS certificates signed by a shared CA.

aaronlevy:
Last piece (step-by-step guide) resolved in https://github.com/coreos/coreos-kubernetes/pull/170

PR #170, opened 2015-11-27
docs: create unique TLS certs per worker

rwehner:
Using per-worker TLS certificates/keys allows the SubjectAltNames to be
accurate for each worker node and increases security over
shared-certificate solutions like setting `\CN=*` and is more portable
to different cloud providers and on-prem installations than solutions
like using `*.ec2.internal` in the SAN. Sadly, doing this also adds
installation complexity. Having correct SAN settings on the workers allows
`kubectl exec` to work instead of giving TLS errors.

Fixes documentation part of
[#73](https://github.com/coreos/coreos-kubernetes/issues/73)

### DIFF
--- a/Documentation/deploy-workers.md
+++ b/Documentation/deploy-workers.md
@@ -1,27 +1,36 @@
-## Deploy Worker Node(s)
+# Deploy Worker Node(s)
 
 Boot one or more CoreOS nodes which will be used as Kubernetes Workers. You must use a CoreOS version 773.1.0+ for the `kubelet` to be present in the image.
 
 See the [CoreOS Documentation](https://coreos.com/os/docs/latest/) for guides on launching nodes on supported platforms.
 
-### Configure Service Components
+## Configure Service Components
 
-#### TLS Assets
+### TLS Assets
 
-Place the TLS keypairs generated previously in the following locations:
+Place the TLS keypairs generated previously in the following locations. Note that each keypair is unique and should be installed on the worker node it was generated for:
 
 * File: `/etc/kubernetes/ssl/ca.pem`
-* File: `/etc/kubernetes/ssl/worker.pem`
-* File: `/etc/kubernetes/ssl/worker-key.pem`
+* File: `/etc/kubernetes/ssl/${WORKER_FQDN}-worker.pem`
+* File: `/etc/kubernetes/ssl/${WORKER_FQDN}-worker-key.pem`
 
 And make sure you've set proper permission for private key:
 
-```
+```sh
 $ sudo chmod 600 /etc/kubernetes/ssl/*-key.pem
 $ sudo chown root:root /etc/kubernetes/ssl/*-key.pem
 ```
 
-#### flannel Configuration
+Create symlinks to the worker-specific certificate and key so that the remaining configurations on the workers do not have to be unique per worker.
+
+```sh
+$ cd /etc/kubernetes/ssl/
+$ sudo ln -s ${WORKER_FQDN}-worker.pem worker.pem
+$ sudo ln -s ${WORKER_FQDN}-worker-key.pem worker-key.pem
+```
+
+
+### flannel Configuration
 
 *Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
 
@@ -48,7 +57,7 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 
 [dropins]: https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html
 
-#### Docker Configuration
+### Docker Configuration
 
 *Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
 
@@ -64,7 +73,7 @@ Requires=flanneld.service
 After=flanneld.service
 ```
 
-#### Create the kubelet Unit
+### Create the kubelet Unit
 
 Create `/etc/systemd/system/kubelet.service` and substitute the following variables:
 
@@ -94,7 +103,7 @@ RestartSec=10
 WantedBy=multi-user.target
 ```
 
-#### Set Up the kube-proxy Pod
+### Set Up the kube-proxy Pod
 
 Create `/etc/kubernetes/manifests/kube-proxy.yaml`:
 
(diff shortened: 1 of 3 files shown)


Documentation/deploy-workers.md
```diff
@@ -8,19 +8,28 @@ See the [CoreOS Documentation](https://coreos.com/os/docs/latest/) for guides on
 
 #### TLS Assets
 
-Place the TLS keypairs generated previously in the following locations:
+Place the TLS keypairs generated previously in the following locations. Note that each keypair is unique and should be installed on the worker node is was generated for.:
```
aaronlevy: s/is/it

Nit: drop the trailing . or :, but don't have both

Documentation/openssl.md
```diff
@@ -20,6 +20,14 @@ If you will be running a highly-available control-plane consisting of multiple m
 
 The IP address of the Kubernetes API Service. The `K8S_SERVICE_IP` will be the first IP in the `SERVICE_IP_RANGE` discussed in the [deployment guide][deployment-guide]. The first IP in the default range of 10.3.0.0/24 will be 10.3.0.1. If the SERVICE_IP_RANGE was changed from the default, this value must be updated as well.
 
+<hr/>
```
aaronlevy: I know this is just an addition here because it was moved, but we should try and use markdown instead (e.g. `---`)

Documentation/openssl.md
```diff
@@ -20,6 +20,14 @@ If you will be running a highly-available control-plane consisting of multiple m
...
+
+**WORKER_IP**=_no default_
+
+**WORKER_FQDN**=_no default_
+
+The IP addresses and fully qualifed hostnames of all worker nodes will be needed. In most cases these will be the main routable IP/hostname of the worker machines. These will be unique per worker; when you see them used below, consider it a loop and do that step for _each_ worker.
```
aaronlevy: Workers may not have a routable hostname. We may want to change this wording to reflect that it is technically optional.

Maybe something along the lines of: "The certificates generated for the worker nodes will need to reflect how requests will be routed to those nodes. In most cases this will be a routable IP and/or a routable hostname. These will be unique per worker..."

Documentation/openssl.md
```diff
@@ -78,12 +86,42 @@ $ openssl req -new -key apiserver-key.pem -out apiserver.csr -subj "/CN=kube-api
...
+distinguished_name = req_distinguished_name
+[req_distinguished_name]
+[ v3_req ]
+basicConstraints = CA:FALSE
+keyUsage = nonRepudiation, digitalSignature, keyEncipherment
+subjectAltName = @alt_names
+[alt_names]
+DNS.1 = ${WORKER_FQDN}
+IP.1 = 127.0.0.1
+IP.2 = ${WORKER_IP}
```
aaronlevy: couple things:
* If there is a routable hostname, it should technically be the CN rather than a DNS altname. Also, because this is optional, we may want to describe in the section above using a placeholder for WORKER_FQDN (e.g. "kube-worker")
* I feel like we should drop the 127.0.0.1. Generally you will expose insecure api on localhost:8080 anyway.

I also think we can actually make the cnf more generic using env vars. If we change `IP.1 = $ENV::WORKER_IP` we can have a single cnf file and describe command below as:

`WORKER_IP=<replace> openssl req -new -key ${WORKER_FQDN}-worker-key.pem -out ${WORKER_FQDN}-worker.csr -subj "/CN=$WORKER_FQDN" -config worker-openssl.cnf`

aaronlevy:
Hi @rwehner thank you for this PR! Super helpful.

Couple minor comments, and I think depending on the worker openssl.cnf stuff some of the wording might need to be updated (if there is no more per-worker cnf file to describe). Let me know if you think that works.

/cc @robszumski @joshix (for doc changes)

rwehner:
@aaronlevy Thanks for the detailed comments. They all make sense to me. Hopefully, I've got most of them addressed with this commit. Happy to keep tweaking if it is not quite there. Thanks for the review.

aaronlevy:
LGTM after one last minor change. Then can you also squash these into a single commit?

aaronlevy:
ping @robszumski @joshix (docs changes)

aaronlevy:
Fixes: https://github.com/coreos/coreos-kubernetes/issues/5

We will have unique (or wildcard for AWS) certs for clients in vagrant / docs / aws after this lands.

rwehner:
Sounds good, @aaronlevy. Do you prefer that I remove the `<replace>` for both [`WORKER_FQDN=<replace>`](https://github.com/coreos/coreos-kubernetes/pull/170/files#diff-e1eed7996cb6b8e12f51b37a8b4c7d54R117) and [`WORKER_IP=<replace>`](https://github.com/coreos/coreos-kubernetes/pull/170/files#diff-e1eed7996cb6b8e12f51b37a8b4c7d54R119) or just the `WORKER_IP` instances?

I'll get it updated and rebased, either way.

joshix:
This LGTM (where G is "great!"), with the conformist nit that it should be adjusted to match the heading hierarchy we like, described in https://github.com/joshix/docs/blob/styledir-hdings/STYLE.md#headings.

More than anything, document titles should be level 1 headings (`# Heading`)

aaronlevy:
@rwehner I think you can drop this line: `WORKER_FQDN=<replace>` as the env var is already being used directly on the command line.

rwehner:
The headings have been addressed (I think) and the `<replace>` stuff fixed. All rebased.

@aaronlevy @joshix Let me know if I missed anything. Thanks!

joshix:
@rwehner Headings look perfect, thanks for fiddling with that detail to help us keep things consistent.

LGTM for merge.

aaronlevy:
One last thing. I just stepped through these instructions, and if workers have no fqdn, we're suggesting using "kube-worker". However, used verbatim this will just stomp on previously generated keypairs with the same name.

Possible suggestion, instead of:

> If the node does not have a routeable hostname, set `WORKER_FQDN=kube-worker` as a generic placeholder.

Maybe:

> If the node does not have a routable hostname, use a different generic placeholder for each worker node. E.g. `WORKER_FQDN=kube-worker-1`

joshix:
@aaronlevy Good catch. I left it to you to pull the merge lever precisely because I haven't done the procedure.

suggest to your suggest: `s/. E.g./, like /`

rwehner:
Yeah, great catch, @aaronlevy. I tested with hostnames only. bad me. How does this look? I even got the 'like' instead of 'e.g.' in there @joshix. <grin>

joshix:
+1

aaronlevy:
lgtm

PR #171, opened 2015-11-28
adding Dockerfile for building a runnable image, updating README and …

danielwhatmuff:
Hit some annoying config problems with my local Go installation when trying to build the binary. Thought it would be nice to make a little docker build for people to get started more easily.

Have tested the changed locally - so should be good to go.

Ta

### DIFF
--- a/multi-node/aws/cmd/kube-aws/command_up.go
+++ b/multi-node/aws/cmd/kube-aws/command_up.go
@@ -128,17 +128,21 @@ func mustReadFile(loc string) []byte {
 
 func newKubeconfig(cfg *cluster.Config, tlsConfig *cluster.TLSConfig) ([]byte, error) {
 	data := struct {
-		ClusterName       string
-		APIServerEndpoint string
-		AdminCertFile     string
-		AdminKeyFile      string
-		CACertFile        string
+		ClusterName               string
+		ControllerInstanceType    string
+		WorkerInstanceType        string
+		APIServerEndpoint         string
+		AdminCertFile             string
+		AdminKeyFile              string
+		CACertFile                string
 	}{
-		ClusterName:       cfg.ClusterName,
-		APIServerEndpoint: fmt.Sprintf("https://%s", cfg.ExternalDNSName),
-		AdminCertFile:     tlsConfig.AdminCertFile,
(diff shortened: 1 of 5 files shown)


danielwhatmuff:
Also, hit an issue with the Worker and Controller Instance Type parameters. They dont seem to be referenced and so the default is used even if they are specified in the cluster.yaml.

multi-node/aws/cmd/kube-aws/command_up.go
```diff
@@ -128,17 +128,21 @@ func mustReadFile(loc string) []byte {
...
+		AdminKeyFile              string
+		CACertFile                string
 	}{
-		ClusterName:       cfg.ClusterName,
-		APIServerEndpoint: fmt.Sprintf("https://%s", cfg.ExternalDNSName),
-		AdminCertFile:     tlsConfig.AdminCertFile,
-		AdminKeyFile:      tlsConfig.AdminKeyFile,
-		CACertFile:        tlsConfig.CACertFile,
+		ClusterName:                   cfg.ClusterName,
+		ControllerInstanceType:        cfg.ControllerInstanceType,
```
bcwaldon: Where are `ControllerInstanceType` and `WorkerInstanceType` actually used in the template below?

bcwaldon:
I'm sort of meh on maintaining a dockerfile & image for a command-line tool. @aaronlevy do you have any strong feelings?

danielwhatmuff:
@bcwaldon it means you can version control your Kubernetes cluster configuration as well as the tool you use to launch it - I've hit problems in the past with things like version differences between aws/vcloud cli tools on laptop to Jenkins. It makes CI for infra more reliable as Jenkins can just run the versioned container to build an environment in aws, then deploy manifests, then run tests.

aaronlevy:
I agree with @bcwaldon. Also, the approach in general I feel a little iffy about. I wouldn't want credentials in docker images the same way I wouldn't want credentials committed in vcs. So now we just get a versioned binary, and a fixed cluster.yaml -- and I wouldn't want to support an image build in this repo just for that.

bcwaldon:
@danielwhatmuff Thank you for the contribution, but I don't feel like this is something we will carry in coreos-kubernetes.

PR #173, opened 2015-11-30
multi-node setup example using libvirt

gtank:
Demonstrates a basic multi-node Kubernetes setup on CoreOS using libvirt. Cluster consists of a master and a single worker node.

### DIFF
--- a/multi-node/libvirt/README.md
+++ b/multi-node/libvirt/README.md
@@ -22,6 +22,11 @@ qemu-img create -f qcow2 -b coreos_production_qemu_image.img kube-worker.qcow2
 These images will store the differences between the VM and the base image
 separately for each machine.
 
+## Generate TLS keys
+
+See the included script `gentls.sh` for steps. Running the script will generate
+a new CA, issue certs for the master, worker, and `kubectl` command line tool.
+
 ## Generate ssh keys
 
 It is good practice to use dedicated credentials for accessing guest systems.
@@ -30,10 +35,6 @@ Generate a new SSH keypair like this:
 cd /var/lib/libvirt/images/coreos
 ssh-keygen -t rsa -b 2048 -f vm_key
 ```
-
-## Generate TLS keys
-
-TODO
 
 ## Set up config drive
 
@@ -45,14 +46,18 @@ need to operate. Assuming you have cloned `coreos-kubernetes` to $CLONEDIR:
 
 For kube-master:
 ```
+cd /var/lib/libvirt/images/coreos
 cp -R $CLONEDIR/multi-node/libvirt/kube-master /var/lib/libvirt/images/coreos/
+cp $CLONEDIR/multi-node/libvirt/certs/kube-apiserver.tar /var/lib/libvirt/images/coreos/kube-master/openstack/latest/
 cp vm_key.pub /var/lib/libvirt/images/coreos/kube-master/openstack/latest/
 
 ```
 
 For kube-worker:
 ```
+cd /var/lib/libvirt/images/coreos
 cp -R $CLONEDIR/multi-node/libvirt/kube-worker /var/lib/libvirt/images/coreos/
+cp $CLONEDIR/multi-node/libvirt/certs/kube-worker.tar /var/lib/libvirt/images/coreos/kube-worker/openstack/latest/
 cp vm_key.pub /var/lib/libvirt/images/coreos/kube-worker/openstack/latest/
 ```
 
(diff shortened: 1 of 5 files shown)


brianredbeard:
Question... Since you're using `libvirt`/`qemu` is there any reason you didn't look at presenting the `cloud-config` via a `9p` filesystem rather than `config-drive`?

multi-node/libvirt/README.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+## Configure kubectl
+Assuming you have cloned `coreos-kubernetes` to $CLONEDIR and run `gentls.sh`:
+```
+MASTER_HOST=192.168.120.10 CA_CERT=$CLONEDIR/multi-node/libvirt/certs/ca.pem ADMIN_KEY=$CLONEDIR/multi-node/libvirt/certs/admin-key.pem ADMIN_CERT=$CLONEDIR/multi-node/libvirt/certs/admin.pem
+kubectl config set-cluster default-cluster --server=https://${MASTER_HOST} --certificate-authority=${CA_CERT}
+kubectl config set-credentials default-admin --certificate-authority=${CA_CERT} --client-key=${ADMIN_KEY} --client-certificate=${ADMIN_CERT}
+kubectl config set-context default-system --cluster=default-cluster --user=default-admin
+kubectl config use-context default-system
```
aaronlevy: nit: A user may have many other contexts already in their kubeconfig. Can we use something a little more descriptive? coreos-libvirt maybe?

multi-node/libvirt/kube-master/openstack/latest/user_data
```diff
@@ -0,0 +1,616 @@
+#!/bin/bash
```
aaronlevy: Does anything differ about this file than multi-node/generic? (same for the kube-worker)

multi-node/libvirt/kube-worker/openstack/latest/user_data
```diff
@@ -0,0 +1,205 @@
+#!/bin/bash
```
aaronlevy: If the only difference here from multi-node/generic is the etcd/controller env vars, maybe instead we override them in a separate ENV_FILE (see line 20 of this file).

multi-node/libvirt/kube-master/openstack/latest/user_data
gtank: Yeah, it handles installing TLS/SSH assets, finding the right IP address without a metadata service to query, and making etcd advertise the correct IP address.

multi-node/libvirt/network-kubernetes.xml
```diff
@@ -0,0 +1,14 @@
...
+  <name>kubernetes</name>
+  <uuid>dcbef342-16a3-4f3e-8307-84bff91076e1</uuid>
+  <forward mode='nat'/>
+  <bridge name='virbrk8s' stp='on' delay='0'/>
+  <mac address='82:84:01:56:df:f3'/>
+  <ip address='192.168.120.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.120.2' end='192.168.120.254'/>
+      <host mac='54:52:00:fe:b3:b0' name='kube-master' ip='192.168.120.10'/>
+      <host mac='54:52:00:fe:b3:b1' name='kube-worker' ip='192.168.120.11'/>
```
aaronlevy: really minor nit: maybe leave some ip span between masters/workers. In the case of multi-master / multi-worker clusters it makes it easier to immediately see the roles from a network perspective. But this may just be my laziness...

multi-node/libvirt/README.md
gtank: Good point. I'll change it.

aaronlevy:
Awesome! Thanks @gtank 

Some minor comments, and then the only big thing would be if we can try and use generic copies of the user data. It's fine if not, it just makes it easier than maintaining a copy for each provider.

/cc @dghubble

aaronlevy:
nit: one of your commits is "mult-node/" rather than "multi-node"

multi-node/libvirt/kube-master/openstack/latest/user_data
aaronlevy: Ah, I see. Okay, fine by me to manage this separately

bcwaldon:
@gtank Were you able to revisit this?

bcwaldon:
@gtank Please reopen when you've had a chance to revisit this.

PR #178, opened 2015-12-03
Multi-AZ & EtcdBackup

cedbossneo:
Add Multi-AZ for workers & controllers
Add ETCD Auto-Backup
Upgrade dependencies

### DIFF
--- a/multi-node/aws/.gitignore
+++ b/multi-node/aws/.gitignore
@@ -2,3 +2,4 @@
 /bin
 /clusters
 /cluster.yaml
+Godeps/_workspace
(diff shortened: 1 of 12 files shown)


multi-node/aws/.gitignore
```diff
@@ -2,3 +2,4 @@
 /bin
 /clusters
 /cluster.yaml
+Godeps/_workspace
```
bcwaldon: I don't believe this is necessary

bcwaldon:
There are a lot of changes here that aren't all necessarily related. Please separate this code apart into small, succinct patches so we have a chance at reviewing it well.

PR #182, opened 2015-12-06
set cloud-provider=aws on aws worker kubelet

### LINKED ISSUE #130: PersistentVolume mounting fails on AWS
With the current alpha version of CoreOS, with kubelet v1.0.6 installed mounting a persistent volume onto a pod fails because the Kublet is unable to find `safe_format_and_mount` which it tries to find @ `/usr/share/google/safe_format_and_mount`. In the latest Kubernetes master this has been resolved by moving the utility into Go and packaging it with the release. However, I'd like to avoid using the prerelease version of Kubernetes if possible.

I understand that `/usr/share` is read only, so it seems that this might require a patch to be included into CoreOS itself? Is anyone aware of any other workarounds? I've also seen people compile custom versions of the Kubelet so they would point to different paths for the script. I'm happy to do the work to get this running if there is interest in fixing it, I'm just not sure what the preferred approach would be.

Related Issues:
https://github.com/kubernetes/kubernetes/issues/7042
https://github.com/kubernetes/kubernetes/pull/8530
https://github.com/kubernetes/kubernetes/issues/11011

Also, the ultimate source of the bug (resolved in master):
https://github.com/kubernetes/kubernetes/blob/v1.0.6/pkg/volume/aws_ebs/aws_util.go#L139

phemmer:
closes #130

### DIFF
--- a/multi-node/aws/artifacts/scripts/install-worker.sh
+++ b/multi-node/aws/artifacts/scripts/install-worker.sh
@@ -85,6 +85,7 @@ ExecStart=/usr/bin/kubelet \
   --config=/etc/kubernetes/manifests \
   --cluster_dns=${DNS_SERVICE_IP} \
   --cluster_domain=cluster.local \
+  --cloud-provider=aws \
   --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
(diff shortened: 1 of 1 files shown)


bcwaldon:
LGTM, thank you!

PR #188, opened 2015-12-08
Add heads-up in replacement of variable when using curl

hermanjunge:
Assuming there will be another poor soul that will be confused... In retrospective, this seems obvious. But you never know.

*EDIT*: Oops! :droplet: In `getting-started` is stated _clearly_ that the syntaxis for `ETCD_ENDPOINTS` is `http://ip:port`. So, I'm not sure. Do we cancel the PR. or add this redundant note anyways?

*EDIT 2*: Modified on observation by @phemmer.


phemmer:
What about just rewording it to:

> Replace `$ETCD_SERVER` with one URL from `$ETCD_ENDPOINTS`

^ "URL" instead of "host". After all, `http://example.com:1234` is not a host (`example.com` would be the host).

hermanjunge:
You are right @phemmer, is clearer that way. I modified the line accordingly.

joshix:
Thanks for making this a little clearer.

PR #185, opened 2015-12-07
allow icmp destination unreachable

phemmer:
closes #184

PR #189, opened 2015-12-08
*: add travis ci tests for AWS build

aaronlevy:


.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: go
+go:
+  - 1.4
+  - 1.5
+install:
+  - go get golang.org/x/tools/cmd/vet
+script:
+  - cd $TRAVIS_BUILD_DIR/multi-node/aws
```
phemmer: Why not test the entire repo? Currently the only go code is `multi-node/aws`, but it would be good to catch other code in the future if/when it's added.

phemmer: Oh, nevermind, godeps only exists within `multi-node/aws`, so build environment would have to be set up separately for any future code.

phemmer:
Another test I add on my own stuff is:

    test -z "$(find . -name '*.go' -exec gofmt -d {} + | tee /dev/stderr)"

This will make sure code is formatted properly. Resolves issues such as [this](https://github.com/coreos/coreos-kubernetes/pull/185/files#diff-da222392c29da368a3a2c580c10576e2R285)

aaronlevy:
@phemmer thanks, I'll add this to the tests. Any reason not to just prune "Godeps" explicitly rather than "_*"?

phemmer:
Either would work. `_*` more closely mimmicks go's expansion of `./...`


-------- Original Message --------

bcwaldon:
@aaronlevy happy to land it if you can get the simple commit together that fixes the tests

PR #208, opened 2015-12-16
Documentation: Explain what is a drop-in

hermanjunge:
A `drop-in` is explained at the `docker` item, right after the `flannel` one. Is desirable to give that explanation before.


aaronlevy:
lgtm

PR #211, opened 2015-12-18
Specificity required during ssl step

bernadinm:
When I was walking though this the first time, was not sure if that directory should have already existed. I am adding this to call out this step.


robszumski:
LGTM.

@joshix Aren't there other locations where we need to create directories?

PR #213, opened 2015-12-21
Multi-az support

cedbossneo:


multi-node/aws/build
```diff
@@ -15,7 +15,8 @@ if [ -n "$(git diff --shortstat 2> /dev/null | tail -n1)" ]; then
 fi
 
 
-
+echo "Restoring deps"
+godep restore
```
bcwaldon: This should not be necessary. The `go build` command below overrides `GOPATH` to place the workspace earlier in the search path than your existing `GOPATH`.

multi-node/aws/.gitignore
```diff
@@ -2,3 +2,4 @@
 /bin
 /clusters
 /cluster.yaml
+Godeps/_workspace
```
bcwaldon: This is not necessary

multi-node/aws/pkg/cluster/cloud_config_util.go
```diff
@@ -8,7 +8,7 @@ import (
 var templateRef *regexp.Regexp
 
 func init() {
-	templateRef = regexp.MustCompile(`{{\s*[a-zA-Z0-9]+(\|[a-zA-Z0-9]+)*\s*}}`)
+	templateRef = regexp.MustCompile(`{{\s*[a-zA-Z0-9\:]+(\|[a-zA-Z0-9]+)*\s*}}`)
```
bcwaldon: What is this change for?

multi-node/aws/pkg/cluster/cloud_config_worker.go
```diff
@@ -7,7 +7,7 @@ coreos:
 
   flannel:
     interface: $private_ipv4
-    etcd_endpoints: http://10.0.0.50:2379
+    etcd_endpoints: http://master.kube.internal:2379
```
bcwaldon: I feel like this should be `kubernetes.cluster.internal`. @aaronlevy ?

multi-node/aws/pkg/cluster/stack_template.go
```diff
@@ -13,19 +13,23 @@ const (
 	resNameVPCGatewayAttachment         = "VPCGatewayAttachment"
 	resNameRouteTable                   = "RouteTable"
 	resNameRouteToInternet              = "RouteToInternet"
-	resNameSubnet                       = "Subnet"
-	resNameSubnetRouteTableAssociation  = "SubnetRouteTableAssociation"
+	resNameSubnetA 						= "SubnetA"
```
bcwaldon: spacing is off

multi-node/aws/pkg/cluster/stack_template.go
```diff
@@ -13,19 +13,23 @@ const (
...
-	resNameAlarmControllerRecover       = "AlarmControllerRecover"
 	resNameAutoScaleWorker              = "AutoScaleWorker"
 	resNameLaunchConfigurationWorker    = "LaunchConfigurationWorker"
+	resNameAutoScaleController          = "AutoScaleController"
+	resNameLaunchConfigurationController= "LaunchConfigurationController"
 	resNameIAMRoleController            = "IAMRoleController"
 	resNameIAMInstanceProfileController = "IAMInstanceProfileController"
 	resNameIAMRoleWorker                = "IAMRoleWorker"
 	resNameIAMInstanceProfileWorker     = "IAMInstanceProfileWorker"
+	resNameRoute53Zone					= "VPCZone"
```
bcwaldon: spacing is off

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -156,25 +157,18 @@ func (c *Cluster) Create(tlsConfig *TLSConfig) error {
...
-		parameters = append(parameters, &cloudformation.Parameter{
-			ParameterKey:     aws.String(parAvailabilityZone),
-			ParameterValue:   aws.String(c.cfg.AvailabilityZone),
-			UsePreviousValue: aws.Bool(true),
-		})
-	}
-
 	tmplURL := fmt.Sprintf("%s/template.json", c.cfg.ArtifactURL)
-	return createStackAndWait(cloudformation.New(c.aws), c.stackName(), tmplURL, parameters)
+	return createStackAndWait(cloudformation.New(session.New(c.aws)), c.stackName(), tmplURL, parameters)
```
bcwaldon: Why is the added `session.New()` call necessary?

multi-node/aws/pkg/cluster/stack_template.go
```diff
@@ -396,82 +431,86 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
...
-			"Statistic":          "Minimum",
-			"Period":             "60",
-			"EvaluationPeriods":  "5",
-			"ComparisonOperator": "GreaterThanThreshold",
-			"Threshold":          "0",
-			"AlarmActions": []interface{}{
+			"HostedZoneConfig":  map[string]interface{}{
+				"Comment": "The Kubernetes Internal zone",
+			},
+			"Name": "kuber.internal",
```
bcwaldon: typo

multi-node/aws/pkg/cluster/stack_template.go
```diff
@@ -590,17 +669,18 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
 		"Description": "Number of worker instances to create, may be modified later",
 	}
 
+	par[parControllerCount] = map[string]interface{}{
```
bcwaldon: This needs to be configurable through cluster.yaml

multi-node/aws/pkg/cluster/cloud_config_util.go
cedbossneo: This allow handling AWS::Region in template

multi-node/aws/pkg/cluster/cloud_config_worker.go
cedbossneo: This allows every node access the kubernetes master through load balancer

multi-node/aws/pkg/cluster/cluster.go
cedbossneo: I had to upgrade deps to handle ELB and it's now necessary to create a session instance.

cedbossneo:
Ready for another review.

multi-node/aws/pkg/cluster/cluster.go
bcwaldon: OK, I see you updated the vendored godeps

bcwaldon:
@cedbossneo Overall, this looks good. The one architectural change I want to suggest is to not manage the `kube.internal` zone at all, and to route traffic from workers directly to the ELB hostname. Would this be possible?

cedbossneo:
This should not be a problem. I will make this change quickly after
Christmas. :)

cedbossneo:
Ready for review

aaronlevy:
Sorry I'm late to this PR. So it looks like a few things are going on here:

- Adds multiple master nodes (and ELB fronting masters)
- Adds option of 2 availability zones
- Bumps AWS SDK version
- Modifies template replacements

I feel like each of these would need to really be broken out into its own PR. As it stands this is pretty difficult to adequately review, because it touches on so much.

Some general notes:

- This doesn't seem to handle etcd replication. It would seem that a request routed through the ELB would be hitting a different (and isolated) etcd instance on each master node.

- Only supports 2 static AZs. This leads to a lot of "A/B" code duplication and I think it would be more valuable to allow launching nodes into arbitrary # of AZs without the static AZ-A/AZ-B code.

    An option here could be modifying the tool to allow you to re-run and launch "more" workers in different AZs that attach to an existing master. See upstream style support for this: https://github.com/kubernetes/kubernetes/pull/16735

- Bumping AWS SDK should be done in its own PR, and also tested in isolation. I believe the version we currently have supports ELBs, but either way, both changing the SDK and the functionality of the tool should be done separately.

- Changing the template replacements also should likely be broken out into another PR so the implications / upgradability / etc can be discussed and tested in isolation of functionality changes in the rest of the tool.

As this PR currently stands, I don't feel comfortable merging. This could be broken out into smaller units of work that are more easily tested / reasoned about. I might even suggest for the larger pieces (multi-master / multi-az) writing a proposal document so the implementation can be discussed prior to you spending a lot of time on code.

Related (automated conformance testing): https://github.com/coreos/coreos-kubernetes/issues/224

PR #214, opened 2015-12-21
Etcd Auto Backup & Restore

cedbossneo:


cedbossneo:
Ready for review but i'm not sure about implementing the restore on server reboot.

aaronlevy:
Thank you for this contribution, but I feel like this is out of the scope of the intentions / motivations of this tool, primary of which is to launch a fully-functioning Kubernetes cluster that an administrator can build upon.

In this case the etcd backup/restore process should be left to an administrator to decide, rather than making the assumption for them.

PR #222, opened 2015-12-23
kube-aws: add parameter for spot price

iameli:
I made the parameter a string, 'cause I'm scared about the floating-point math from YAML through Go to AWS. It appears to be a string in the AWS API, in any event.

### DIFF
--- a/multi-node/aws/cluster.yaml.example
+++ b/multi-node/aws/cluster.yaml.example
@@ -33,6 +33,9 @@ externalDNSName:
 # Disk size (GiB) for worker nodes
 #workerRootVolumeSize: 30
 
+# Price (Dollars) to bid for spot instances. Omit for on-demand instances.
+# workerSpotPrice: "0.05"
+
 # Location of kube-aws artifacts used to deploy a new
 # Kubernetes cluster. The necessary artifacts are already
 # available in a public S3 bucket matching the version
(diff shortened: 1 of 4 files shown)


aaronlevy:
In general this seems fine to me. I guess my reservation would be that there are a lot of configurables in cloud-formation, and do we really want to plumb them all through to a yaml config? e.g. How do we weigh that vs just assuming the cloud-formation can be modified after render.

/cc @colhom

colhom:
In general, I feel that the Cloudformation template should be edited directly to extend a template _beyond_ the scope of Kubernetes cluster parameters. Examples include peering connections to other VPCs, additional security group rules to allow external access to nodeports, etc. 

We already plumb ec2 instance type and worker count for the ASG through the cluster.yaml. The rational behind ec2 instance type and worker count being part of cluster.yaml is that 
* _Necessity_: These decisions are pre-reqs for creating a working cluster on AWS
* _Convenience_: We assume majority of users will want to define these paramters, so we expose them via the cluster yaml rather than just inferring the choice and leaving it to the end user to edit the cloudformation in order to change them. 

By extension, I would say the _on-demand vs spot instance_ decision falls in the same category and should be part of the cluster yaml.

\cc @aaronlevy

aaronlevy:
LGTM pending this passing tests (CI or Sanity). @colhom mind shepherding this through that process?

colhom:
@aaronlevy i'll run this PR along with #252 through the conformance tests, and will merge them if successful.

colhom:
Tested this PR via conformance test with and without `workerSpotPrice` param specified. Everything checks out (coreos-kubernetes-kube-aws, jobs 50 and 51 for coreos people)

PR #227, opened 2015-12-30
Added Egress to WorkerSecurityGroup for External ICMP Pings

ankitforcode:
cc @aaronlevy

### DIFF
--- a/contrib/hyperkube-image/.gitignore
+++ b/contrib/hyperkube-image/.gitignore
@@ -1,0 +1,1 @@
+_workdir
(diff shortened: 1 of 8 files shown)


aaronlevy:
Any reason not to just allow all outbound ICMP?

ankitforcode:
@aaronlevy I don't see any reasons as such. Since, I am just following the same pattern for Egress as we had for Ingress.

aaronlevy:
Would you be able to change this so we allow all outbound ICMP traffic, & squash commits?

PR #228, opened 2015-12-30
contrib: Add hyperkube image build assets

aaronlevy:


contrib/hyperkube-image/Makefile
```diff
@@ -0,0 +1,20 @@
+TAG = v1.1.2
+PREFIX = quay.io/coreos
+
+.PHONY: all container push clean
+
+all: container
+
+hyperkube:
+	#TODO(aaron): build from source?
```
philips: Yes, aren't we building our own containers for tectonic releases? If not we will need to at some point to do hot fixes.

aaronlevy: Good point. And we currently do have patches we are carrying for the Kubelet in the OS build.

philips:
LGTM.

Only minor feedback would be to use REPO="quay.io/coreos/hyperkube" instead of the prefix thing since it is a bit more clear. Or maybe even just TAG="quay.io/coreos/hypberkube:v1.4.0".

PR #232, opened 2016-01-02
docs: add details on /srv vs. /etc; s/master/controller/ throughout

### LINKED ISSUE #203: kubernetes deploy-master doesn't suggest creating /srv/ on coreos
Moving issue to correct location: @mischief.

Do you have any more info you could share?

omkensey:
Added more details on the distinction between /srv/kubernetes and
/etc/kubernetes at the top of the master deployment instructions.

Updated docs to use "controller" instead of "master" to conform with
current terminology, and to use "node" to refer specifically to worker
nodes in most places.  (Variable names and directory path references
left unchanged to avoid inadvertently breaking other docs.)

Fixes #203

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -1,12 +1,19 @@
-# Deploy Kubernetes Master Machine
+# Deploy Kubernetes Master Node(s)
 
-Boot a single CoreOS machine which will be used as the Kubernetes master. You must use a CoreOS version 773.1.0+ on the Alpha or Beta channel for the `kubelet` to be present in the image.
+Boot a single CoreOS machine which will be used as the Kubernetes master node. You must use a CoreOS version 773.1.0+ on the Alpha or Beta channel for the `kubelet` to be present in the image.
 
 See the [CoreOS Documentation](https://coreos.com/os/docs/latest/) for guides on launching nodes on supported platforms.
 
-Manual configuration of the required Master services is explained below, but most of the configuration could also be done with cloud-config, aside from placing the TLS assets on disk. These secrets shouldn't be stored in cloud-config for enhanced security.
+Manual configuration of the required master node services is explained below, but most of the configuration could also be done with cloud-config, aside from placing the TLS assets on disk. These secrets shouldn't be stored in cloud-config for enhanced security.
 
-If you are deploying multiple master nodes in a high-availability cluster, these instructions can be repeated for each master node you wish to launch.
+The instructions below configure the required master node services using two main directories:
+
+* `/etc/kubernetes/manifests` for services that the kubelet should run on every master node
+* `/srv/kubernetes/manifests` for services that should run only on a single master node at a time (determined by a leader election).  The podmaster on that node manages the startup of these services by copying their manifests from here to the `/etc/kubernetes/manifests` directory, and the kubelet picks them up from there.
+
+It is vital to understand the distinction between the two and ensure that the scheduler and controller manager manifests are not accidentally placed directly in the `/etc/kubernetes/manifests` directory.
+
+If you are deploying multiple master nodes in a high-availability cluster, these instructions can be repeated for each one you wish to launch.
 
 ## Configure Service Components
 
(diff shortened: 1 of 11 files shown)


Documentation/deploy-master.md
```diff
@@ -1,12 +1,19 @@
...
 
 See the [CoreOS Documentation](https://coreos.com/os/docs/latest/) for guides on launching nodes on supported platforms.
 
-Manual configuration of the required Master services is explained below, but most of the configuration could also be done with cloud-config, aside from placing the TLS assets on disk. These secrets shouldn't be stored in cloud-config for enhanced security.
+Manual configuration of the required controller services is explained below, but most of the configuration could also be done with cloud-config, aside from placing the TLS assets on disk. These secrets shouldn't be stored in cloud-config for enhanced security.
 
-If you are deploying multiple master nodes in a high-availability cluster, these instructions can be repeated for each master node you wish to launch.
+The instructions below configure the required controller services using two main directories:
+
+* /etc/kubernetes/manifests for services that the kubelet should run on every controller
```
robszumski: backticks for `/etc/kubernetes/manifests`

Documentation/deploy-master.md
```diff
@@ -1,12 +1,19 @@
...
+* /srv/kubernetes/manifests for services that should run only on a single controller at a time (determined by a leader election).  The podmaster on that controller manages the startup of these services by copying their manifests from here to the /etc/kubernetes/manifests directory, and the kubelet picks them up from there.
```
robszumski: backticks

Documentation/deploy-master.md
```diff
@@ -1,12 +1,19 @@
...
+
+It is vital to understand the distinction between the two and ensure that the scheduler and controller manager manifests are not accidentally placed directly in the /etc/kubernetes/manifests directory.
```
robszumski: backticks

robszumski:
LGTM other than minor nits

omkensey:
Picked nits.

aaronlevy:
Internally we've been using "controller", but upstream it is referred to as "master node", "master components", and "control plane". I think should try to move our wording to be more in line with the upstream terminology.

See: http://kubernetes.io/v1.1/docs/design/architecture.html

aaronlevy:
@robszumski @omkensey thoughts on controller vs master?

omkensey:
Yeah, it seems they're now using "node" to refer to all members of the cluster (which is in line with terminology for most other cluster software), and distinguishing into "master nodes" and "worker nodes" (looking also at https://github.com/kubernetes/kubernetes/blob/master/docs/admin/cluster-management.md).  Concur we should conform to that usage.  I'll update PR with a new commit in the AM.

Documentation/kubernetes-upgrade.md
```diff
@@ -65,9 +65,9 @@ The kube-proxy is run as a "static pod". To upgrade the pod definition, simply m
...
 1. Prepare new pod manifests for master nodes
-1. Prepare new pod manifests for worker nodes
-1. For each master node:
+2. Prepare new pod manifests for worker nodes
+3. For each master node:
     1. Back up existing manifests
-    1. Update manifests
-1. Repeat same steps above for each worker node
+    2. Update manifests
+4. Repeat item 3 for each worker node
```
aaronlevy: nit: This makes it more clear when looking at the raw markdown, but the rendered output won't necessarily follow this numbering (why I tend to use "1." for every list item).

aaronlevy: Maybe we just add explicit steps for worker too. e.g. 

```
1. Prepare new pod manifests for master nodes
1. Prepare new pod manifests for worker nodes
1. For each master node:
    1. Back up existing manifests
    1. Update manifests
1. For each worker node
    1. Back up existing manifests
    1. Update manifests
```

aaronlevy:
One nit but otherwise looks good. Can you also squash cleanup commits.

Documentation/kubernetes-upgrade.md
omkensey: It will still render correctly regardless though, right (at the cost of possibly getting out of sync with rendered numbering)?  I guess there's sort of precedent in HTML's use of generic "numbered/bulleted list item" notation.  Do we have a project doc-markup style guide?

Will fix and push in a bit.

aaronlevy:
LGTM

PR #234, opened 2016-01-07
kube-aws: fix network defaults bug

### LINKED ISSUE #226: Failed initializing TLS infrastructure: x509: certificate contained IP address of length 0
In order to get the cidr configs from #190 I'm attempting to use the master kube-aws build. I have reproduced this error with go 1.5.1 and go 1.5.2. I'm on Mac OS 10.11.2

```
$ kube-aws up
Failed initializing TLS infrastructure: x509: certificate contained IP address of length 0
```

After some debugging I've determined it's coming from `return x509.ParseCertificate(certDERBytes)` in `NewSelfSignedCACertificate`.

I've tried to resolve this my self by adding IP addresses to the x509 Certificate details, but nothing changed.

(This did not work):
```diff
diff --git a/multi-node/aws/pkg/tlsutil/x509.go b/multi-node/aws/pkg/tlsutil/x509.go
index 206403c..f909229 100644
--- a/multi-node/aws/pkg/tlsutil/x509.go
+++ b/multi-node/aws/pkg/tlsutil/x509.go
@@ -35,7 +35,9 @@ type ClientCertConfig struct {

 func NewSelfSignedCACertificate(cfg CACertConfig, key *rsa.PrivateKey) (*x509.Certificate, error) {
        now := time.Now()
+       ips := []net.IP{net.ParseIP("127.0.0.1")}
        tmpl := x509.Certificate{
+               IPAddresses:  ips,
                SerialNumber: new(big.Int).SetInt64(0),
                Subject: pkix.Name{
                        CommonName:   cfg.CommonName,
```

colhom:
make sure default network configuration values are applied to the configuration struct at parse time. the defaults were being validated if the supplied value is empty, but the default value was not being written back to the configuration struct.

this is important so that the inferred default values are set before
- the network toplogy is validated
- ssl certificates are generated

fixes #226

### DIFF
--- a/multi-node/aws/pkg/cluster/config_test.go
+++ b/multi-node/aws/pkg/cluster/config_test.go
@@ -1,0 +1,90 @@
+package cluster
+
+import (
+	"testing"
+)
+
+const testVer = "1"
+const minimalConfigYaml = `externalDNSName: test-external-dns-name
+keyName: test-key-name
+region: test-region
+clusterName: test-cluster-name
+`
+
+var goodNetworkingConfigs []string = []string{
+	``, //Tests validity of default network config values
+	`
+vpcCIDR: 10.4.3.0/24
+instanceCIDR: 10.4.3.0/24
+controllerIP: 10.4.3.5
+podCIDR: 172.4.0.0/16
+serviceCIDR: 172.5.0.0/16
+kubernetesServiceIP: 172.5.100.100
+dnsServiceIP: 172.5.100.101
+`, `
+vpcCIDR: 10.4.0.0/16
+instanceCIDR: 10.4.3.0/24
+controllerIP: 10.4.3.5
+podCIDR: 10.6.0.0/16
+serviceCIDR: 10.5.0.0/16
+kubernetesServiceIP: 10.5.100.100
+dnsServiceIP: 10.5.100.101
+`,
+}
+
+var incorrectNetworkingConfigs []string = []string{
+	`
+vpcCIDR: 10.4.0.0/16
+instanceCIDR: 10.5.3.0/24 #instanceCIDR not in vpcCIDR
+controllerIP: 10.5.3.5
+podCIDR: 10.6.0.0/16
+serviceCIDR: 10.5.0.0/16
+kubernetesServiceIP: 10.5.100.100
+dnsServiceIP: 10.5.100.101
+`, `
+vpcCIDR: 10.4.3.0/16
+instanceCIDR: 10.4.3.0/24
+controllerIP: 10.4.3.5
+podCIDR: 172.4.0.0/16
+serviceCIDR: 172.5.0.0/16
+kubernetesServiceIP: 172.10.100.100 #kubernetesServiceIP not in service CIDR
+dnsServiceIP: 172.5.100.101
+`, `
+vpcCIDR: 10.4.3.0/16
+instanceCIDR: 10.4.3.0/24
+controllerIP: 10.4.3.5
+podCIDR: 10.4.0.0/16 #vpcCIDR overlaps with podCIDR
+serviceCIDR: 172.5.0.0/16
+kubernetesServiceIP: 172.5.100.100
+dnsServiceIP: 172.5.100.101
+
+`, `
+vpcCIDR: 10.4.3.0/16
+instanceCIDR: 10.4.3.0/24
+controllerIP: 10.4.3.5
+podCIDR: 172.4.0.0/16
+serviceCIDR: 172.5.0.0/16
+kubernetesServiceIP: 172.5.100.100
+dnsServiceIP: 172.6.100.101 #dnsServiceIP not in service CIDR
+`,
+}
+
+func TestNetworkValidation(t *testing.T) {
+
+	for _, networkConfig := range goodNetworkingConfigs {
+		config := NewDefaultConfig(testVer)
+		configBody := minimalConfigYaml + networkConfig
+		if err := decodeConfigBytes(config, []byte(configBody)); err != nil {
+			t.Errorf("Correct config tested invalid: %s\n%s", err, networkConfig)
+		}
+	}
+
+	for _, networkConfig := range incorrectNetworkingConfigs {
+		config := NewDefaultConfig(testVer)
+		configBody := minimalConfigYaml + networkConfig
+		if err := decodeConfigBytes(config, []byte(configBody)); err == nil {
+			t.Errorf("Incorrect config tested valid, expected error:\n%s", networkConfig)
+		}
+	}
+
+}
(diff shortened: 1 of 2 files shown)


colhom:
@aaronlevy can you PTAL when you get a chance?

aaronlevy:
In general this seems fine. Couple things:

Can you add unit tests for this package? This seems like a bug that would have been quickly surfaced had there been tests.

I don't really like that there is a `Valid()` function which modifies the config struct with default values. I would expect that the function just checks what I have and reports back.

Would it be possible to modify this so that the Valid function is not making changes to the underlying config? Maybe something like:

config := NewDefaultConfig()
config.foo = bar
config.bar = baz
if !config.Valid() {exit}

aaronlevy:
Also, CI check is failing for some reason...

colhom:
@aaronlevy defaults are defined in a sane way. Also basic unit tests for the config parser/validator!

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	if err = yaml.Unmarshal(d, &out); err != nil {
+		return fmt.Errorf("failed decoding config file: %v", err)
+	}
+
+	return nil
+}
+
+func TestCorrectMinimalConfig(t *testing.T) {
```
aaronlevy: This seems unnecessary. It is just testing a function that is defined in the test itself.

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+	cfg := correctMinimalConfig()
+	if err := cfg.Valid(); err != nil {
+		t.Fatalf("Could not validate: %s\n : %+v\n", err, cfg)
+	}
+}
+
+func writeTestFile(contents string) (string, error) {
+	tpath := filepath.Join(os.TempDir(), fmt.Sprintf("test-file-%d", rand.Int()))
```
aaronlevy: Unit tests should avoid network/disk IO. We should be able to test the config functionality without hitting disk (or need to restructure code such that the parts we wish to test take a reader/writer interface (not necessarily require a file).

multi-node/aws/pkg/cluster/config_test.go
colhom: It's testing that all the defaults pass the validity check.

multi-node/aws/pkg/cluster/config_test.go
colhom: I'll extract an inner function from DecodeConfigFile which takes a Reader

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	defer f.Close()
+	if _, err := io.Copy(f, bytes.NewBuffer([]byte(contents))); err != nil {
+		return "", err
+	}
+
+	return tpath, nil
+}
+
+func TestConfigParserDefaults(t *testing.T) {
```
aaronlevy: This function also pretty much just tests other functions defined in this test file. The only external call is .Valid(), but that is tested elsewhere. I think this can be dropped as well

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+func TestNetworkValidation(t *testing.T) {
+	testConfig := func(networkConfig string) error {
+		configPath, err := writeTestFile(minimalConfigYaml + networkConfig)
+		if err != nil {
+			t.Fatalf("error opening test file: %s", err)
+		}
+
+		config := NewDefaultConfig(testVer)
+		if err := parseConfig(config, configPath); err != nil {
+			t.Fatalf("error decoding config file: %s", err)
```
aaronlevy: Why use fatal instead of error?

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+var minimalConfigYaml string = fmt.Sprintf(
+	minimalYamlTemplate,
+	externalDNSName,
+	keyName,
+	region,
+	clusterName,
+)
+
+func correctMinimalConfig() *Config {
```
aaronlevy: Given that we probably don't need the other functions that make use of this, we can simplify the minimalConfigYaml

I'd probably suggest getting rid of this function (no longer called). Then also the individual config consts + minimalConfigYaml + minimalConfigTemplate can just be converted into a simple 4 line string var with fixed test values.

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+var goodNetworkingConfigs []string = []string{
+	`
+vpcCIDR: 10.4.3.0/24
+instanceCIDR: 10.4.3.0/24
+controllerIP: 10.4.3.5
+podCIDR: 172.4.0.0/16
+serviceCIDR: 172.5.0.0/16
+kubernetesServiceIP: 172.5.100.100
```
aaronlevy: Not sure that this is correct. Last I checked the kubernetes service is not configurable and it will be the first IP in the --service-ip-range (serviceCIDR).

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+dnsServiceIP: 172.5.100.101
```
aaronlevy: Nit: if we wanted to do more validation, this also should not be the first IP in the service ip range (it will conflict with kubernetes service).

colhom: yeah, i think I ran into that actually. I should remove that option entirely, but in a separate PR?

colhom: We should not expose `kubernetesServiceIP` in the yaml but still compute it from `serviceCIDR` and validate it doens't conflict with the `dnsServiceIP`.

multi-node/aws/pkg/cluster/config_test.go
```diff
@@ -0,0 +1,188 @@
...
+	correctConfig := correctMinimalConfig()
+	if *parsedConfig != *correctConfig {
+		t.Fatalf("parsed config does not match:\n parsed: %+v \n correct %+v\n",
+			parsedConfig,
+			correctConfig,
+		)
+	}
+}
+
+var goodNetworkingConfigs []string = []string{
```
aaronlevy: minor nit: unless these network configs are used more globally, I would put them as part of the table driven tests inside the actual test functions (would also separate them into "testGoodConfig() and testBadConfig()")

aaronlevy:
LGTM after squash of cleanup commits & assuming this has been tested.

Also, the KubernetesServiceIP should be removed soon as it's not a configurable, but fine to do that in another PR.

colhom:
squashed and tested. we really need to get that conformance test thing going....

PR #236, opened 2016-01-07
enable heapster and influxdb by default

joeatwork:
This change pulls in the heapster, influxdb, and grafana addons from Kubernetes and deploys them to the single node vagrant cluster by default. If the approach seems acceptable, I'll submit similar changes for our multi-node tools.

### DIFF
--- a/multi-node/generic/controller-install.sh
+++ b/multi-node/generic/controller-install.sh
@@ -520,6 +520,216 @@ EOF
 EOF
     }
 
+    # For one controller and one worker node, fix memory at
+    # 224Mi. Multiple node clusters should scale up their memory by
+    # about 12Mi per node
+    local TEMPLATE=/srv/kubernetes/manifests/heapster-rc.json
+    [ -f $TEMPLATE ] || {
+        echo "TEMPLATE: $TEMPLATE"
+        mkdir -p $(dirname $TEMPLATE)
+        cat << EOF > $TEMPLATE
+{
+  "apiVersion": "v1",
+  "kind": "ReplicationController",
+  "metadata": {
+    "name": "heapster-v10",
+    "namespace": "kube-system",
+    "labels": {
+      "k8s-app": "heapster",
+      "version": "v10",
+      "kubernetes.io/cluster-service": "true"
+    }
+  },
+  "spec": {
+    "replicas": 1,
+    "selector": {
+      "k8s-app": "heapster",
+      "version": "v10"
+    },
+    "template": {
+      "metadata": {
+        "labels": {
+          "k8s-app": "heapster",
+          "version": "v10",
+          "kubernetes.io/cluster-service": "true"
+        }
+      },
+      "spec": {
+        "containers": [
+          {
+            "image": "gcr.io/google_containers/heapster:v0.18.2",
+            "name": "heapster",
+            "resources": {
+              "limits": {
+                "cpu": "100m",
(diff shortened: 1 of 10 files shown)


single-node/user-data
```diff
@@ -111,6 +111,7 @@ spec:
     - proxy
     - --master=http://127.0.0.1:8080
     - --proxy-mode=iptables
+    - --v=4
```
joeatwork: Hey, how'd that get in here?

multi-node/generic/controller-install.sh
```diff
@@ -520,6 +520,296 @@ EOF
...
+              },
+              "requests": {
+                "cpu": "100m",
+                "memory": "224Mi"
+              }
+            },
+            "command": [
+              "/heapster",
+              "--source=kubernetes:''",
+              "--sink=influxdb:http://monitoring-influxdb.kube-system.svc.cluster.local:8086",
```
aaronlevy: We may want to be less specific here. We should be able to just use "http://monitoring-influxdb:8086".

multi-node/generic/controller-install.sh
```diff
@@ -520,6 +520,296 @@ EOF
...
+    }
+  }
+}
+EOF
+    }
+
+    # This configuration leaves Grafana open without authentication (other than
+    # authentication via the Kubernetes API). In production, grafana
+    # authentication should be enabled.
+    local TEMPLATE=/srv/kubernetes/manifests/influxdb-grafana-rc.json
```
aaronlevy: Do we necessarily *need* grafana as a default addon? I agree that heapster is necessary because a core feature (horizontal scaling) requires it. However, I lean toward not installing optional applications, and leaving this decision up to user (and something we could possible add to documentation). Thoughts?

multi-node/vagrant/Vagrantfile
```diff
@@ -21,6 +21,10 @@ if File.exist?(CONFIG)
   require CONFIG
 end
 
+if $worker_vm_memory < 1024
+  puts "Controllers should have at least 1024 MB of memory"
```
aaronlevy: checking worker memory, but commenting on controller memory.

aaronlevy:
Couple minor comments/questions. 

Only other thing would be that we probably need to also keep the AWS assets in sync with this too: https://github.com/coreos/coreos-kubernetes/tree/master/multi-node/aws/artifacts/scripts

multi-node/generic/controller-install.sh
joeatwork: Fixed in bfa2f27

multi-node/vagrant/Vagrantfile
joeatwork: Fixed in 4d7b502

multi-node/generic/controller-install.sh
joeatwork: Remove grafana, add evidence that horizontal scaling works.

joeatwork:
Still TBD
* [ ] ensure influxdb url is simple for all references
* [ ] remove grafana
* [ ] one more review pass

aaronlevy:
Can you move the aws docs changes into a separate PR (or at least commit) they seem to be somewhat unrelated.

After that, LGTM.

joeatwork:
Docs changes split into a separate commit - thanks!

aaronlevy:
LGTM

Issue #237, opened 2016-01-08
Enable Daemonsets by default


aaronlevy:
Daemonsets are not currently enabled unless an extra flag is passed to the apiserver. We should consider enabling this feature.

See: http://kubernetes.io/v1.1/docs/admin/daemons.html#caveats

Ref: https://github.com/coreos/coreos-kubernetes/issues/235

PR #239, opened 2016-01-08
k8s: enable deployment and daemonset extension APIs

aaronlevy:
Fixes https://github.com/coreos/coreos-kubernetes/issues/237


colhom:
Looks good, LGTM.

On that topic, I've just filed #243

PR #240, opened 2016-01-11
Reproducible assets

colhom:
This gzips all assets into the cloudformation template (and subsequently into cloud-config). Removes the need for an s3 intermediary between cloudformation and on-disk deployment assets.


colhom:
The next step (on this PR, or another) would be changing kube-aws flow such that:
```
kube-aws render --output=~/src/deployments/colins-cluster
(colins-cluster folder contains the entire deployment vector, can be checked into version control)
kube-aws up --assets=~/src/deployments/colins-cluster
(launch cluster from reproducible assets)
```
besides auditing / reproducilblity, should provide a solid foundation on which to build cluster upgrades and what-not.

colhom:
\cc @aaronlevy PTAL

robszumski:
If `kube-aws render` included all assets, it would fulfill the `--dry-run` idea in https://github.com/coreos/coreos-kubernetes/issues/241.

Right now it's missing cert generation/output.

multi-node/aws/artifacts/scripts/install-controller.sh
```diff
@@ -111,17 +108,17 @@ EOF
 	}
 
 	mkdir -p /etc/kubernetes/manifests
-	template manifests/controller/kube-proxy.yaml /etc/kubernetes/manifests/kube-proxy.yaml
-	template manifests/controller/kube-apiserver.yaml /etc/kubernetes/manifests/kube-apiserver.yaml
-	template manifests/controller/kube-podmaster.yaml /etc/kubernetes/manifests/kube-podmaster.yaml
+	template /tmp/manifests/controller/kube-proxy.yaml /etc/kubernetes/manifests/kube-proxy.yaml
+	template /tmp/manifests/controller/kube-apiserver.yaml /etc/kubernetes/manifests/kube-apiserver.yaml
+	template /tmp/manifests/controller/kube-podmaster.yaml /etc/kubernetes/manifests/kube-podmaster.yaml
```
aaronlevy: Can we namespace this? Maybe "/tmp/coreos/manifests/..." Or even just switch $ARTIFACT_URL to $ARTIFACT_DIR to allow a user to easily override..

multi-node/aws/artifacts/scripts/install-controller.sh
```diff
@@ -37,16 +34,16 @@ ENV_FILE=/run/coreos-kubernetes/options.env
 
 function template {
 	# use a heredoc so the quoting & whitespace in the
-	# downloaded artifact is preserved, but env variables
+	# artifact is preserved, but env variables
 	# can still be evaluated
 	eval "cat <<EOF
-$(curl --silent -L "${ARTIFACT_URL}/$1")
+$(cat $1)
```
aaronlevy: I don't know a good alternative to this off the top of my head, but eval'ing a heredoc that is the subshell output of cat'ing a file ... seems like there must be a better way.

Quick googling brings up envsubst, but looks like that my not be available in coreOS.

I won't block on this, but maybe just do a quick search to see if there's a better way of doing this...

multi-node/aws/cmd/kube-aws/command_up.go
```diff
@@ -14,8 +14,12 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"archive/tar"
+	"compress/gzip"
+	"encoding/base64"
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/cluster"
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/tlsutil"
+	"io"
```
aaronlevy: Import ordering is off. Can you group stdlib first, then package specific, then other. See: https://github.com/golang/go/wiki/CodeReviewComments#imports

multi-node/aws/cmd/kube-aws/command_up.go
```diff
@@ -116,14 +130,89 @@ func getCloudFormation(url string) (string, error) {
...
+	}
+
+	if err := filepath.Walk(path, tarHandler); err != nil {
+		stderr("Error tar-ing directory %s: %s", path, err)
+		os.Exit(1)
+	}
+	return buf
+}
+
+func mustReadFile(loc string) *bytes.Buffer {
```
aaronlevy: This function name doesn't accurately describe what this function does anymore (read file / gzip / base64 encode)

multi-node/aws/cmd/kube-aws/command_up.go
```diff
@@ -226,19 +329,19 @@ func initTLS(cfg *cluster.Config, dir string) (*cluster.TLSConfig, error) {
 
 	tlsConfig := cluster.TLSConfig{
 		CACertFile:        caCertPath,
-		CACert:            mustReadFile(caCertPath),
+		CACert:            mustReadFile(caCertPath).Bytes(),
```
aaronlevy: If you're just calling bytes() everywhere anyway, for code clarity might make sense to just return []byte

aaronlevy:
I'm having doubts about this approach. How close are we to the 16K limit when gzipping assets? For example, we're going to need to add new templates really soon (heapster/influxdb).

multi-node/aws/pkg/cluster/stack_template.go
```diff
@@ -560,35 +564,54 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
...
 	par[parWorkerCert] = map[string]interface{}{
 		"Type":        "String",
-		"Description": "PEM-formatted kubelet (worker) certificate, base64-encoded",
+		"Description": "PEM-formatted kubelet (worker) certificate, (gzip+base64) encoding",
 	}
 
 	par[parWorkerKey] = map[string]interface{}{
 		"Type":        "String",
-		"Description": "PEM-formatted kubelet (worker) key, base64-encoded",
+		"Description": "PEM-formatted kubelet (worker) key, (gzip+base64) encoding",
```
aaronlevy: We probably need to add documentation to our "launch stack" button on the website if we're requiring this. The base64 tripped up enough people, but requiring them to gzip then base64 is going to be more of an issue

multi-node/aws/pkg/cluster/config.go
```diff
@@ -152,7 +153,7 @@ func decodeConfigBytes(out *Config, d []byte) error {
 func NewDefaultConfig(ver string) *Config {
 	return &Config{
 		ClusterName:         "kubernetes",
-		ArtifactURL:         DefaultArtifactURL(ver),
+		ArtifactPath:        "./artifacts",
```
aaronlevy: This probably should still be configurable.

multi-node/aws/pkg/cluster/config.go
```diff
@@ -3,11 +3,9 @@ package cluster
 import (
 	"errors"
 	"fmt"
+	"gopkg.in/yaml.v2"
```
aaronlevy: Import ordering is off

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -2,13 +2,15 @@ package cluster
...
-	"encoding/base64"
+
 	"fmt"
 	"text/tabwriter"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"io/ioutil"
+	"path/filepath"
```
aaronlevy: import ordering is off

aaronlevy:
Also, just in terms of general notes: It's worrisome that this changes so much and touches zero tests...

colhom:
@aaronlevy w.r.t to the 16kb user-data limit
```
$ tar -cz -O artifacts/manifests/{controller,cluster} artifacts/scripts/install-controller.sh | base64| wc -c
5520
$ tar -cz -O artifacts/manifests/{worker,cluster} artifacts/scripts/install-worker.sh | base64 | wc -c 
4013
```
That is how much space the manifests and scripts take up currently (in bytes).

Here's the user-data sizes on the controller
```
$ wget -q -O- http://169.254.169.254/latest/user-data | wc -c
11347
$ wget -q -O- http://169.254.169.254/latest/user-data | gzip |wc -c
8091
```
and the worker
```
$ wget -q -O- http://169.254.169.254/latest/user-data | wc -c
9333
$ wget -q -O- http://169.254.169.254/latest/user-data | gzip |wc -c
6711
```
We definitely should be gzipping the cloud-config userdata blobs.

It is not ideal, though I think we have enough room to grow for the foreseeable future. 

Eventually we should have ignition support for remote/encrypted files, at which point we should probably switch over to ignition and remote files for everything (ssl certs, manifests, the whole shabang). As of now, the only way we have to pass the certs off to the ec2 machines is via embedding in the user-data blob.

Having kube-aws verify and sync the s3 directories on `up` would be a really good start. Remote syncing and cloudformation creation should be as atomic as possible from the user's standpoint. There should not be two separate steps for that process.

Also, I'm of the opinion that it's way too easy to give two clusters the same s3 URL and have them clobber eachother, so I think each render should generate a random subfolder name and store that in the artifact bundle.

colhom:
I agree that this approach will not work forever. The benefits of this as I see it are:

1. Make `kube-aws up` the only step between cluster artifacts on disk and a running cluster
2. Side-step cloud-config's shortcomings until ignition supports encrypted/verified remote files

We can later patch a much-improved s3 upload component back into `kube-aws up` once we have cloud-config replaced with ignition.

aaronlevy:
What about just building a full cloud-config, then gzipping that in its entirety. It is much easier to reason about and cloud-init will take the gzipped blob (also smaller).

So we would essentially need to insert the raw assets into the cloud-config (where normally they were curled) then just gzip at the end.

1. Start with a directory of assets
1. Render assets into controller/worker/etcd cloud configs (raw text format)
1. Render cloud-formation with (gzipped) controller/worker/etcd cloud configs

colhom:
Templating plain text (newlines and all) into a YAML template is not advisable. New lines and tabs in the file bodies we're templating into the cloud-config will break out of yaml. So, in place of the `#!/bin/bash curl` approach we'd have to base64 encode. At that point it's not raw text, so we may as well be gzipping.

colhom:
We should run a cluster this creates through @aaronlevy 's upcoming e2e tests before we merge this in.

colhom:
fixes #259

multi-node/aws/.gitignore
```diff
@@ -1,4 +1,5 @@
 /artifacts/template.json
 /bin
-/clusters
-/cluster.yaml
+/assets
+*.yaml
```
aaronlevy: Need to be more specific here. `*.yaml` will match any yaml files in the tree.

multi-node/aws/.gitignore
```diff
@@ -1,4 +1,5 @@
...
+*~
```
aaronlevy: Same. Be more explicit with what you want to exclude.

multi-node/aws/artifacts/scripts/install-controller.sh
aaronlevy: Any updates on this?

multi-node/aws/cmd/kube-aws/command_destroy.go
```diff
@@ -38,8 +39,8 @@ func runCmdDestroy(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	clusterDir := path.Join("clusters", cfg.ClusterName)
-	if err := os.RemoveAll(clusterDir); err != nil {
+	credentialsDir := path.Join(rootOpts.AssetDir, "credentials")
+	if err := os.RemoveAll(credentialsDir); err != nil {
```
aaronlevy: Why does destroy just remove the credentials directory and not the cluster dir?

multi-node/aws/.gitignore
colhom: gah that's bad, oops

multi-node/aws/.gitignore
colhom: the *~ is actually what i want there. emacs by default suffixes ~ for autosave files

multi-node/aws/cmd/kube-aws/command_destroy.go
colhom: agreed on that.

multi-node/aws/.gitignore
colhom: The idea is to git-ignore all the random cluster configurations that people will dump in there.

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -18,30 +26,321 @@ var (
...
+	}
+
+	inCfg, err := os.Open(renderOpts.ConfigPath)
+	if err != nil {
+		return err
+	}
+	defer inCfg.Close()
+
+	outCfgPath := filepath.Join(rootOpts.AssetDir, "cluster.yaml")
+	outCfg, err := os.OpenFile(outCfgPath, os.O_WRONLY|os.O_CREATE, 0700)
```
aaronlevy: 0600?

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -18,30 +26,321 @@ var (
...
+	}
+
+	credentialsDir := filepath.Join(rootOpts.AssetDir, "credentials")
+	if err := os.Mkdir(credentialsDir, 0700); err != nil {
+		return fmt.Errorf("Error creating credentials directory %s: %v", credentialsDir, err)
+	}
+
+	if err := initTLS(cfg, credentialsDir); err != nil {
+		stderr("Failed initializing TLS infrastructure: %v", err)
+		os.Exit(1)
```
aaronlevy: Nit: this is inconsistent. The caller will os.Exit anyway when an error is returned.

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -18,30 +26,321 @@ var (
...
+	}
+
+	fmt.Println("Initialized TLS infrastructure")
+
+	kubeconfig, err := newKubeconfig(cfg)
+	if err != nil {
+		stderr("Failed rendering kubeconfig: %v", err)
+		os.Exit(1)
```
aaronlevy: Nit: same as above. Half of this function returns an error and half exits.

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -18,30 +26,321 @@ var (
...
+	}
+
+	kubeconfigPath := path.Join(credentialsDir, "kubeconfig")
+	if err := ioutil.WriteFile(kubeconfigPath, kubeconfig, 0600); err != nil {
+		stderr("Failed writing kubeconfig to %s: %v", kubeconfigPath, err)
+		os.Exit(1)
```
aaronlevy: Nit: same as above.

multi-node/aws/pkg/blobutil/blobutil.go
```diff
@@ -0,0 +1,101 @@
...
+
+func MustTarCompressDirectory(basepath, path string) []byte {
+	buf := &bytes.Buffer{}
+
+	b64Writer := base64.NewEncoder(base64.StdEncoding, buf)
+	defer b64Writer.Close()
+
+	gzWriter, err := gzip.NewWriterLevel(b64Writer, gzip.BestCompression)
+	if err != nil {
+		stderr("Error creating gzip writer: %s", err)
```
aaronlevy: %v

multi-node/aws/pkg/blobutil/blobutil.go
```diff
@@ -0,0 +1,101 @@
...
+			if _, err := io.Copy(tarWriter, f); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}
+
+	if err := filepath.Walk(path, tarHandler); err != nil {
+		stderr("Error tar-ing directory %s: %s", path, err)
```
aaronlevy: %v for err

multi-node/aws/pkg/blobutil/blobutil.go
```diff
@@ -0,0 +1,101 @@
...
+		os.Exit(1)
+	}
+	return buf.Bytes()
+}
+
+func MustReadCompressFile(loc string) []byte {
+	f, err := os.Open(loc)
+	if err != nil {
+		stderr("Failed opening file %s: %s", loc, err)
```
aaronlevy: %v for err

multi-node/aws/pkg/blobutil/blobutil.go
```diff
@@ -0,0 +1,101 @@
...
+
+	gzWriter, err := gzip.NewWriterLevel(b64Writer, gzip.BestCompression)
+	if err != nil {
+		stderr("Failed creating gzip context: %s", err)
+		os.Exit(1)
+	}
+	defer gzWriter.Close()
+
+	if _, err := io.Copy(gzWriter, f); err != nil {
+		stderr("Failed reading file %s: %s", loc, err)
```
aaronlevy: %v for err

multi-node/aws/pkg/cluster/cloud_config_worker.go
```diff
@@ -22,23 +24,29 @@ write_files:
...
-    ARTIFACT_URL={{ ArtifactURL }}
     DNS_SERVICE_IP={{ DNSServiceIP }}
+
 - path: /tmp/install-worker.sh
-  content: |
-    #!/bin/bash
+  encoding: gzip+base64
+  content: {{ InstallWorkerScript }}
+
+- path: /tmp/cluster-manifests.tar
```
aaronlevy: Do workers need this? Seems like it is just the addons + kube-system namespace.

multi-node/aws/pkg/blobutil/blobutil.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	if err := filepath.Walk(path, tarHandler); err != nil {
+		stderr("Error tar-ing directory %s: %s", path, err)
+		os.Exit(1)
+	}
+	return buf.Bytes()
+}
+
+func MustReadCompressFile(loc string) []byte {
```
aaronlevy: nit: MustReadAndCompressFile (I read this as "must read compressed file")

jeffb-stell:
I don't know what the current size of the CloudFormation template is with these changes, but the old template is approximately 27k, and there is a hard limit of 51k template size without uploading to S3.

If template size is going to be > 51k, some documentation for the user that their workflow is going to be
kube-aws render foo.json
aws s3 cp foo.json s3://destination-bucket/
edit cluster.yaml to change artifactURL to https://destination-bucket.s3.amazonaws.com/foo.json
kube-aws up

would be needed

multi-node/aws/pkg/cluster/cloud_config_worker.go
colhom: it's definitely not necessary for the worker

colhom:
@jeffb-stell we're doing away with the artifactURL parameter. For the near future, we expect deployments via this tool to stay within the limits imposed by AWS. A few to be aware of.

* cloudformation template: 51k
* ec2 userdata : 16k
* single cloudformation parameter: 4k

The template is currently ~34k, user-data is numbers are above. 

Going foward, we plan to free ourselves of these constraints. The plan is to move from cloud-config to [ignition](https://github.com/coreos/ignition), which will soon support verifiable/encrypted remote resources. ([open proposal](https://github.com/coreos/bugs/issues/1067))

Verfiable/encrypted remote resources will allow users to encrypt/sign and upload any assets (including TLS secrets). Then, we can provide ways to reference remote blobs from the cloudformation template in a reasonable way- by parameterizing ignition's config with the necessary decryption key(s), urls and expected signatures for the remote resources.

From a UX perspective, the encrypting/signing, uploading, referencing and verification of remote assets is another issue entirely that needs to be worked through.

Remote resources should be detected, encrypted and uploaded automatically on `kube-aws up` step. The manual `s3 upload` step and `artifactURL` parameter, besides being tedious and redundant, introduces the possibility of unintentionally over-writing another active cluster's configuration, and having that cluster pull in the botched config when new machines are created from the cf template. (Autoscaling, for example).

Such an error may not be discovered until you have, say, a newly created kube-worker kubelet that's looking for the api-server at the wrong address because you accidentally used the same `artifactURL` when you deployed a new cluster.

multi-node/aws/pkg/blobutil/blobutil.go
colhom: I'll do the same for MustTarCompressDirectory as well

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -64,7 +93,26 @@ func (c *Cluster) stackName() string {
...
+
+	credentialsDir := filepath.Join(assetDir, "credentials")
+	tlsConfig := NewTLSConfig(credentialsDir)
+	tlsConfig.ReadFilesFromPaths()
+	return tlsConfig
+}
+
+func (c *Cluster) Create(assetDir string) error {
+
+	tlsConfig := c.initAssets(assetDir)
```
aaronlevy: nit: This method seems a little overloaded: read/compress a bunch of different paths then return a random tls struct.

I won't block on this, but maybe we can come up with some more linear workflow. Right now it's somewhat scattered in responsibility.

If "up" expects that all information required is available in ./artifacts/$cluster then this could be as simple as:

```
cluster, err := NewCluster(assetDir)
if err return {err}
cluster.Create()
```

As it stands, "up" requires a clusterConfig which is use to generate an AWS config, then both are used to create a cluster object, which then creates a TLSconfig as part of Create() ...

multi-node/aws/pkg/cluster/config.go
aaronlevy: This still needs to be fixed

multi-node/aws/.gitignore
```diff
@@ -1,4 +1,5 @@
 /artifacts/template.json
 /bin
-/clusters
-/cluster.yaml
+/assets
+/*.yaml
+*~
```
aaronlevy: What specifically are you trying to ignore here? Can we have explicit files instead of * expansion?

colhom: "All files in the tree that end in tilda"
On Jan 25, 2016 6:43 PM, "Aaron Levy" <notifications@github.com> wrote:

In multi-node/aws/.gitignore
<https://github.com/coreos/coreos-kubernetes/pull/240#discussion_r50790070>:

> @@ -1,4 +1,5 @@
>  /artifacts/template.json
>  /bin
> -/clusters
> -/cluster.yaml
> +/assets
> +/*.yaml
> +*~

What specifically are you trying to ignore here? Can we have explicit files
instead of * expansion?

—
Reply to this email directly or view it on GitHub
<https://github.com/coreos/coreos-kubernetes/pull/240/files#r50790070>.

aaronlevy: I understand the regex & I mean for both `/*.yaml` and `*~`.

As far as *~ what files are being generated? This is not something I run into needing to be ignored.
As far as /*.yaml, what yaml files are being generated in the top level dir that you want ignored?

aaronlevy: NM. Just saw your previous comment that it's emacs temp files. That's fine, but still wonder about the *.yaml

aaronlevy: I guess email is better than trying to read inline. Saw your comment about *.yaml. I'd rather not have a wildcard ignores for what a person *might* do.

colhom: I was on BART and the github mobile page is wayyy out it's league with this thread ;)

multi-node/aws/pkg/cluster/cluster.go
colhom: Agreed on that. We have a lot of complexity dangling outside the black box for no good reason.

aaronlevy:
To be completely honest, I'm not feeling very good about this. I understand the goal, and it make sense for a short-term solution, but we're introducing a bunch of complexity / debt for something that should be pretty straight-forward:

    static templates --> variable replacement --> copy to new dir --> assemble cloud-formation

Instead we are adding layers of complexity + no tests. This makes me pretty nervous from a longer term support standpoint.

I won't block on this going in, but I feel like there needs to be some time spent refactoring this tool in the near future.

Before this does go in, squash the cleanup commits, and build a jenkins job for this + conformance tests (can copy the test that is in private jenkins repo and build a new one in parallel until this is ready to land). Let me know when that's ready and I'll take another look.

colhom:
superceded by #277

PR #242, opened 2016-01-12
k8s: Update admission-control list to 1.1 recs

aaronlevy:
NamespaceExists functionality rolled into NamespaceLifecycle.
See: http://kubernetes.io/v1.1/docs/admin/admission-controllers.html


colhom:
Should the use of the admission control plugins be configurable?

Otherwise LGTM

aaronlevy:
It's configurable from the sense the template can be changed by a deployer. There are a lot of nobs that can be moved around. We just want to provide sane defaults, then the configuration can be changed easily enough.

PR #246, opened 2016-01-14
kubelet agent

pbx0:
First stab at the kubelet agent to restart and update the kubelet.


pbx0:
@aaronlevy PTAL I tried to streamline the logic a bit, and rewrote some of the comments. Should have addressed most things. I could use you to double check the logic in getVersionFromNode to make sure i'm actually understand when its ok to signal a temporary error.

pbx0:
Closing this in hopes of taking this work upstream in some form.

jonboulle:
@pbx0 hmm, is there a concrete place to follow along for that? excited to see this, but "hopes" sounds a little vague ;-)

PR #250, opened 2016-01-15
Update openssl.md

drocsid:
https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html descibes the open-ssl config with  

kubernetes.default.svc
kubernetes.default.svc.cluster.local

in the certificate. Furthermore other project use kubernetes.default.svc . See https://github.com/pires/kubernetes-elasticsearch-cluster/issues/27

### DIFF
--- a/Documentation/openssl.md
+++ b/Documentation/openssl.md
@@ -54,6 +54,8 @@ subjectAltName = @alt_names
 [alt_names]
 DNS.1 = kubernetes
 DNS.2 = kubernetes.default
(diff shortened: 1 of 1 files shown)


philips:
This makes sense to me. @colhom @aaronlevy ?

aaronlevy:
lgtm. 

"cluster.local" is configurable, but all of our docs use that, so I'm fine also being explicit in the openssl docs

PR #263, opened 2016-01-28
aws: set flannel backend to vxlan

aaronlevy:


philips:
@aaronlevy did you test it?

aaronlevy:
Need to hold off on this. The conformance tests are a little flaky, but still seeing some suspect failures with this change.

PR #265, opened 2016-01-30
Added ICMP Egress for Worker Nodes

ankitforcode:
1. Added ICMP Egress for Worker Nodes 
2. Added Echo Type ICMP For Controller Ingress and Worker Ingress

### DIFF
--- a/multi-node/aws/pkg/cluster/stack_template.go
+++ b/multi-node/aws/pkg/cluster/stack_template.go
@@ -231,7 +231,7 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
 				map[string]interface{}{"IpProtocol": sgProtoUDP, "FromPort": 0, "ToPort": sgPortMax, "CidrIp": sgAllIPs},
 			},
 			"SecurityGroupIngress": []map[string]interface{}{
-				map[string]interface{}{"IpProtocol": sgProtoICMP, "FromPort": 3, "ToPort": -1, "CidrIp": sgAllIPs},
+				map[string]interface{}{"IpProtocol": sgProtoICMP, "FromPort": 0, "ToPort": -1, "CidrIp": sgAllIPs},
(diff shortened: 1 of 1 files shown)


ankitforcode:
@aaronlevy Please check now and do the needful.

aaronlevy:
Sorry for the delayed response, I was out of office.

This actually changes existing behavior now. The original intention (as I understood it) was to allow outbound ICMP requests.

This PR now removes inbound "destination unreachable" ICMP requests (type 3), and adds "echo" (type 0) ingress to controllers/workers. And also only allows outbound echo requests from workers.

If this satisfies a particular use-case for you, can you provide some additional details? Otherwise the expected change would be:

- Ingress rules do not change
- Egress rules allow for all ICMP requests

ankitforcode:
@aaronlevy The PR does not remove anything as it is "from port"  -  "to port" action on security group. This PR only adds echo "type 0" to the existing Ingress rules.

aaronlevy:
I may be misunderstanding, but the ToPort/FromPort fields have special meaning when specifying for ICMP.

From: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-rule.html

**FromPort:** the start of port range for the TCP and UDP protocols, or an ICMP type number. An ICMP type number of -1 indicates a wildcard (i.e., any ICMP type number).

**ToPort:** The end of port range for the TCP and UDP protocols, or an ICMP code. An ICMP code of -1 indicates a wildcard (i.e., any ICMP code).

The existing ingress rules allowed for a ICMP Type of 3-destination unreachable (FromPort=3) and any ICMP Code (ToPort=-1).

This modifies those rules to only allow ICMP Type 0-echo relpy (FromPort=0) and any ICMP Code (ToPort=-1).

aaronlevy:
Closing due to inactivity. Please let us know if this should be reopened.

PR #268, opened 2016-02-02
Updated: Download the kubectl Executable

idvoretskyi:
For more flexible management of kubectl executable, the procedure steps were changed and updated.


aaronlevy:
Thank you for this proposal, but I'm inclined to leave the instructions as-is. I feel like curl'ing a binary and moving it into place is a well enough understood process.

PR #269, opened 2016-02-02
URL updated to the latest kubectl version

idvoretskyi:


aaronlevy:
All other instructions / assets point to v1.1.2 currently. These will all be updated together when a new release (via coreos) is available. A newer minor version kubectl likely won't cause any issue, but rather not recommend only updating that one piece.

PR #275, opened 2016-02-16
Fix formatting

aknuds1:
I think your markdown needs to be formatted in this way, so that list contents aren't treated as code blocks (and contained formatting is applied, e.g. hyperlinks).

### DIFF
--- a/Documentation/kubernetes-on-aws.md
+++ b/Documentation/kubernetes-on-aws.md
@@ -9,15 +9,15 @@ Choose one of the following paths to get started:
 
 1. `kube-aws`
 
-        Use the `kube-aws` CLI tool to automate cluster deployment. See the [kube-aws Quickstart section](#kube-aws-quickstart) below.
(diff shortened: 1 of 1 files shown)


robszumski:
Ha, looks like crap currently:

image

This broken in https://github.com/aknuds1/coreos-kubernetes/commit/9e6b28144a2b99a682423e685f18cf8663e70d06 which was also trying to fix this formatting. The problem lies with differences in the markdown renders between Github and coreos.com, although they should both be Kramdown AFAIK. I'll be sure to preview this on my local machine this time to make sure we fix it once and for all.

aknuds1:
@robszumski How do I preview coreos.com's rendering? On GitHub my changes looked OK.

robszumski:
You can't, I will do that for you! It should match Github rendering about 95%. There are a few edge cases unfortunately.

aknuds1:
@robszumski :( What would be the right formatting in this case? Would be interested in learning.

robszumski:
Checked this out locally, it looks good. The issue was the previous PR used a tab, which gets interpreted as a code block, instead of 4 spaces, as you used, which will nest the content correctly.

PR #280, opened 2016-02-18
stack template: enforce well-formed strings

meonkeys:
stack template: enforce well-formed strings

Try harder to help users choose sensible, well-formed values for key pair name,
IP addresses, availability zone, and CIDR ranges. I know we're already doing
much of this client-side with `net.ParseCIDR()` and `net.ParseIP()`, but it
can't hurt to have CloudFormation double-check for us, and it's handy if/when
these CloudFormation parameter stanzas are copy-pasted elsewhere.

I tested these changes. CloudFormation errors out as expected if invalid
values are used (with helpful error messages!).

Related CloudFormation "best practices" docs:

1. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/best-practices.html#parmtypes
2. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/best-practices.html#parmconstraints

### DIFF
--- a/multi-node/aws/pkg/cluster/stack_template.go
+++ b/multi-node/aws/pkg/cluster/stack_template.go
@@ -624,73 +624,45 @@ func StackTemplateBody(defaultArtifactURL string) (string, error) {
 	}
 
 	par[parVPCCIDR] = map[string]interface{}{
-		"Type":                  "String",
-		"Default":               DefaultVPCCIDR,
-		"Description":           "CIDR for Kubernetes vpc",
-		"MinLength":             "9",
-		"MaxLength":             "18",
-		"AllowedPattern":        "(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})/(\\d{1,2})",
-		"ConstraintDescription": "must be a valid IP CIDR range of the form x.x.x.x/x.",
+		"Type":        "String",
+		"Default":     DefaultVPCCIDR,
+		"Description": "CIDR for Kubernetes vpc",
 	}
 
 	par[parInstanceCIDR] = map[string]interface{}{
-		"Type":                  "String",
-		"Default":               DefaultInstanceCIDR,
-		"Description":           "CIDR for Kubernetes subnet",
-		"MinLength":             "9",
-		"MaxLength":             "18",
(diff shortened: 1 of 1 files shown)


colhom:
@meonkeys I do think validating the pre-defined types (KeyName and AZ) is a good idea.

The regex's for CIDR/IP are unnecessary and possibly misleading. By only looking at the number of base10 digits, IP addresses and CIDRs can be out of range/malformed and still satisfy the regex. (think 999.999.999.999/99). I think we should trust the .net package to validate those.

meonkeys:
Good call, agreed. Simplified to use only the CloudFormation-specific parameter type checks in c9ef379238cb83bcb58ed6f49abe1ea9763b4ee5

colhom:
@meonkeys I have tested your PR, and using aws region `us-west-1` and not specifying an availability zone. There is a cloudformation parameter validation error:

` Failed creating cluster: Parameter validation failed: parameter value us-east-1 for parameter name AvailabilityZone does not exist`. 

This is odd, because us-east-1 is a region, not an availability zone. I will look into this further to see what the root cause is.

meonkeys:
It fell back to the default availability zone, which is incorrectly `us-west-1`. I'll make that `us-west-1b`, a valid availability zone.

meonkeys:
@colhom fixed in 23c6c9d32016e1110af7b6137f1602d750507fb7

colhom:
@meonkeys thanks. I'll do an e2e test once travis finishes up.

colhom:
@meonkeys the issue is uglier than we though. I'm creating the stack in the us-west-1 region, so us-east-1a is *still not* a valid availability zone according the parameter validation.

There are multiple layers of defaults in this current setup. I think it should be fine if you eliminate the default value for the availabilityZone parameter. If it's not specified, the cf template knows how to look for a valid AZ in the region specified.

meonkeys:
@colhom odd. I don't get it, but like you said, should be fine without _another_ default here.

colhom:
Hi @meonkeys. I think we just can't win with this one
```sh
Failed creating cluster: ValidationError: Parameters: [AvailabilityZone] must have values
```
I'm thinking we may want to just close this one out. I don't think using cf to validate this parameter will work.

meonkeys:
Bummer! Ok, let's roundfile it.

PR #281, opened 2016-02-18
multi-node/aws: existing VPC & private controller

koendc:
Add configuration options to deploy in an existing VPC and subnet
Add a configuration option to deploy the controller private, without
an elastic IP.


multi-node/aws/pkg/config/default_stack_template.go
```diff
@@ -94,6 +94,28 @@ var defaultStackTemplate = `
...
+					]
+				},
+				{
+					"Fn::Equals": [
+						"{{.SubnetID}}",
+						""
+					]
+				}
+			]
+		},
```
colhom: This should this be split up into CreateVPCResources and CreateSubnetResources? As of now, specifying a VPCID but not a SubnetID would be an invalid configuration (and not detected as such until the cloudformation template is submitted to the API). 

In the case that VPCID is specified but not SubnetID, a new subnet should be created in the existing VPC.

PR #284, opened 2016-02-19
Heed deprecation warning

aknuds1:
`go build` emits a deprecation warning if invoking `-X` switch without assignment sign.


aaronlevy:
Looks like it causes go1.4 to fail. We could stop supporting v1.4, or make the flag version aware.

aaronlevy:
This was resolved in https://github.com/coreos/coreos-kubernetes/pull/316

PR #288, opened 2016-02-23
update kubectl to version 1.1.7

kbrwn:
Updates kubectl from v1.1.2 to v1.1.7. This will prevent users who follow our documentation to launch their k8s cluster from getting the error below when attempting to launch Tectonic Console:

```
 $ kubectl get pods -l tectonic-app=console -o name | sed s:^pod/:: | xargs -i{} kubectl port-forward {} 9000
error: output format "name" not recognized
```

### DIFF
--- a/Documentation/configure-kubectl.md
+++ b/Documentation/configure-kubectl.md
@@ -9,13 +9,13 @@ Download `kubectl` from the Kubernetes release artifact site with the `curl` too
 The linux `kubectl` binary can be fetched with a command like:
 
 ```sh
-$ curl -O https://storage.googleapis.com/kubernetes-release/release/v1.1.2/bin/linux/amd64/kubectl
+$ curl -O https://storage.googleapis.com/kubernetes-release/release/v1.1.7/bin/linux/amd64/kubectl
 ```
 
 On an OS X workstation, replace `linux` in the URL above with `darwin`:
 
 ```sh
-$ curl -O https://storage.googleapis.com/kubernetes-release/release/v1.1.2/bin/darwin/amd64/kubectl
+$ curl -O https://storage.googleapis.com/kubernetes-release/release/v1.1.7/bin/darwin/amd64/kubectl
 ```
 
 After downloading the binary, ensure it is executable and move it into your `PATH`:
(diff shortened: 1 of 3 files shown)


robszumski:
I don't think we're ready to do this. Can you use a newer kubectl against an older cluster?

I had thought you were gonna rewrite the script to work with 1.1.2.

joeatwork:
Rob - I've been using a kubectl that works with the script and also works with coreos-kubernetes - that leads me to suspect that we're ok with the upgrade.

robszumski:
@joeatwork define "work"? Do things like log streaming, exec work? If so, I guess we can go for it. 

@aaronlevy anything to add regarding compatibility?

joeatwork:
"Work" meaning everything I can do seems to be fine - it's the only kubectl I use, and I use it regularly to manage coreos-kubernetes clusters (including exec and log-following) but I haven't made a systematic test or anything.

aaronlevy:
The version skew of kubectl v1.1.7 against a v1.1.2 api-server will probably be fine, but I haven't personally tested it.

But either way, rather than piping into sed can probably just use built-in templating:

`kubectl get pods -l foo=bar -o template --template="{{range.items}}{{.metadata.name}}, {{end}}"`

We're going to have to support multiple versions of kubernetes no matter what, so there may end up needing to be version specific examples. The `-o template` should work for both v1.1.2 and 1.1.7 fwiw (so I would prefer we update the instructions rather than using different version of kubectl).

robszumski:
@joe or @kbrwn can you test that out really quick?

idvoretskyi:
We already had the similar pull request - https://github.com/coreos/coreos-kubernetes/pull/269

kbrwn:
worked for v1.1.2
```
kubectl get pods -l tectonic-app=console -o template --template="{{range.items}}{{.metadata.name}}{{end}}" | xargs -i{} kubectl port-forward {} 9000
```

robszumski:
@idvoretskyi we'll update to a newer kubectl in lock step with a newer version of Kubernetes, which should be soon. Stay tuned!

@kbrwn Great, we can go with that for now. Thanks for testing!

PR #298, opened 2016-03-02
Added releaseChannel to cluster.yaml.example

four43:
Release channel is nice to configure and from what I can tell is already supported.

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -517,7 +517,7 @@ func stringInSlice(a string, list []string) bool {
 	for _, b := range list {
 		if b == a {
 			return true
- 		}
+		}
 	}
 	return false
 }
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -363,9 +363,9 @@ func (cfg Cluster) valid() error {
 	if cfg.KMSKeyARN == "" {
 		return errors.New("kmsKeyArn must be set")
 	}
-	var supportedReleaseChannels = []string{"alpha","beta"}
+	var supportedReleaseChannels = []string{"alpha", "beta"}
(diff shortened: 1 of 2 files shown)


aaronlevy:
Unfortunately, we should only be supporting the alpha channel at the moment. The reason is that we are not using a rkt-based deployment for the kubelet, and the helper script ( [kubelet-wrapper](https://coreos.com/kubernetes/docs/latest/kubelet-wrapper.html) ) is only available in 962.0.0+ (currently just alpha).

You can manually add the kubelet-wrapper to the host, but this tool assumes it is available. So we should probably wait to make this configurable until the wrapper is available in other channels.

aaronlevy:
The beta channel is now supported, so exposing this configurable would make more sense now. @four43 if you're still in need of this, would you be able to update this PR?

four43:
Sure can, I'll see what I can put together

-Seth

four43:
I just rebased this and updated it. It isn't much... It just exposes that variable in the example yaml file. Are there other places I should be looking to implement this feature? 

Thanks guys, 
-Seth

colhom:
@four43 sorry for the delay. Can you add a few more things to this PR:
* in pkg/config/config.go define  `var supportedReleaseChannels = []string{"alpha","beta"}

* Check that cluster.ReleaseChannel is in `supportedReleaseChannels` in [cluster.valid()](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/config.go#L328)

* Add a tests to `pkg/config/config_test.go` that tests: 
  * accepting valid release channel
  * rejecting unsupported release channel

Thanks!

four43:
Strugglin hard over here guys lol. I'm not a golang guy, but I'm hacking my way through it here.

four43:
I'm gonna tag out for the test cases, that looks over my head! @colhom sorry man!

colhom:
@four43 thanks for trying. I have #427 up now!

PR #301, opened 2016-03-04
Attempt to improve formatting for commands

hkjn:
The page looks fine in the previous state here on GitHub, but on the coreos.com site the commands inside the triple backticks get bunched up on the same line:
https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html

E.g: `[default] aws_access_key_id = AKID1234567890 aws_secret_access_key = MY-SECRET-KEY`.

I'm not sure what causes the difference in how the markdown is rendered on Github vs on the coreos.com site, so if someone more clued in could review or improve the change that'd be great, but wanted to call attention to the issue and try to improve things.

### DIFF
--- a/Documentation/kubernetes-on-aws.md
+++ b/Documentation/kubernetes-on-aws.md
@@ -36,24 +36,24 @@ $ tar -xf <file> kube-aws
 
 Configure your local workstation with AWS credentials using one of the following methods:
 
-1. Environment Variables
+#### Method 1: Environment Variables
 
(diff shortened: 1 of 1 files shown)


robszumski:
These nested lists have been giving us a ton of trouble. Moving these into a header style vs a list is a good idea.

Would you mind doing the same to the other list a few lines up in the doc? This one: https://github.com/coreos/coreos-kubernetes/pull/301/files#diff-9019fab2ed3a020e67f912ba2482cb59R10

aaronlevy:
@robszumski what is state of this? Should this be merged or still waiting on work?

robszumski:
Completed this work in https://github.com/coreos/coreos-kubernetes/pull/309

Issue #94, opened 2015-10-03
NoCredentialProviders: no valid providers in chain


rclayton-the-terrible:
"Failed creating cluster: NoCredentialProviders: no valid providers in chain" - keep getting this error.  Noticed that there was a similar issue registered in the AWS GO SDK.

aaronlevy:
@rclayton-the-terrible are you still seeing this issue?

rclayton-the-terrible:
@aaronlevy Honestly, I've stop trying.  The closest I got to a full Kubernetes deployment on AWS was mirroring Kelsey's Terraform scripts.  Since then, I've gotten really busy in development and haven't put any effort into it.

aaronlevy:
@rclayton-the-terrible if you get some time to come back to this, and you're still having issues, please let us know. I'm gonna close this for now, but feel free to re-open.

Also, just as a note, a quick way to get started deploying a cluster to AWS should be using the defaults in our quickstart guide: https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html#kube-aws-quickstart

munhitsu:
It might be related to: https://github.com/aws/aws-sdk-go/issues/345

jasdel:
@rclayton-the-terrible Just saw this was linked to the SDK. It looks like coreos-kubernetes is using a fairly old version of the SDK (0.9.5)  But I've verified that this version of the SDK should be returning the configured role credential.

You can run the following curl commands on the EC2 host to see what error the SDK is running into.

```
# Get the role name
curl http://169.254.169.254//latest/meta-data/iam/security-credentials/

# Use role name to get credentials
curl http://169.254.169.254//latest/meta-data/iam/security-credentials/<my-ec2-iam-role-name>
```

The AWS SDK for Go does have aws/aws-sdk-go#413 to improve this error message.

kbrwn:
I ran into this issue yesterday. Using the latest version of kube-aws and my credential file is set up correctly.

four43:
+1 having this issue as well. Anything we can do to help debug?

aymanosman:
I also experienced the error.

```
Initialized TLS infrastructure
Wrote kubeconfig to ...
Waiting for cluster creation...
Failed creating cluster: NoCredentialProviders: no valid providers in chain
```

I was using `export AWS_DEFAULT_PROFILE=foo`

[...]

PR #303, opened 2016-03-10
kube-aws: better support for AWS credentials file

bcwaldon:
Using the SharedCredentials codepath allows the user to better
control what AWS credentials are used. Specifically, the ability
to set AWS_PROFILE=<profile> is desireable so that users do not
have to export credentials into environment variables, and are
no longer required to set credentials to the "default" profile
in ~/.aws/credentials.

### DIFF
--- a/multi-node/aws/cmd/kube-aws/main.go
+++ b/multi-node/aws/cmd/kube-aws/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/spf13/cobra"
 
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/cluster"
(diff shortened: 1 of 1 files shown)


spacepluk:
thanks, I've been testing this and it makes my life a lot easier :)

aaronlevy:
@colhom

colhom:
Running this through the conformance tests now.

colhom:
The conformance test reveals that this breaks the "old way" of doing things (manually exporting AWS_ACCESS_KEY_ID etc, not using ~/.aws/credentials)

```shell
Failed creating cluster: SharedCredsLoad: failed to load shared credentials file
caused by: open /home/jenkins/.aws/credentials: no such file or directory
```
@bcwaldon hello again! Is there anyway we can make this compatible with both approaches?

aymanosman:
You can already do this. `AWS_PROFILE=myprofile kube-aws up` works for me. Unless I've misunderstood something.

cgag:
This already seems to work as far as I can tell, as @aymanosman said.  I just successfully ran `kube-aws validate` with AWS_PROFILE=<myprofile> and again with the creds exposted as environment variables.

aaronlevy:
Let us know if the issue isn't resolved and should be re-opened. Also closed related issue: https://github.com/coreos/coreos-kubernetes/issues/94

bcwaldon:
I'm not sure what I experienced to make me think this was needed, but you're right, setting AWS_PROFILE already works. Sorry for the noise!

PR #309, opened 2016-03-15
docs: convert to paragraph from list

robszumski:
Completes/fixes https://github.com/coreos/coreos-kubernetes/pull/301

cc: @aaronlevy


aaronlevy:
lgtm

Issue #308, opened 2016-03-14
conformance-tests.md uses removed branch of kubernetes git repo


rutsky:
[Here](https://coreos.com/kubernetes/docs/latest/conformance-tests.html#checkout-conformance-test-branch):

```
$ cd kubernetes
$ git checkout conformance-test-v1
```

however `https://github.com/kubernetes/kubernetes` doesn't have `conformance-test-v1` branch.

Looks like this branch was [removed](https://github.com/kubernetes/kubernetes/issues/17536).

aaronlevy:
Yeah, conformance tests were changed to be run directly off the release you wish to test. The instructions need to be updated. Feel free to submit a PR, or we can get these updated soon.

/cc @robszumski

PR #310, opened 2016-03-15
docs: update conformance instructions

robszumski:
Fixes https://github.com/coreos/coreos-kubernetes/issues/308

cc: @aaronlevy @rutsky

### DIFF
--- a/Documentation/conformance-tests.md
+++ b/Documentation/conformance-tests.md
@@ -27,7 +27,7 @@ Next, checkout the branch or release you'd like to test against:
 
 ```sh
 $ cd kubernetes
-$ git checkout release-1.0
+$ git checkout v1.1.8
 ```
 
 ### Create Kubernetes Binaries
--- a/Documentation/conformance-tests.md
+++ b/Documentation/conformance-tests.md
@@ -2,41 +2,52 @@
 
 The conformance test checks that a kubernetes installation supports a minimum required feature set.
 
-## Kubernetes Cluster
+## Prerequisites
 
-You will need to have a cluster already running, and a kubeconfig with the current-context set to the cluster you wish to test.
+You will need to have a Kubernetes cluster already running, and a kubeconfig with the current-context set to the cluster you wish to test.
 
 Make sure only the essential pods are running, and there are no failed or pending pods. If you are testing a small / development cluster, you may need to increase the node memory or tests could inconsistently fail (e.g. 2048mb for single-node vagrant).
 
-## Clone Kubernetes
+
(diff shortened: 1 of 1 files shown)


Documentation/conformance-tests.md
```diff
@@ -2,41 +2,52 @@
...
 
-## Checkout Conformance Test branch
+### Checkout Branch
+
+Next, checkout the branch or release you'd like to test against:
 
 ```sh
 $ cd kubernetes
-$ git checkout conformance-test-v1
+$ git checkout release-1.0
```
rutsky: Why 1.0 release?

robszumski: I figured we had to point you towards something, and if we prodded you towards a release rather than master it would be helpful. What better than the 1.0!

If you have suggestions, I'm open to it.

rutsky: It would be nice to specify latest stable release here, however there is no such thing as "stable" branch in Kubernetes git.

Maybe specify current latest stable release, e.g. `release-1.1`?

robszumski: I considered that, but its just going to get out of date when they release a new version. I figured the 1.0 was a nice enough one to suggest that won't look out of date. For example, 1.2 will be out any day now...

aaronlevy: In our examples we should use the tagged releases, e.g. `v1.1.8`. The release branches will work, but generally people will be building their clusters off a tagged release. Maybe we just document `vX.Y.Z` to keep it generic?

rutsky: It's better to provide specific version, IMO, then generic vX.Y.Z. Your phrase `checkout the branch or release you'd like to test against` already specifies, that version can be changed.

I would suggest to use specific release tag, so that user will see that this is Kubernetes version, like `v1.1.8` in other places in documentation (since user usually already knows which version of Kubernetes he is using).

aaronlevy: I think we are going to have to target the docs to a particular release because the conformance tests themselves change. For example NUM_MINIONS becomes NUM_NODES between 1.1 and 1.2. So while our targeted release is v1.1.8, our docs should probably reflect that. Assuming that, what about the following changes:

Heading: "Checkout Release"
Code: `git checkout v1.1.8`

Then after the bump to 1.2.x we will have to update the instructions anyway.

rutsky: Sounds good, thanks for such detailed consideration of this issue!

aaronlevy:
lgtm

PR #277, opened 2016-02-17
Major refactor of kube-aws tool

colhom:
* kube-aws can now generate all assets/templates. We can now distribute just the binary.

* No S3 intermediary. All assets are compressed on `kube-aws up` and sent directly to cloudformation api.

* `kube-aws render` generates a reproducible cluster asset pack which contains tls assets and all templates

* Much simpler templating pipeline. Single templating phase on `kube-aws up`.

* New `validate` command to validate user-data and stack json

* Replace confusing cloudformation parameters `kube-aws up --update`, which will update the cloudformation stack.

* Unit tests for all config modules

* Replace `godeps` with `glide`!

PR #316, opened 2016-03-16
kube-aws refactor

aaronlevy:
This is just a cleanup of #277 to make reviewing a little easier.


.travis.yml
```diff
@@ -1,12 +1,15 @@
 language: go
 go:
-  - 1.4
   - 1.5
 install:
+  - export GO15VENDOREXPERIMENT=1
+  - wget https://github.com/Masterminds/glide/releases/download/0.8.3/glide-0.8.3-linux-amd64.tar.gz
+  - tar zxvf glide-0.8.3-linux-amd64.tar.gz
```
aaronlevy: I don't think glide is a requirement anymore

.travis.yml
```diff
@@ -1,12 +1,15 @@
...
+  - export PATH="${PATH}:$(pwd)/linux-amd64"
   - go get golang.org/x/tools/cmd/vet
 script:
   - cd $TRAVIS_BUILD_DIR/multi-node/aws
-  - GOPATH=${PWD}/Godeps/_workspace:${GOPATH} go test -v ./...
-  - test -z "$(find . -path ./Godeps -prune -o -name '*.go' -exec gofmt -d {} + | tee /dev/stderr)"
-  - go vet ./...
+  - test -z "$(find . -prune -o -name '*.go' -exec gofmt -d {} + | tee /dev/stderr)"
```
aaronlevy: I don't think this will work anymore. Instead I think we should just change this to pruning the `./vendor` path. The output of the find command should be all the .go files we want to pass into gofmt.

.travis.yml
```diff
@@ -1,12 +1,15 @@
...
   - ./build
+  - go test -v $(find ./{cmd,pkg} -maxdepth 1 -mindepth 1 -type d)
+  - go vet $(find ./{cmd,pkg} -maxdepth 1 -mindepth 1 -type d)
```
aaronlevy: Curious why this would need to change from `./...` to find/maxdepth ?

multi-node/aws/README.md
```diff
@@ -5,24 +5,86 @@ View the full instructions at https://coreos.com/kubernetes/docs/latest/kubernet
 
 ## Development
 
+### Download pre-built binary
+
+```sh
+wget https://<binary-url>
+# check checksum
+chmod +x ./kube-aws
+sudo mv kube-aws /usr/bin/
```
aaronlevy: nit: I won't block on this, but I dislike that our instructions would require sudo for this tool. I think this could be mitigated by allowing and `--output` flag or something of the sort, so we don't rely on running the tool in the directory we wish to act on.

Or another option would be that our instructions say something like `alias kube-aws='/path/to/kube-aws'`.

Or if left as-is, this should probably be /usr/local/bin

multi-node/aws/README.md
```diff
@@ -5,24 +5,86 @@ View the full instructions at https://coreos.com/kubernetes/docs/latest/kubernet
...
-### Custom Kubernetes Manifests
+## Initialize an asset directory
+```sh
+$ mkdir my-cluster
+$ cd ./my-cluster
+$ kube-aws init --cluster-name=my-cluster-name --external-dns-name=my-cluster-endpoint --region=us-west-1 --availability-zone=us-west-1c --key-name=key-pair-name
+```
 
-You may deploy a cluster using a custom CloudFormation template, Kubernetes manifests and install scripts using the `artifactURL` option in your cluster config.
+There will now be a ./cluster.yaml file in the asset directory.
```
aaronlevy: nit: drop the "./" prefixes. Considering this is for human consumption:

"there will be a `cluster.yaml` file in the asset directory" VS
"there will be a `./cluster.yaml` file in the asset directory"

multi-node/aws/README.md
```diff
@@ -5,24 +5,86 @@ View the full instructions at https://coreos.com/kubernetes/docs/latest/kubernet
...
+```sh
+$ kube-aws render
 ```
-$ kube-aws render --output=artifacts/template.json
-$ aws s3 cp --recursive --acl=public-read artifacts/ s3://<bucket>/
+This generates the default set of cluster assets in your asset directory. These assets are templates and credentials that are used to create, update and interact with your Kubernetes cluster.
+
+You can now customize your cluster by editing asset files:
+
+* **./cluster.yaml**
```
aaronlevy: nit: same as above, would suggest dropping the "./" prefixes. Here and in headers below

multi-node/aws/build
```diff
@@ -14,13 +14,17 @@ if [ -n "$(git diff --shortstat 2> /dev/null | tail -n1)" ]; then
...
-rm -rf ./gopath
-link_path="${PWD}/gopath/src/github.com/coreos"
-mkdir -p $link_path
-ln -s ${PWD}/../../ "${link_path}/coreos-kubernetes"
+export GOPATH="$(pwd)/_output"
+export KUBEAWS="$GOPATH/src/github.com/coreos/coreos-kubernetes/multi-node/aws"
 
-GOPATH=${PWD}/Godeps/_workspace:${PWD}/gopath CGO_ENABLED=0 go build -ldflags "-X main.VERSION ${VERSION}" -a -tags netgo -installsuffix netgo -o bin/kube-aws github.com/coreos/coreos-kubernetes/multi-node/aws/cmd/kube-aws
-rm -r ./gopath
+if [ -d $GOAPTH ]; then
```
aaronlevy: typo: GOPATH
Also, this check is redundant for `rm -rf`

multi-node/aws/build
```diff
@@ -14,13 +14,17 @@ if [ -n "$(git diff --shortstat 2> /dev/null | tail -n1)" ]; then
...
+    rm -rf $GOPATH
+fi
+
+mkdir -p $GOPATH/src
+cp -r ./vendor/* $GOPATH/src
+mkdir -p $KUBEAWS
+cp -r ./pkg $KUBEAWS/pkg
+cp -r ./cmd $KUBEAWS/cmd
```
aaronlevy: nit: this build dir could be lighter weight using symlinks rather than copying everything around

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -28,225 +34,154 @@ func (c *ClusterInfo) String() string {
...
+			continue
+		default:
+			return "", fmt.Errorf("unexpected stack status: %s", statusString)
+		}
 	}
+}
 
-	info.Name = c.cfg.ClusterName
-	return info, nil
+//TODO: validate cluster
```
aaronlevy: nit: add name on TODOs. Also some more context would be helpful

multi-node/aws/pkg/config/config.go
```diff
@@ -0,0 +1,326 @@
...
+func newDefaultCluster() *Cluster {
+	return &Cluster{
+		ClusterName:              "kubernetes",
+		ReleaseChannel:           "alpha",
+		VPCCIDR:                  "10.0.0.0/16",
+		InstanceCIDR:             "10.0.0.0/24",
+		ControllerIP:             "10.0.0.50",
+		PodCIDR:                  "10.2.0.0/16",
+		ServiceCIDR:              "10.3.0.0/24",
+		KubernetesServiceIP:      "10.3.0.1",
```
aaronlevy: This shouldn't be a configurable - it should be derived from the first ip in ServiceCIDR. I won't block on this, but we need to remove this as a configurable.

multi-node/aws/pkg/config/config.go
```diff
@@ -0,0 +1,326 @@
...
+		DNSServiceIP:             "10.3.0.10",
+		K8sVer:                   "v1.1.4",
```
aaronlevy: I know that using the coreos hyperkube images is still a TODO with this PR - but just noting here this version will need to be updated.

multi-node/aws/pkg/config/config.go
```diff
@@ -0,0 +1,326 @@
...
+		return fmt.Errorf("serviceCIDR (%s) overlaps with podCIDR (%s)", cfg.ServiceCIDR, cfg.PodCIDR)
+	}
+
+	kubernetesServiceIPAddr := net.ParseIP(cfg.KubernetesServiceIP)
+	if kubernetesServiceIPAddr == nil {
+		return fmt.Errorf("Invalid kubernetesServiceIP: %s", cfg.KubernetesServiceIP)
+	}
+	if !serviceNet.Contains(kubernetesServiceIPAddr) {
+		return fmt.Errorf("serviceCIDR (%s) does not contain kubernetesServiceIP (%s)", cfg.ServiceCIDR, cfg.KubernetesServiceIP)
+	}
```
aaronlevy: Add a test case for kubernetesServiceIPAddr == first ip in ServiceCIDR. Either that or we should remove it as a configurable. Whichever you find easier.

multi-node/aws/pkg/config/config.go
```diff
@@ -0,0 +1,326 @@
...
+
+	dnsServiceIPAddr := net.ParseIP(cfg.DNSServiceIP)
+	if dnsServiceIPAddr == nil {
+		return fmt.Errorf("Invalid dnsServiceIP: %s", cfg.DNSServiceIP)
+	}
+	if !serviceNet.Contains(dnsServiceIPAddr) {
+		return fmt.Errorf("serviceCIDR (%s) does not contain dnsServiceIP (%s)", cfg.ServiceCIDR, cfg.DNSServiceIP)
+	}
```
aaronlevy: Also should add a test that dnsServiceIPAddr != first ip in serviceCIDR and != kubernetesServiceIPAddr

multi-node/aws/pkg/config/region_map.go
```diff
@@ -0,0 +1,62 @@
...
+	"sa-east-1",
+	"us-east-1",
+	"us-gov-west-1",
+	"us-west-1",
+	"us-west-2",
+}
+
+var supportedChannels = []string{
+	"alpha",
+	"beta",
```
aaronlevy: beta will not be supported once this PR is updated to include the kubelet-wrapper

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -1,22 +1,25 @@
...
-# clusterName: kubernetes
+
+clusterName: {{.ClusterName}}
+
+# DNS name routable to the Kubernetes controller nodes
+# from worker nodes and external clients. The deployer
+# is responsible for making this name routable
+externalDNSName: {{.ExternalDNSName}}
+
+### DO NOT CHANGE PARAMETERS ABOVE HERE ###
```
aaronlevy: nit: maybe move them below a "Do not change below". Little weird to read about configurables, then see a comment saying I shouldn't change them

aaronlevy:
Do we actually want to commit `multi-node/aws/pkg/config/templates.go`? Could generating this instead just be part of the build process?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -83,6 +100,19 @@ write_files:
...
 
+  - path: /opt/bin/decrypt-tls-assets
+    owner: root:root
+    permissions: 0700
+    content: |
+      #!/bin/bash -e
+
+      for encKey in $(find /etc/kubernetes/ssl/*.pem);do
+        tmpPath="/tmp/$(basename $encKey).tmp"
+        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext | base64 --decode > $tmpPath
```
aaronlevy: It looks like you're overwriting the assets in-place. Is this safe to run multiple times on the same assets (e.g. after reboot)?

multi-node/aws/cmd/kube-aws/command_init.go
```diff
@@ -29,6 +29,7 @@ func init() {
 	cmdInit.Flags().StringVar(&initOpts.Region, "region", "", "The aws region to deploy to")
 	cmdInit.Flags().StringVar(&initOpts.AvailabilityZone, "availability-zone", "", "The aws availability-zone to deploy to")
 	cmdInit.Flags().StringVar(&initOpts.KeyName, "key-name", "", "AWS key-pair for ssh access to nodes")
+	cmdInit.Flags().StringVar(&initOpts.KMSKeyARN, "kms-key-arn", "", "AWS key-pair for ssh access to nodes")
```
aaronlevy: help text is copy/paste

multi-node/aws/cmd/kube-aws/command_init.go
```diff
@@ -41,6 +42,7 @@ func runCmdInit(cmd *cobra.Command, args []string) error {
 		{"--region", initOpts.Region},
 		{"--availability-zone", initOpts.AvailabilityZone},
 		{"--key-name", initOpts.KeyName},
+		{"--kms-key-arn", initOpts.KMSKeyARN},
```
aaronlevy: If this is required, I think we need documentation (at minimum pointing to aws docs) of what an end-user needs to configure prior to using this tool. The ssh-key is well-understood (and required, iirc) to launch new nodes, but setting up kms keys is not as well-understood or required to launch nodes.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Cloud-config re-applies the initial filesystem state on every reboot prior that service script running.

multi-node/aws/pkg/config/config.go
```diff
@@ -243,6 +242,18 @@ type Config struct {
 	TLSConfig *CompactTLSAssets
 }
 
+//This functionality now requires AWS API access
+//Separated out from cluster.Config() so that we can only compact
+//assets on command=up
```
aaronlevy: nit: this is information that should probably be in the commit message. This will quickly lose value as a comment (when is "now")

multi-node/aws/pkg/config/config.go
```diff
@@ -243,6 +242,18 @@ type Config struct {
...
+func (c *Config) CompactTLSAssets(tlsConfig *RawTLSAssets) error {
```
aaronlevy: The function name is the same as the type. This could get confusing. It also doesn't need to be exported.

colhom:
@ericchiang, should we be checking in `templates.go` or have it be part of the build process?

multi-node/aws/pkg/config/tls_config.go
```diff
@@ -170,15 +174,32 @@ func compressData(d []byte) (string, error) {
...
+	awsConfig = awsConfig.WithRegion(cfg.Region)
+	kmsSvc := kms.New(session.New(awsConfig))
+
 	var err error
 	compact := func(data []byte) string {
 		if err != nil {
 			return ""
 		}
+
+		if cfg.KMSKeyARN != "" {
```
aaronlevy: If this is a required parameter, why do we check this?

ericchiang:
@colhom check it in

aaronlevy:
@ericchiang humor me with details? The file itself is just a bunch of bytes. So any change to the template is going to be an opaque 2000+ line change to commit?

See the rkt-fly change in this PR: https://github.com/coreos/coreos-kubernetes/pull/316/commits/d0d332d7ab03adb1b142de40b214ecba188d15b8

ericchiang:
I usually commit generated code (all be it in a different commit), but yeah
you could easily .gitignore it and add the go generate command to the build
script.

If it helps with the code review by all means omit it.

multi-node/aws/pkg/config/tls_config.go
colhom: This is a hack so that we can still assemble/validate the cloud-config/stack template files if there are not aws credentials available. For unit testing without aws api access.

aaronlevy:
I'd lean toward having it be part of the build script (or makefile) and not committed directly. We are going to update those templates fairly frequently. But I could go either way.

I don't really see this being used as a package imported by other projects, so the issue of downstream users needing to run go generate to use this seems not very likely. Most of the time people are going to be using the release-binaries anyway.

multi-node/aws/pkg/config/config.go
colhom: This is a pattern introduced by @ericchiang. 

Also used here: https://github.com/coreos/coreos-kubernetes/blob/aws-rewrite/multi-node/aws/pkg/config/config.go#L97

multi-node/aws/pkg/config/templates/cluster.yaml
colhom: Yeah agree. Honestly that is really only relevant if you're going to be attempting to upgrade your cluster, which we are officially not supporting yet for end users. I'll remove it for now.

multi-node/aws/pkg/config/config.go
ericchiang: This originally returned the type rather than assigning to a struct field. Yes it shouldn't be exported.

It might even be better just to remove the function entirely and run

```
compact, err := tlsConfig.Compact(c)
 if err != nil {
 	return fmt.Errorf("failed to compress TLS assets: %v", err)
 }
 c.TLSConfig = compact
```

when you need the TLSConfig. I don't see much value in having a helper function that does that. It's unclear when it should be called.

ericchiang:
> I don't really see this being used as a package imported by other projects, so the issue of downstream users needing to run go generate to use this seems not very likely. Most of the time people are going to be using the release-binaries anyway.

On that note, we should probably name the packages `internal` so no one can import them. (Can be done in a follow up PR).

multi-node/aws/pkg/config/config.go
ericchiang: Let's add a TODO to the code here

aaronlevy:
Agreed. Also in the spirit of moving this forward, we can separate the generated code into the build process in a follow up PR as well.

.travis.yml
colhom: Using `./...` attempts to vet/test all the vendored dependencies.

aaronlevy: k. We could probably make this more generic with using `-path ./vendor -prune -o` similar to the gofmt test above. restricting by specific name/depth can lead toward untested files / packages and requires us to always be aware that this list needs to be updated.

ericchiang:
```
go vet $(go list ./... | grep -v '/vendor/')
```

aaronlevy: even better

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -37,7 +37,12 @@ coreos:
       enable: true
       content: |
         [Service]
-        ExecStart=/usr/bin/kubelet \
+        ExecStartPre=/usr/bin/mkdir -p /var/log/containers
+        ExecStartPre=/usr/bin/mkdir -p /var/log/fluentd-position
+        Environment="RKT_OPTS=--volume var-log-fluentd-position,kind=host,source=/var/log/fluentd-position --mount volume=var-log-fluentd-position,target=/var/log/fluentd-position --volume var-log-containers,kind=host,source=/var/log/containers --mount volume=var-log-containers,target=/var/log/containers"
```
aaronlevy: This tool doesn't configure fluentd (afaik) so I would lean toward leaving this up to the deployer rather than deciding for them.

multi-node/aws/pkg/config/tls_config.go
aaronlevy: Can you add a todo/note here about why we do this

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: discussed in meatspace. Lets' back this out for now and follow up mounting /var/log/containers in a follow-up PR

colhom:
@robszumski could you take a look at this? Review items addressed, docs are ready to go.

\cc @aaronlevy

aaronlevy:
Can you back out the kubernetesServiceIP changes -- that will need more tests written and I'd like to get this merged without adding more to review / test. Just leave it as a TODO right now.

colhom:
Done. I'll do that as a PR

robszumski:
Don't these docs need to be reproduced on https://github.com/coreos/coreos-kubernetes/blob/aws-rewrite/Documentation/kubernetes-on-aws.md#kube-aws-quickstart, which is what the users will actually follow?

colhom:
@robszumski I would assume so. Do you want to use the README markdown for that?

robszumski:
@colhom yeah I think you can copy the bits that make sense and tweak it slightly and its good to go

colhom:
@robszumski I kind of did a merge between the old content and the new stuff.

joshix:
not super easy to review, but LGTM.

PTBF.

colhom:
Conformance tests passed, we're merging.

In case anyone's wondering, PTBF = push the button frank

PR #319, opened 2016-03-16
Fixed Curl commands to avoid breakage in 1.2

jbowen93:
per [this issue](https://github.com/kubernetes/kubernetes/issues/22904)

### DIFF
--- a/multi-node/generic/controller-install.sh
+++ b/multi-node/generic/controller-install.sh
@@ -774,14 +774,14 @@ function start_addons {
     done
     echo
     echo "K8S: kube-system namespace"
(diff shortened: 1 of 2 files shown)


aaronlevy:
@jbowen93 thanks for this!

Would you mind also updating the aws scripts to also reflect this change: 
https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-controller
https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-worker

jbowen93:
https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-worker looks to be free of curl command so it should be fine. I'll try to fix the 4 commands I'm seeing in the cloud-config-controller.

aaronlevy:
closed in: https://github.com/coreos/coreos-kubernetes/pull/341

PR #323, opened 2016-03-17
Fix broken link in manual set up doc

tomdee:


Documentation/deploy-addons.md
```diff
@@ -157,5 +157,5 @@ $ kubectl get pods --namespace=kube-system | grep kube-dns-v9
 <div class="co-m-docs-next-step">
   <p>Now that you have a working Kubernetes cluster with a functional CLI tool, you are free to deploy Kubernetes-ready applications.</p>
   <p>Start with a multi-tier web application (Guestbook) from the official Kubernetes documentation to visualize how the various Kubernetes components fit together.</p>
-  <a href="http://kubernetes.io/v1.1/examples/guestbook-go/README.html" class="btn btn-default btn-icon-right" data-category="Docs Next" data-event="kubernetes.io: Guestbook">Deploy the Guestbook Sample app</a>
+  <a href="https://github.com/kubernetes/kubernetes/tree/release-1.1/examples/guestbook-go/" class="btn btn-default btn-icon-right" data-category="Docs Next" data-event="kubernetes.io: Guestbook">Deploy the Guestbook Sample app</a>
```
robszumski: I'd rather not link to a specific release. Maybe we should swap this entire thing out for http://kubernetes.io/docs/hellonode/?

"Deploy a Helloworld app" or similar

aaronlevy: I think we should probably merge this as-is:

The hellonode doc is for creating an app from scratch (and some steps assume a GCE account). The "Kubernetes 101" link in the sidebar ends up linking to the versioned Guestbook example in github (like we have here).

We could link both, e.g. "example building an application from scratch" or "deploy an example guestbook app". But the from-scratch approach alone is going to be more difficult for first time users.

As far as versioning, in some cases we're going to need to keep it tied to our currently targeted version (1.1.X at time of writing).

For example, `kubectl run` functions slightly differently. In v1.2 it will launch a deployment + replica set. In v1.1 it will launch a replication controller.

In this case unless something changed from the last time I tested 1.2 the upstream docs won't actually work as expected (because they document `kubectl get rc` but that will be empty -- because it launches a replicaSet).

aaronlevy:
Could we also update the other references at the same time: 
Documentation/kubernetes-on-vagrant-single.md
Documentation/kubernetes-on-vagrant.md

Documentation/deploy-addons.md
robszumski: I'm ok with this, but we're almost at the point where we need to have a list of all the places that need to be updated when a new release comes out

robszumski:
Made these changes in https://github.com/coreos/coreos-kubernetes/pull/342 since I wanna get these broken links fixed asap.

PR #326, opened 2016-03-18
Correct location for the example multi-tier web application

j16r:


aaronlevy:
Closing as duplicate of https://github.com/coreos/coreos-kubernetes/pull/323 (they touch different part of docs, but we should fix in same PR).

PR #304, opened 2016-03-10
use vxlan backend in aws

spacepluk:

PR #328, opened 2016-03-21
Add aws-vpc support to flannel & IAM rules

pieterlange:
Add a bunch of aws specific features & minor nits/bugs

Features:
  - Use aws-vpc plugin for flannel (and set up IAM rules) for VPC
    fabric native pod routing
  - Add IAM rules for ECR

Bugfix/workaround:
  - Let hosts set their own source/dest check (required for new flannel
    setup)
  - Set Content-Type header when submitting resources to the kubernetes
    apiserver (fails on >v1.2.0)

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -30,7 +30,7 @@ coreos:
           content: |
             [Service]
             ExecStartPre=/usr/bin/curl --silent -X PUT -d \
-            "value={\"Network\" : \"{{.PodCIDR}}\"}" \
+            "value={\"Network\" : \"{{.PodCIDR}}\", \"Backend\": {\"Type\": \"aws-vpc\"} }" \
             http://localhost:2379/v2/keys/coreos.com/network/config?prevExist=false
     - name: kubelet.service
       command: start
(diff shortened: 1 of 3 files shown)


pieterlange:
> Let hosts set their own source/dest check (required for new flannel setup)

This is a little bit of a nasty workaround due to a limitation in AWS (can't set the option to disable source/destination checks in the launch configuration of the auto scaling group). The main downside to this is we have to assign the "ec2:ModifyInstanceAttribute" permission to instances.

aaronlevy:
There is currently an open PR (https://github.com/coreos/coreos-kubernetes/pull/304) where we will be switching flannel to use the vxlan backend (to reduce overhead). As far as the vpc-backend, in the past there have been issues related to [VPC limits](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Appendix_Limits.html#vpc-limits-route-tables).

Would the vxlan backend satisfy your use-case? Alternatively, I could see this as a possible configurable item -- but I think due to the vpc limits our default for the near future is going to be the vxlan backend.

pieterlange:
Ah, the VPC limits are a good point. I don't expect to grow my clusters beyond 50 nodes (soft limit) nor 100 (hard limit) but i guess if you are this would be a regression. In any case, if anyone wants to run the vpc-backend, this is how you could do it. :-)

Do you want a seperate pull request for the ECR iam rules?

pieterlange:
Closing because there's better options available (vxlan) and the content-type 'bug' is fixed in another pull request.

PR #330, opened 2016-03-21
End init command instructions with a period.

cgag:


colhom:
@cgag is this the entirety of the missing punctuation? Ideally would like to take care of this problem in one fell swoop.

cgag:
Poked through the other commands and found some more and opened a new PR: #347, closing this one.

PR #332, opened 2016-03-22
kube-aws: kubernetesServiceIP is no longer a parameter.

colhom:
It is now cacluated from Cluster.ServiceCIDR, so it is now a member of the Config struct.

\cc @aaronlevy @ericchiang

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -31,7 +31,6 @@ func newDefaultCluster() *Cluster {
 		ControllerIP:             "10.0.0.50",
 		PodCIDR:                  "10.2.0.0/16",
 		ServiceCIDR:              "10.3.0.0/24",
-		KubernetesServiceIP:      "10.3.0.1",
 		DNSServiceIP:             "10.3.0.10",
 		K8sVer:                   "v1.1.8_coreos.0",
 		HyperkubeImageRepo:       "quay.io/coreos/hyperkube",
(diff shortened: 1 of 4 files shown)


aaronlevy:
Rather than restructuring a bunch of stuff, couldn't we just set the KubernetsServiceIP at the entry points to the cluster struct? More or less looks like that would just be `configFromBytes()` (where we could just parse the IP from the serviceCIDR)

colhom:
It's a matter of keeping with convention. Derived values that are computed from the yaml parameters are set in `cluster.Config` method, and are members of the `Config` struct.

The `Config` struct is `Cluster` + derived values

multi-node/aws/pkg/config/config.go
```diff
@@ -96,12 +94,18 @@ type Cluster struct {
 
 func (c Cluster) Config() (*Config, error) {
```
colhom: All configuration values computed from Cluster parameters are derived/set in this method.

aaronlevy:
My gut reaction is that this feels off. If this simple of a change requires this much code, either we structured it in a way that makes maintainability difficult, or there could be some better options.

Some concerns:

We now have validation of a config happening in 2 places: `cluster.valid()` and `cluster.Config()`

The valid() method actually only validates against the ServiceCIDR, not the stored KubernetesServiceIP. So there could be situations where that (public) field is not valid, but the valid() code reports that everything is fine.

What about, as another option:

It looks like the only places the KubernetesServiceIP are referenced are in `valid()` and `NewTLSAssets`. The `valid()` function already just uses the ServiceCIDR to determine the kubernetes serviceIP, why don't we just update NewTLSAssets to do the same (and not store as a field on config{})

colhom:
I agree, it's better now.

aaronlevy:
lgtm

PR #333, opened 2016-03-22
kube-aws: do go generate in build script

colhom:


aaronlevy:
lgtm. May want to eventually move this into an "internal" package as suggested by @ericchiang

PR #334, opened 2016-03-22
kube-aws: stub out kms encryption interface

colhom:
The tls compaction codepath does not behave conditionally on Config.KmsKeyARN being set.

\cc @aaronlevy

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -53,7 +53,7 @@ coreos:
         WantedBy=multi-user.target
 
     - name: decrypt-tls-assets.service
-      enable: {{ne .KMSKeyARN ""}}
+      enable: true
       content: |
         [Unit]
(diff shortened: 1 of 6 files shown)


multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -53,7 +53,7 @@ coreos:
         WantedBy=multi-user.target
 
     - name: decrypt-tls-assets.service
-      enable: {{ne .KMSKeyARN ""}}
+      enable: true
```
aaronlevy: Do we need to keep this flag at all?

aaronlevy:
Not sure we need those enable flags anymore, but other than that lgtm

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: `enable: true` allows the `Install` section to be evaluated. That's how it registers as a dependency of `kubelet.service`

PR #341, opened 2016-03-22
all: curl commands to avoid breakage in 1.2

pbx0:
Based on https://github.com/coreos/coreos-kubernetes/pull/319/files.


pbx0:
@aaronlevy

aaronlevy:
lgtm

PR #344, opened 2016-03-23
kube-aws: improve cloudformation json validation

colhom:
When editing the cloudformation JSON, syntax errors can be very hard to track down based on the error message from the JSON parser alone.

SyntaxError does provide a byte offset, which is now used to show a few lines of context surrounding the location in the JSON document where the error occurred.


aaronlevy:
lgtm

PR #346, opened 2016-03-23
Support for deploying to existing VPC

colhom:
I've added some validation as well so folks don't waste time creating clusters and tracking down subnet conflicts in the cloudformation events logs when the create fails.

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
+	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/config"
 )
@@ -42,38 +43,104 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
 	}
 
 	return &Cluster{
-		clusterName: aws.String(cfg.ClusterName),
-		svc:         cloudformation.New(session.New(awsConfig)),
+		Cluster: *cfg,
+		session: session.New(awsConfig),
 	}
 }
 
 type Cluster struct {
-	clusterName *string
-	svc         *cloudformation.CloudFormation
+	config.Cluster
+	session *session.Session
 }
 
 func (c *Cluster) ValidateStack(stackBody string) (string, error) {
-	input := &cloudformation.ValidateTemplateInput{
+	validateInput := cloudformation.ValidateTemplateInput{
 		TemplateBody: &stackBody,
 	}
 
-	validationReport, err := c.svc.ValidateTemplate(input)
+	cfSvc := cloudformation.New(c.session)
+	validationReport, err := cfSvc.ValidateTemplate(&validateInput)
 	if err != nil {
-		return "", fmt.Errorf("Invalid cloudformation stack: %v", err)
+		return "", fmt.Errorf("invalid cloudformation stack: %v", err)
 	}
 
-	return validationReport.String(), err
+	return validationReport.String(), nil
+}
+
+type ec2Service interface {
+	DescribeVpcs(*ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error)
+	DescribeSubnets(*ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error)
+}
+
+func (c *Cluster) validateExistingVPCState(ec2Svc ec2Service) error {
+	if c.VPCID == "" {
+		//The VPC will be created. No existing state to validate
+		return nil
+	}
+
+	describeVpcsInput := ec2.DescribeVpcsInput{
+		VpcIds: []*string{aws.String(c.VPCID)},
+	}
+	vpcOutput, err := ec2Svc.DescribeVpcs(&describeVpcsInput)
+	if err != nil {
+		return fmt.Errorf("error describing existing vpc: %v", err)
+	}
+	if len(vpcOutput.Vpcs) == 0 {
+		return fmt.Errorf("could not find vpc %s in region %s", c.VPCID, c.Region)
+	}
+	if len(vpcOutput.Vpcs) > 1 {
+		//Theoretically this should never happen. If it does, we probably want to know.
+		return fmt.Errorf("found more than one vpc with id %s. this is NOT NORMAL.", c.VPCID)
+	}
+
+	existingVPC := vpcOutput.Vpcs[0]
+
+	if *existingVPC.CidrBlock != c.VPCCIDR {
+		//If this is the case, our network config validation cannot be trusted and we must abort
+		return fmt.Errorf("configured vpcCidr (%s) does not match actual existing vpc cidr (%s)", c.VPCCIDR, *existingVPC.CidrBlock)
+	}
+
+	describeSubnetsInput := ec2.DescribeSubnetsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("vpc-id"),
+				Values: []*string{existingVPC.VpcId},
+			},
+		},
+	}
+
+	subnetOutput, err := ec2Svc.DescribeSubnets(&describeSubnetsInput)
+	if err != nil {
+		return fmt.Errorf("error describing subnets for vpc: %v", err)
+	}
+
+	subnetCIDRS := make([]string, len(subnetOutput.Subnets))
+	for i, existingSubnet := range subnetOutput.Subnets {
+		subnetCIDRS[i] = *existingSubnet.CidrBlock
+	}
+
+	if err := c.ValidateExistingVPC(*existingVPC.CidrBlock, subnetCIDRS); err != nil {
+		return fmt.Errorf("error validating existing VPC: %v", err)
+	}
+
+	return nil
 }
 
 func (c *Cluster) Create(stackBody string) error {
+	ec2Svc := ec2.New(c.session)
+	if err := c.validateExistingVPCState(ec2Svc); err != nil {
+		return err
+	}
+
+	cfSvc := cloudformation.New(c.session)
 	creq := &cloudformation.CreateStackInput{
-		StackName:    c.clusterName,
+		StackName:    aws.String(c.ClusterName),
 		OnFailure:    aws.String("DO_NOTHING"),
 		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
 		TemplateBody: &stackBody,
 	}
 
-	resp, err := c.svc.CreateStack(creq)
+	resp, err := cfSvc.CreateStack(creq)
 	if err != nil {
 		return err
 	}
@@ -82,7 +149,7 @@ func (c *Cluster) Create(stackBody string) error {
 		StackName: resp.StackId,
 	}
 	for {
-		resp, err := c.svc.DescribeStacks(&req)
+		resp, err := cfSvc.DescribeStacks(&req)
 		if err != nil {
 			return err
 		}
@@ -106,13 +173,14 @@ func (c *Cluster) Create(stackBody string) error {
 }
 
 func (c *Cluster) Update(stackBody string) (string, error) {
+	cfSvc := cloudformation.New(c.session)
 	input := &cloudformation.UpdateStackInput{
 		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
-		StackName:    c.clusterName,
+		StackName:    aws.String(c.ClusterName),
 		TemplateBody: &stackBody,
 	}
 
-	updateOutput, err := c.svc.UpdateStack(input)
+	updateOutput, err := cfSvc.UpdateStack(input)
 	if err != nil {
 		return "", fmt.Errorf("error updating cloudformation stack: %v", err)
 	}
@@ -120,7 +188,7 @@ func (c *Cluster) Update(stackBody string) (string, error) {
 		StackName: updateOutput.StackId,
 	}
 	for {
-		resp, err := c.svc.DescribeStacks(&req)
+		resp, err := cfSvc.DescribeStacks(&req)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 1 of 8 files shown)


pieterlange:
Nice work! Making a separate VPC for each k8s cluster would be a real pain :)

four43:
Hey thanks! This is great!

colhom:
\cc @aaronlevy

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,132 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
 	}
 
-	return validationReport.String(), err
+	describeStacksInput := cloudformation.DescribeStacksInput{
+		StackName: aws.String(c.ClusterName),
+	}
+
+	//Find out if stack exists already. This determines whether we should do subnet conflict validatio
+	var stackExists bool
+	stackNotExistExpr := regexp.MustCompile(fmt.Sprintf("^ValidationError: Stack with id %s does not exist", c.ClusterName))
```
aaronlevy: This is going to be really fragile. There has to be a better way to determine if a stack exist.

What about just `len(describeStackOutput.Stacks) > 0` ?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,132 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
-		return "", fmt.Errorf("Invalid cloudformation stack: %v", err)
+		return "", fmt.Errorf("invalid cloudformation stack: %v", err)
 	}
 
-	return validationReport.String(), err
+	describeStacksInput := cloudformation.DescribeStacksInput{
+		StackName: aws.String(c.ClusterName),
+	}
+
+	//Find out if stack exists already. This determines whether we should do subnet conflict validatio
```
aaronlevy: typo: validatio

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,132 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+	}
+
+	subnetOutput, err := ec2Svc.DescribeSubnets(&describeSubnetsInput)
+	if err != nil {
+		return fmt.Errorf("error describing subnets for vpc: %v", err)
+	}
+
+	//Config validation has already ensured this subnet is contained by the existing VPC CIDR
+	//We need access to the net objects so we can detect conflicts
+	subnetIP, subnetNet, err := net.ParseCIDR(c.InstanceCIDR)
```
aaronlevy: nit: could have more descriptive variable names (s/subnetIP/instanceIP?)

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,132 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+	if err != nil {
+		return fmt.Errorf("error parsing instances cidr %s : %v", c.InstanceCIDR, err)
+	}
+
+	//Loop through all existing subnets in the VPC and look for conflicting CIDRS
+	for _, existingSubnet := range subnetOutput.Subnets {
```
aaronlevy: This seems to just check overlap of instanceCIDR + VPC CIDR. What about service / pod CIDRS?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,132 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+		fmt.Println("Existing VPC detected. Will validate for subnet cidr conflicts")
+		if err := c.validateExistingVPC(); err != nil {
+			return "", err
+		}
+	}
+
+	return validationReport.String(), nil
+}
+
+func (c *Cluster) validateExistingVPC() error {
```
aaronlevy: Would help to add a comment for this function describing it's intended functionality.

multi-node/aws/pkg/config/config.go
```diff
@@ -108,9 +114,25 @@ func (c Cluster) Config() (*Config, error) {
 		return nil, fmt.Errorf("failed getting AMI for config: %v", err)
 	}
 
+	//Set logical name constants
+	config.VPCLogicalName = vpcLogicalName
+
+	//Set reference strings
+	config.VPCRef = dynamicResourceRef(vpcLogicalName, config.VPCID)
```
aaronlevy: It seems like there are only 2 options for dynamicResourceRef. Might suggest just changing this to:

```
if resID != "" {
  config.VPCRef = fmt.Sprintf("%q", resID)
} else {
  config.VPCRef = `{ "Ref" : "VPC"}`
}
```

multi-node/aws/pkg/config/config_test.go
```diff
@@ -77,6 +82,8 @@ controllerIP: 10.4.3.5
 podCIDR: 172.4.0.0/16
 serviceCIDR: 172.5.0.0/16
 dnsServiceIP: 172.6.100.101 #dnsServiceIP not in service CIDR
+`, `
+routeTableId: rtb-xxxxxx # routeTableId specified without vpcId
```
aaronlevy: Given that you're not testing the actual failure type, this test is likely failing for other reasons (has no other config items). Ideally we could also test for the expected failure, but in the short term you could at least add the rest of the values to make this an otherwise passing case (with the exception of the missing vpcID).

multi-node/aws/pkg/cluster/cluster.go
colhom: Config validation already ensures that that the service/pod CIDRs are disjoint from VPC CIDR.

multi-node/aws/pkg/config/config_test.go
colhom: All these strings are appended to [`MinimalYamlConfig`](https://github.com/colhom/coreos-kubernetes/blob/existing-vpc/multi-node/aws/pkg/config/config_test.go#L8) before parsing, which has all the mandatory config options set.

This validity of `MinimalYamlConfig` itself is tested [here](https://github.com/colhom/coreos-kubernetes/blob/existing-vpc/multi-node/aws/pkg/config/config_test.go#L17)

multi-node/aws/pkg/cluster/cluster.go
colhom: `DescribeStacks` returns an Validation error if there are no stacks that have the given name.

The other option is to use `ListStacks` and loop through all the stacks checking for that stack name.

multi-node/aws/pkg/config/config.go
colhom: TIL that `%q` is better than `"%s"` in most cases

colhom:
@aaronlevy review items addressed.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,136 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
 	}
 
-	return validationReport.String(), err
+	describeStacksInput := cloudformation.DescribeStacksInput{
+		StackName: aws.String(c.ClusterName),
+	}
+
+	//Find out if stack exists already. This determines whether we should do subnet conflict validation
+	var stackExists bool
+	stackNotExistExpr := regexp.MustCompile(fmt.Sprintf("^ValidationError: Stack with id %s does not exist", c.ClusterName))
```
aaronlevy: Why do we care at all if the stack exists? Couldn't a VPC exist outside of a cloud-formation stack (and this would skip validation code).

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,136 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+			},
+		},
+	}
+
+	subnetOutput, err := ec2Svc.DescribeSubnets(&describeSubnetsInput)
+	if err != nil {
+		return fmt.Errorf("error describing subnets for vpc: %v", err)
+	}
+
+	//Config validation has already ensured this subnet is contained by the existing VPC CIDR
```
aaronlevy: This concerns me a bit. Relying on another piece of code for verification, which is merely enforced by comment+convention could lead to future issues if people are not careful to update both in unison. My initial thought would be that you either specify a VPCID or a VPC CIDR. Then validation path becomes something like:

if vpcID {
  existingVPCCidr, existingInstanceCIDRS = getCidrsFromAWS(vpcID)
}
networkValidation(existingVPCCIDR, existingInstanceCIDRs, config)

Then the validation code is more easily tested (separate concerns from network calls to aws), and is contained in one place. If no vpcID is given as part of the config, then there is no "existing" to validate against (and config passes through the same validation steps .. just with less to check against).

Doesn't have to be structured like the above, but I'm wary of relying on convention + no tests.

multi-node/aws/pkg/cluster/cluster.go
aaronlevy: I was checking on this and came across a sort of related nit: we are checking the same overlap twice:
https://github.com/colhom/coreos-kubernetes/blob/existing-vpc/multi-node/aws/pkg/config/config.go#L336
https://github.com/colhom/coreos-kubernetes/blob/existing-vpc/multi-node/aws/pkg/config/config.go#L347

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +45,136 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+	}
+
+	//Loop through all existing subnets in the VPC and look for conflicting CIDRS
+	for _, existingSubnet := range subnetOutput.Subnets {
+		existingSubnetIP, existingSubnetNet, err := net.ParseCIDR(*existingSubnet.CidrBlock)
+		if err != nil {
+			return fmt.Errorf("error parsing existing subnet cidr %s : %v", *existingSubnet.CidrBlock, err)
+		}
+
+		if existingSubnetNet.Contains(instanceIP) || instanceNet.Contains(existingSubnetIP) {
```
aaronlevy: nit: I believe we do this check elsewhere. Might make sense to have a helper function:

```
cidrOverlap(a, b *net.IPNet) bool {
  return a.Contains(b.IP) || b.Contains(a.IP)
}
```

Another nit: `subnetNet` is a strange variable name. existingInstanceNet?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -66,46 +64,19 @@ func (c *Cluster) ValidateStack(stackBody string) (string, error) {
...
-Ensures that the new instance subnet being created does not overlap with any existing subnets in the VPC.
-*/
-func (c *Cluster) validateExistingVPC() error {
-	ec2Svc := ec2.New(c.session)
+type ec2Service interface {
+	DescribeVpcs(*ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error)
+	DescribeSubnets(*ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error)
+}
+
+func (c *Cluster) validateExistingVPCState(ec2Svc ec2Service) error {
```
colhom: The unit test which utilizes this dummy interface is in the next commit

multi-node/aws/pkg/config/config.go
```diff
@@ -61,7 +61,7 @@ func ClusterFromFile(filename string) (*Cluster, error) {
 }
 
 //Necessary for unit tests, which store configs as hardcoded strings
-func clusterFromBytes(data []byte) (*Cluster, error) {
+func ClusterFromBytes(data []byte) (*Cluster, error) {
```
colhom: Need to expose this for pkg/cluster unit test in next commit

multi-node/aws/pkg/config/config_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 )
 
-const MinimalConfigYaml = `externalDNSName: test-external-dns-name
+const minimalConfigYaml = `externalDNSName: test-external-dns-name
```
colhom: it's not useful to expose this, so took the liberty of cleaning it up

colhom:
@aaronlevy ready for review

multi-node/aws/pkg/config/config.go
```diff
@@ -300,25 +328,25 @@ func (cfg Cluster) valid() error {
 		)
 	}
 
-	podNetIP, podNet, err := net.ParseCIDR(cfg.PodCIDR)
+	_, podNet, err := net.ParseCIDR(cfg.PodCIDR)
 	if err != nil {
 		return fmt.Errorf("invalid podCIDR: %v", err)
 	}
-	if vpcNet.Contains(podNetIP) {
+	if vpcNet.Contains(podNet.IP) {
```
aaronlevy: This seems redundant with https://github.com/coreos/coreos-kubernetes/pull/346/files#diff-e5ad8b828a5de3fe2229bee5c64ab2a0R346

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +43,104 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+	if err != nil {
+		return fmt.Errorf("error describing subnets for vpc: %v", err)
+	}
+
+	subnetCIDRS := make([]string, len(subnetOutput.Subnets))
+	for i, existingSubnet := range subnetOutput.Subnets {
+		subnetCIDRS[i] = *existingSubnet.CidrBlock
+	}
+
+	if err := c.ValidateExistingVPC(*existingVPC.CidrBlock, subnetCIDRS); err != nil {
```
aaronlevy: Up to you if you want to change this:

I had suggested breaking this out to make testing possibly easier, but the way you have it mocked out, it's not really necessary (and this function ends up only being called here).

You could inline this function again in that case, if it seemed simpler than creating temp vars to pass into this. But again, totally up to you (just noticed my original suggestion doesn't really help in this case).

aaronlevy:
Couple minor comments, but otherwise lgtm

multi-node/aws/pkg/config/config.go
colhom: Yep, will nix this block

multi-node/aws/pkg/cluster/cluster.go
colhom: All the cidrOverlap logic is consolidated in the config package, such that cluster pacakge can deal only with aws api interaction related stuff. I kind of like it.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -42,38 +43,104 @@ func New(cfg *config.Cluster, awsDebug bool) *Cluster {
...
+	}
+	vpcOutput, err := ec2Svc.DescribeVpcs(&describeVpcsInput)
+	if err != nil {
+		return fmt.Errorf("error describing existing vpc: %v", err)
+	}
+	if len(vpcOutput.Vpcs) == 0 {
+		return fmt.Errorf("could not find vpc %s in region %s", c.VPCID, c.Region)
+	}
+	if len(vpcOutput.Vpcs) > 1 {
+		//Theoretically this should never happen. If it does, we probably want to know.
```
cgag: I feel like this an invariant we can trust to amazon.

rothgar:
How is this supposed to be used/configured in cluster.yaml? I'm assuming I'll need to build from git but don't see documentation in the repo.

pieterlange:
@rothgar

PR #348, opened 2016-03-23
docs: fix code output

robszumski:
Lack of newline breaks rendering on coreos.com:

image


robszumski:
Removed in https://github.com/coreos/coreos-kubernetes/pull/350

PR #349, opened 2016-03-23
docs: link to kube-aws releases

robszumski:
Seems like this is a better location for this link. This gets you release notes, binaries, and an easy way to browse the project if you're interested.


colhom:
Agreed, they can easily get from there to the repo homepage if they so desire. LGTM

PR #350, opened 2016-03-23
docs: remove duplicate credential step

robszumski:
Right now we have two stanzas that tell you how to configure credentials. I remove the first and kept the second since it has the two options instead of just env vars.


colhom:
@robszumski the first way (with ~/.aws/credentials) does not actually work with kube-aws currently.

robszumski:
@colhom It did for me on 0.5.1

colhom:
Discussed in meatspace, it seems that this does in fact work if the credentials are set up properly. We need to figure out why the shared credentials don't seem to work for everyone.

LGTM

PR #351, opened 2016-03-23
docs: tweak AWS language

robszumski:


colhom:
@robszumski do you want to change all the `sh` blocks to `shell blocks`?

robszumski:
@colhom I don't see any more in that doc. I did it only because I _know_ that `sh` works and wasn't sure about `shell`. I don't think its a big deal either way.

colhom:
lgtm

PR #355, opened 2016-03-25
upgrade from v1.1.8 to v1.2.0

pbx0:
I changed some doc links, moved all references of the `v1.1.8_coreos.0` image to `v1.2.0_coreos.0`. Most importantly the kube-dns addon was updated to v11 which allows the v1.2.0 conformance tests to pass.

cc @aaronlevy

### DIFF
--- a/Documentation/deploy-addons.md
+++ b/Documentation/deploy-addons.md
@@ -140,7 +140,7 @@ spec:
             cpu: 10m
             memory: 20Mi
         args:
-        - -cmd=nslookup kubernetes.default.svc.cluster.local localhost >/dev/null
+        - -cmd=nslookup kubernetes.default.svc.cluster.local 127.0.0.1 >/dev/null
         - -port=8080
         ports:
         - containerPort: 8080
--- a/Documentation/deploy-addons.md
+++ b/Documentation/deploy-addons.md
@@ -172,5 +172,5 @@ $ kubectl get pods --namespace=kube-system | grep kube-dns-v11
 <div class="co-m-docs-next-step">
   <p>Now that you have a working Kubernetes cluster with a functional CLI tool, you are free to deploy Kubernetes-ready applications.</p>
   <p>Start with a multi-tier web application (Guestbook) from the official Kubernetes documentation to visualize how the various Kubernetes components fit together.</p>
-  <a href="https://github.com/kubernetes/kubernetes/blob/release-1.1/examples/guestbook-go/README.md" class="btn btn-default btn-icon-right" data-category="Docs Next" data-event="kubernetes.io: Guestbook">Deploy the Guestbook Sample app</a>
+  <a href="https://github.com/kubernetes/kubernetes/blob/release-1.2/examples/guestbook/README.md" class="btn btn-default btn-icon-right" data-category="Docs Next" data-event="kubernetes.io: Guestbook">Deploy the Guestbook Sample app</a>
 </div>
--- a/Documentation/deploy-addons.md
+++ b/Documentation/deploy-addons.md
@@ -6,7 +6,7 @@ Add-ons are built on the same Kubernetes components as user-submitted jobs &mdas
 
 First create `dns-addon.yml` on your local machine and replace the variable. There is a lot going on in there, so let's break it down after you create it.
 
-[k8s-dns]: http://kubernetes.io/v1.1/docs/admin/dns.html
+[k8s-dns]: http://kubernetes.io/docs/admin/dns.html
 
 * Replace `${DNS_SERVICE_IP}`
 
@@ -39,29 +39,32 @@ spec:
 apiVersion: v1
 kind: ReplicationController
 metadata:
-  name: kube-dns-v9
+  name: kube-dns-v11
   namespace: kube-system
   labels:
     k8s-app: kube-dns
-    version: v9
+    version: v11
     kubernetes.io/cluster-service: "true"
 spec:
   replicas: 1
   selector:
     k8s-app: kube-dns
-    version: v9
+    version: v11
   template:
     metadata:
       labels:
         k8s-app: kube-dns
-        version: v9
+        version: v11
         kubernetes.io/cluster-service: "true"
     spec:
       containers:
       - name: etcd
-        image: gcr.io/google_containers/etcd:2.0.9
+        image: gcr.io/google_containers/etcd-amd64:2.2.1
         resources:
           limits:
+            cpu: 100m
+            memory: 500Mi
+          requests:
             cpu: 100m
             memory: 50Mi
         command:
@@ -78,24 +81,47 @@ spec:
         - name: etcd-storage
           mountPath: /var/etcd/data
       - name: kube2sky
-        image: gcr.io/google_containers/kube2sky:1.11
+        image: gcr.io/google_containers/kube2sky:1.14
         resources:
           limits:
             cpu: 100m
+            memory: 200Mi
+          requests:
+            cpu: 100m
             memory: 50Mi
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 60
+          timeoutSeconds: 5
+          successThreshold: 1
+          failureThreshold: 5
+        readinessProbe:
+          httpGet:
+            path: /readiness
+            port: 8081
+            scheme: HTTP
+          initialDelaySeconds: 30
+          timeoutSeconds: 5
         args:
         # command = "/kube2sky"
-        - -domain=cluster.local
+        - --domain=cluster.local
       - name: skydns
-        image: gcr.io/google_containers/skydns:2015-03-11-001
+        image: gcr.io/google_containers/skydns:2015-10-13-8c72f8c
         resources:
           limits:
+            cpu: 100m
+            memory: 200Mi
+          requests:
             cpu: 100m
             memory: 50Mi
         args:
         # command = "/skydns"
-        - -machines=http://localhost:4001
+        - -machines=http://127.0.0.1:4001
         - -addr=0.0.0.0:53
+        - -ns-rotate=false
         - -domain=cluster.local.
         ports:
         - containerPort: 53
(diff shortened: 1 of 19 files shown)


Documentation/kubernetes-on-vagrant.md
```diff
@@ -95,6 +95,7 @@ NAME          LABELS                               STATUS
 <div class="co-m-docs-next-step">
   <p><strong>Is kubectl working correctly?</strong></p>
   <p>Now that you've got a working Kubernetes cluster with a functional CLI tool, you are free to deploy Kubernetes-ready applications.
-Start with a <a href="https://github.com/kubernetes/kubernetes/blob/release-1.1/examples/guestbook-go/README.md" data-category="Docs Next" data-event="kubernetes.io: Guestbook">multi-tier web application</a> from the official Kubernetes documentation to visualize how the various Kubernetes components fit together.</p>
-  <a href="https://github.com/kubernetes/kubernetes/blob/release-1.1/examples/guestbook-go/README.md" class="btn btn-default btn-icon-right" data-category="Docs Next" data-event="kubernetes.io: Guestbook">View the Guestbook example app</a>
+<<<<<<< 4d6aed27743828349ce661f6055baa9f6144abc6
```
aaronlevy: leftover conflict hash

aaronlevy:
Awesome!
Few last things we should tick off for a 1.2 release:

* replace `--register-node=false` with `--register-schedulable=false` on master nodes
* Remove podmaster manifest from master nodes 
* Add `--leader-elect=true` flag to scheduler + controller-manager
* move controller manager + scheduler manifests from `/srv/kubernetes/manifests` to `/etc/kubernetes/manifests` (they can run on all master nodes now with leader-elect)
* bump version for heapster + influxdb addon manifests

pieterlange:
As per #367 we should also add `Environment=RKT_OPTS="--volume=resolv,kind=host,source=/etc/resolv.conf --mount volume=resolv,target=/etc/resolv.conf"` to the `kubelet.service` files so kubelet can properly resolve itself if the host is using an internal DNS hostname.

Otherwise the kubelet on the master won't properly register and goes into a loop deleting master pod definitions.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -359,12 +312,16 @@ write_files:
                       "-initial-cluster-token",
                       "skydns-etcd"
                     ],
-                    "image": "gcr.io/google_containers/etcd:2.0.9",
+                    "image": "gcr.io/google_containers/etcd:2.2.1",
```
aaronlevy: Upstream seems to be using `etcd-amd64`: https://github.com/kubernetes/kubernetes/blob/release-1.2/cluster/addons/dns/skydns-rc.yaml.in#L24

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -391,18 +374,45 @@ write_files:
                     "args": [
                       "-machines=http://localhost:4001",
                       "-addr=0.0.0.0:53",
+                      "-ns-rotate=false", 
                       "-domain=cluster.local."
                     ],
-                    "image": "gcr.io/google_containers/skydns:2015-03-11-001",
+                    "image": "gcr.io/google_containers/skydns:2015-10-13-8c72f8c",
```
aaronlevy: Hard to tell here, but this container is being defined twice (and missing name field so apiserver complains). We should remove one

Documentation/deploy-addons.md
```diff
@@ -104,26 +130,15 @@ spec:
...
         image: gcr.io/google_containers/exechealthz:1.0
         resources:
           limits:
             cpu: 10m
             memory: 20Mi
+          requests:
+            cpu: 10m
+            memory: 20Mi
         args:
         - -cmd=nslookup kubernetes.default.svc.cluster.local localhost >/dev/null
```
aaronlevy: Looks like upstream has swapped "localhost" with 127.0.0.1: https://github.com/kubernetes/kubernetes/blob/release-1.2/cluster/addons/dns/skydns-rc.yaml.in#L121

minor, but could be an issue if system had "localhost" misconfigured for some reason.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -416,25 +426,22 @@ write_files:
...
                       }
-                    }
+                    },
+                    "terminationMessagePath": "/dev/termination-log",
+                    "imagePullPolicy": "IfNotPresent"
                   },
                   {
                     "args": [
-                      "-cmd=nslookup kubernetes.default.svc.cluster.local localhost",
+                      "-cmd=nslookup kubernetes.default.svc.cluster.local localhost >/dev/null",
```
aaronlevy: s/localhost/127.0.0.1

single-node/user-data
```diff
@@ -275,150 +275,183 @@ EOF
...
+                "cpu": "100m",
+                "memory": "50Mi"
+              }
+            },
+            "terminationMessagePath": "/dev/termination-log",
+            "imagePullPolicy": "IfNotPresent"
+          },
+          {
+            "args": [
+              "-cmd=nslookup kubernetes.default.svc.cluster.local localhost >/dev/null",
```
aaronlevy: s/localhost/127.0.0.1

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -391,18 +374,45 @@ write_files:
                     "args": [
                       "-machines=http://localhost:4001",
```
aaronlevy: s/localhost/127.0.0.1

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: This got all messed up in the last commit, i'm trying to untangle things now.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -446,20 +453,29 @@ write_files:
...
+                "terminationGracePeriodSeconds": 30,
                 "dnsPolicy": "Default",
                 "volumes": [
                   {
                     "emptyDir": {},
                     "name": "etcd-storage"
                   }
-                ]
+                ],
+                "securityContext": {}
```
aaronlevy: This doesn't seem to be in the upstream copy

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -446,20 +453,29 @@ write_files:
...
                         "cpu": "10m",
                         "memory": "20Mi"
                       }
-                    }
+                    },
+                    "terminationMessagePath": "/dev/termination-log",
+                    "imagePullPolicy": "IfNotPresent"
                   }
                 ],
+                "restartPolicy": "Always",
```
aaronlevy: Not in upstream copy

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -446,20 +453,29 @@ write_files:
...
+                        "memory": "20Mi"
+                      }, 
                       "limits": {
                         "cpu": "10m",
                         "memory": "20Mi"
                       }
-                    }
+                    },
+                    "terminationMessagePath": "/dev/termination-log",
+                    "imagePullPolicy": "IfNotPresent"
```
aaronlevy: terminationMessagePath and ImagePullPolicy are not in upstream copy

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -416,25 +426,22 @@ write_files:
...
+                      }, 
                       "limits": {
                         "cpu": "100m",
-                        "memory": "50Mi"
+                        "memory": "200Mi"
                       }
-                    }
+                    },
+                    "terminationMessagePath": "/dev/termination-log",
+                    "imagePullPolicy": "IfNotPresent"
```
aaronlevy: terminationMessagePath and ImagePullPolicy are not in upstream copy

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: @joeatwork Do you know what this is? Also stuff like 
```
"terminationMessagePath": "/dev/termination-log",
334                     "imagePullPolicy": "IfNotPresent"
```

aaronlevy:
There looks to be quite a few discrepancies between our addons and the upstream copies. In the interest of making this easier, can we maybe solidify a process for converting to json in a somewhat consistent way (to make reviewing this easier:

Maybe:

```
python -c "import json,yaml; print json.dumps(yaml.load(open('upstream-dns-rc.yaml')), indent=2, sort_keys=True)"
```

You will still need to manually remove the "{{ pillar['foo'] }}" in the upstream copies (complains about not being valid yaml) -- but hopefully gives us an easier starting point.

Documentation/deploy-addons.md
pbx0: Got it, changing all the localhost stuff.

aaronlevy:
Another minor change to docs we need to make is s/NUM_MINIONS/NUM_NODES/ in Documentation/conformance-tests.md

ref: https://github.com/coreos/coreos-kubernetes/pull/369#issuecomment-204019045

multi-node/aws/pkg/config/templates/cloud-config-controller
joeatwork: I tested this without the architecture and it worked - I wanted to keep the diff as small as possible. However, I'm pretty sure the arch won't hurt.

multi-node/aws/pkg/config/templates/cloud-config-controller
joeatwork: I generated the JSON by round tripping the YAML through an actual running cluster, and it gave me back a bunch of default fields (including imagePullPolicy/securityDomain/etc)

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: Let's add the arch back, given that's what they are using upstream. I believe they are eventually deprecating their non-arch builds (see: https://github.com/kubernetes/kubernetes/blob/master/cluster/images/etcd/Makefile#L56)

aaronlevy:
Changes LGTM. Assuming final tests pass, let's get this merged.

xied75:
Given we haven't released v.0.5.2, and k8s v1.2.1 has lots of bug fixes, any chance we could land on that other than v1.2.0?

aaronlevy:
I'd like to get this landed as soon as possible, so likely going to get this out with 1.2.0 - but the 1.2.1 release should follow up shortly (fewer moving parts for a patch update).

PR #362, opened 2016-03-29
kube-aws: validate that the key pair exists

### LINKED ISSUE #331: Validate should check if the key pair exists.
Without a valid key pair `kube-aws up` will fail after several minutes. 

```Error creating cluster: Stack creation failed: CREATE_FAILED :  The following resource(s) failed to create: [LaunchConfigurationWorker, InstanceController].```

We should be able to check for that quickly check for the key pair before attempting to create the cluster.

### PR DESCRIPTION
**Pull request by cgag**
fixes #331

### DIFF
--- a/multi-node/aws/cmd/kube-aws/command_validate.go
+++ b/multi-node/aws/cmd/kube-aws/command_validate.go
@@ -34,14 +34,13 @@ func runCmdValidate(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Unable to load cluster config: %v", err)
 	}
 
-	//Validate cloudconfig userdata
+	fmt.Printf("Validating UserData...\n")
 	if err := cfg.ValidateUserData(stackTemplateOptions); err != nil {
 		return err
 	}
+	fmt.Printf("UserData is valid.\n\n")
 
-	fmt.Printf("UserData is valid\n")
-
-	//Validate stack template
+	fmt.Printf("Validating stack template...\n")
 	data, err := cfg.RenderStackTemplate(stackTemplateOptions)
 	if err != nil {
 		return fmt.Errorf("Failed to render stack template: %v", err)
@@ -54,8 +53,10 @@ func runCmdValidate(cmd *cobra.Command, args []string) error {
 	}
 
 	if err != nil {
-		return fmt.Errorf("Error creating cluster: %v", err)
+		return err
 	}
+	fmt.Printf("stack template is valid.\n\n")
+
 	fmt.Printf("Validation OK!\n")
 	return nil
 }
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by colhom** [comment]
#346 exposes a common aws session via the cluster.Cluster struct. Will wait on that.

**Comment by cgag** [comment]
Nice, thought that session stuff smelled a little funny.

**Comment by cgag** [comment]
Had to move some stuff around to avoid a cyclic dependency.  Now that session is in config, the cluster.Cluster struct seems pretty pointless but I left it around to minimize the scope of this PR.

**Comment by colhom** [comment]
@cgag drop the last commit and we should be good to go

File: multi-node/aws/pkg/config/config.go [hunk 1]
```diff
@@ -193,6 +195,27 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
 	return &stackConfig, nil
 }
 
+func (c Cluster) ValidateKey() error {
```
colhom: Move this to cluster package.

multi-node/aws/cmd/kube-aws/command_validate.go
```diff
@@ -34,6 +34,13 @@ func runCmdValidate(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Unable to load cluster config: %v", err)
 	}
 
+	fmt.Printf("Validating Key Pair...\n")
```
colhom: Move this logic to https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/cluster/cluster.go#L56

multi-node/aws/cmd/kube-aws/command_validate.go
```diff
@@ -54,8 +53,10 @@ func runCmdValidate(cmd *cobra.Command, args []string) error {
 	}
 
 	if err != nil {
-		return fmt.Errorf("Error creating cluster: %v", err)
```
cgag: This was just leading to output like "Error creating cluster: Error: Key does not exist".  I think the "Validating stack template..." output added in the validate command makes the extra messaging unnecessary.

multi-node/aws/pkg/config/config.go
cgag: done

multi-node/aws/cmd/kube-aws/command_validate.go
cgag: done

cgag:
Moved the validateKey call to cluster.Create after an offline discussion with @colhom.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -254,3 +263,22 @@ func (c *Cluster) Destroy() error {
 	_, err := cfSvc.DeleteStack(dreq)
 	return err
 }
+
+func (c *Cluster) validateKey() error {
```
colhom: Check out how `validateExistingVPCState` is [set up](`validateExistingVPCState` takes an ec2

cgag: tested

multi-node/aws/pkg/cluster/cluster_test.go
```diff
@@ -1,24 +1,40 @@
...
+type dummyEC2Service struct {
+	VPCs     map[string]VPC
+	KeyPairs map[string]bool
+}
+
 func (svc dummyEC2Service) DescribeVpcs(input *ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error) {
 	output := ec2.DescribeVpcsOutput{}
-	for _, vpcId := range input.VpcIds {
-		if vpc, ok := svc[*vpcId]; ok {
+	for _, vpcID := range input.VpcIds {
```
cgag: This capitalization makes golint happy, but it's inconsistent with the amazon's naming.  I'd prefer to not propagate their deviancy, but I can change this back if we'd rather be consistent.

colhom: golint is our source of truth in these matters.

colhom:
LGTM

PR #369, opened 2016-03-30
Fix typo in sed command in conformance doc

tomdee:


cgag:
As much as I want to consider the current version a mistake, it's actually just written for the mac / bsd version of sed and not the gnu/linux one.  Maybe we should just list both?

tomdee:
Aha. It doesn't work on linux so we should do something. I'll have a think and resubmit the PR.

aaronlevy:
FWIW, internally we always run with the default # of nodes (4), because even changing the worker count, some tests would fail. Not sure how necessary we need the sed line. Could just document that the NUM_MINIONS needs to change if you have less than 4.

Side note. We need to change this to NUM_NODES when version bumped to 1.2

aaronlevy:
An alternative would be to actually have a backup file. I believe that works on both osx/linux.

e.g. `sed -i.bak "s/foo/bar" baz`

aaronlevy:
The conformance tests no longer need the sed replacement. Closing.

PR #375, opened 2016-03-31
Update heapster as a Kubernetes 1.2 addon

joeatwork:
This PRQ is based on the (not yet merged #355 ) - @pbx0 , if it looks good, consider just merging it into your branch?

### DIFF
--- a/multi-node/aws/README.md
+++ b/multi-node/aws/README.md
@@ -30,7 +30,7 @@ export AWS_SECRET_ACCESS_KEY=MY-SECRET-KEY
 Creating a KMS key can be done via the AWS web console or via the AWS cli tool.
 
 ```shell
-$ aws kms --region=<your-region> create-key --description="kube-aws assets"
+$ aws kms --region=us-west-1 create-key --description="kube-aws assets"
 {
     "KeyMetadata": {
         "CreationDate": 1458235139.724,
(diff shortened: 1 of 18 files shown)


pbx0:
@joeatwork This LGTM. Do you think it makes sense if I just cherry-pick your commit on-top of my PR branch for #355? 

One-nit, I would rename the commit msg from `single-node: ...` to something like `all: ..` or whatever your preference of that kinda commit is.

multi-node/aws/README.md
```diff
@@ -30,7 +30,7 @@ export AWS_SECRET_ACCESS_KEY=MY-SECRET-KEY
 Creating a KMS key can be done via the AWS web console or via the AWS cli tool.
 
 ```shell
-$ aws kms --region=<your-region> create-key --description="kube-aws assets"
+$ aws kms --region=us-west-1 create-key --description="kube-aws assets"
```
aaronlevy: Why the change?

joeatwork:
@pbx0 absolutely - I think cherry picking it would be fine. And you're welcome to rename the commit - that commit name is an oops, since this used to be three separate commits (single-node, multi-node/generic, multi-node/aws)

Thanks for your help! (Close this PRQ when you're ready :) )

multi-node/aws/README.md
joeatwork: We explicitly use us-west-1 below in the `kube-aws init` example, and I felt like it was better to shift both to something concrete and working than make the init example use another `<variable>` - (in this case being concrete has a particular value, because the mapping between "US West (N. California)" and "us-west-1" isn't all that obvious to noobs like me)

pbx0:
Will do, thanks! I'll be running some conformance tests with this once I get my pod master commit passing.

multi-node/aws/README.md
aaronlevy: kk. If we use a specific example elsewhere, this seems fine to me.

aaronlevy:
lgtm. Yeah let's just cherry-pick the commit into your branch @pbx0 (should maintain authorship too)

pbx0:
Incorporated into #355

PR #383, opened 2016-04-06
Add info for installing kubectl via homebrew

dgem:
There is a brew for installing kubectl via homebrew (https://github.com/Homebrew/homebrew-core/blob/master/Formula/kubernetes-cli.rb), it would seem sensible to reference it in this documentation (and I guess elsewhere)


colhom:
@dgem I don't know that recommending a method for installing kubectl is in scope for this repo.

\cc @aaronlevy

aaronlevy:
Yeah, I'm going to lean toward just documenting curl. Mainly because it means we would need to continually vet that the brew install target was doing what we expect + it needs to target a specific kubectl version + if it doesn't support a version we want to ship our docs become partially incorrect.

This could just as easily spiral into apt-get/dnf/etc. curl (at least to me) seems like a reasonable common baseline.

Issue #381, opened 2016-04-06
ECR on Kubernetes 1.2.0


elohr:
IAMRoleWorker will need this policy to be able to take advantage of the functionality of pulling images from ECR on AWS:

```
            {
                "Action": [
                    "ecr:GetAuthorizationToken", 
                    "ecr:BatchGetImage"
                ], 
                "Resource": "*", 
                "Effect": "Allow"
            }
```
[https://github.com/kubernetes/kubernetes/issues/23777](url)

spacepluk:
I was just looking into this :)

I had to add also `ecr:GetDownloadUrlForLayer` to make it work.

aaronlevy:
closed in: https://github.com/coreos/coreos-kubernetes/pull/384

PR #384, opened 2016-04-06
kube-aws: add ECR policies to IAMRoleWorker

spacepluk:


spacepluk:
Fixes #381

colhom:
@spacepluk my concern here is that deploying a kube-aws cluster does not _imply_ a need for the instances to have IAM access to ECR images on that AWS account. Maybe this should be configurable, and default to no IAM access?

valichek:
@spacepluk the list of policies is not full, here is suggested policies http://kubernetes.io/docs/user-guide/images/#using-aws-ec2-container-registry, @colhom kubernetes supports ECR by default. So why not to configure ECR from the box?

spacepluk:
Sorry, I completely missed these comments.

@colhom that's a good point but I'd say it makes sense to support it by default.

@valichek thanks, I'll update the patch.

yuvipanda:
It would be awesome to have this merged :D

colhom:
Here is a link to the [docs](http://kubernetes.io/docs/user-guide/images/#using-aws-ec2-container-registry#using-aws-ec2-container-registry) for the ECR-K8s integration @valichek was referring to.

Given that this integration exists, I've been convinced. Thanks everyone!

PR #385, opened 2016-04-06
Documentation: Rephrase leader election/ha section for masters

chancez:
CC @joshix

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -8,7 +8,9 @@ Manual configuration of the required master node services is explained below, bu
 
 The instructions below configure the required master node components using manifests stored in `/etc/kubernetes/manifests`. The kubelet will watch this location for new or modified manifests and run them automatically.
 
-If you are deploying multiple master nodes in a high-availability cluster, these instructions can be repeated for each one you wish to launch. Each of the master components is safe to run on multiple master nodes due to stateless design or built-in leader-election. In the case of leader-election, the service ensures that important cluster management is handled by the leader, with the additional master nodes ready to take over at any time.
+High-availability is achieved by repeating these instructions for each master node. Each of the master components is safe to run on multiple nodes.
+
+The apiserver is stateless, but handles recording the results of leader elections to etcd on behalf of other master components. The controller-manager and scheduler use the leader election mechanism to ensure only one of each is active, leaving the inactive master components ready to assume responsibility in case of failure.
 
 ## Configure Service Components
 


Documentation/deploy-master.md
```diff
@@ -8,7 +8,9 @@ Manual configuration of the required master node services is explained below, bu
 
 The instructions below configure the required master node components using manifests stored in `/etc/kubernetes/manifests`. The kubelet will watch this location for new or modified manifests and run them automatically.
 
-If you are deploying multiple master nodes in a high-availability cluster, these instructions can be repeated for each one you wish to launch. Each of the master components is safe to run on multiple master nodes due to stateless design or built-in leader-election. In the case of leader-election, the service ensures that important cluster management is handled by the leader, with the additional master nodes ready to take over at any time.
+High-availability is achieved by repeating these instructions for each master node. Each of the master components is safe to run on multiple nodes. 
+
+The apiserver is stateless, and handles leader election for the other master components. The controller-manager and scheduler use leader election to ensure only one of each is active at a time. This means the additional master components are ready to take over in the case of failure.
```
joshix: s/apiserver/API server/

joshix: s/This means the/The/
s/in the case of/in case of/

chancez: Good catch. How about this though: `The inactive master components are ready to take over in case of failure.`

chancez: I just realized, the wording `the additional master components` is a bit ambiguous after the previous sentence talking about specific components, and using the word `active`.

aaronlevy: The apiserver doesn't exactly handle leader election. It's implemented as a client lib used in the component code itself (e.g. controller-manager). apiserver handles it in terms of interacting with etcd, but it doesn't have any special code introduced (that I'm aware of).

joshix: @chancez @aaronlevy So how about:

The apiserver is stateless, but handles recording the results of leader elections to etcd on behalf of other master components. The controller-manager and scheduler use the leader election mechanism to ensure only one of each is active, leaving the inactive master components ready to assume responsibility in case of failure.

chancez: That looks solid to me. I'll update the PR.

chancez:
@joshix Updated

joshix:
LGTM

chancez:
It's complaining about `go vet` which I've seen travis have issues with previously. I think travis has been getting rate-limited by uses of `go get` or something like that lately.

PR #386, opened 2016-04-06
kube-aws: enable verbose credential chain errors

### LINKED ISSUE #368: kube-aws: verbose credential chain errors should be enabled
Everyone is confused by the current error messages as seen here: https://github.com/coreos/coreos-kubernetes/issues/94

I think that issue is mostly fixed by the docs update that makes credential configure explicit, but we should do this anyway.  

The fix involves calling `awsConfig.WithCredentialsChainVerboseErrors(true)` on every awsConfig, so we should wait on https://github.com/coreos/coreos-kubernetes/pull/346 since there's some refactoring in there that centralizes the awsConfig creation.

cgag:
There have been a lot of issues with people confused by the opaque
errors the AWS SDK outputs without this option.

Also, not using this option is marked deprecated anyway.

fixes #368

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -36,8 +36,10 @@ func (c *ClusterInfo) String() string {
 }
 
 func New(cfg *config.Cluster, awsDebug bool) *Cluster {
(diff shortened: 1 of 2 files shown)


colhom:
LGTM

PR #388, opened 2016-04-08
kube-aws: fix travis build

colhom:
remove `go get golang.org/x/tools/cmd/vet`


pbx0:
LGTM. Vet is just included in the go distribution starting in 1.6 right?

colhom:
1.5 as well, maybe even earlier than that.

PR #389, opened 2016-04-08
kube-aws: add option to create a record for externalDNSName automatic…

cgag:
…ally

Currently it's on the user to create a record, via Route53 or otherwise,
in order to make the controller IP accessible via externalDNSName.  This
commit adds an option to automatically create a Route53 record in a given
hosted zone.

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/route53"
 
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/config"
 )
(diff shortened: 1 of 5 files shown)


multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -254,3 +259,53 @@ func (c *Cluster) Destroy() error {
...
+
+	if c.RecordSetTTL < 1 {
+		return fmt.Errorf("TTL must be at least 1 second")
+	}
+
+	if c.HostedZone == "" {
+		return fmt.Errorf("hostName cannot be blank when createRecordSet is true")
+	}
+
+	r53 := route53.New(c.session)
```
colhom: `validateDNSSettings` should take the route53 service as a parameter. As is done with `validateExistingVPC`, an interface should be stubbed out and used to exercise this codepath in the  cluster unit tests.

colhom:
@cgag 

* can you fix the commit message? It overflows the github max length.
* rebase on top of master, and drop the validate key pair commit

If at all possible in the future, try to separate the commits out on a per/PR basis. If that can't be avoided for whatever reason, make a note of it in the PR that it depends on another PR being merged and possibly rebased under it, if changes are made.

colhom:
Otherwise, LGTM pending above comments.

Great work cgag.

cgag:
My bad. Dropped the key pair commit and fixed the commit message.

PR #392, opened 2016-04-09
kube-aws: multiple etcd endpoints

orkhanM:
adding ability to specify multiple etcd endpoints.

Currently the kube-aws tool explicitly defines the etcd endpoint as localhost inline in the userdata. This pull request allows for multiple etcd endpoints to be specified in cluster.yaml as a comma delimited list.

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -2,7 +2,6 @@ package config
 
 //go:generate go run templates_gen.go
 //go:generate gofmt -w templates.go
-
 import (
 	"bytes"
 	"encoding/json"
@@ -10,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/url"
 	"strings"
 	"text/template"
 	"unicode/utf8"
@@ -34,6 +34,7 @@ func newDefaultCluster() *Cluster {
 		VPCCIDR:                  "10.0.0.0/16",
 		InstanceCIDR:             "10.0.0.0/24",
 		ControllerIP:             "10.0.0.50",
+		EtcdEndpoints:            "http://10.0.0.50:2379",
 		PodCIDR:                  "10.2.0.0/16",
 		ServiceCIDR:              "10.3.0.0/24",
 		DNSServiceIP:             "10.3.0.10",
@@ -81,23 +82,26 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 }
 
 type Cluster struct {
-	ClusterName              string            `yaml:"clusterName"`
-	ExternalDNSName          string            `yaml:"externalDNSName"`
-	KeyName                  string            `yaml:"keyName"`
-	Region                   string            `yaml:"region"`
-	AvailabilityZone         string            `yaml:"availabilityZone"`
-	ReleaseChannel           string            `yaml:"releaseChannel"`
-	ControllerInstanceType   string            `yaml:"controllerInstanceType"`
-	ControllerRootVolumeSize int               `yaml:"controllerRootVolumeSize"`
-	WorkerCount              int               `yaml:"workerCount"`
-	WorkerInstanceType       string            `yaml:"workerInstanceType"`
-	WorkerRootVolumeSize     int               `yaml:"workerRootVolumeSize"`
-	WorkerSpotPrice          string            `yaml:"workerSpotPrice"`
-	VPCID                    string            `yaml:"vpcId"`
-	RouteTableID             string            `yaml:"routeTableId"`
-	VPCCIDR                  string            `yaml:"vpcCIDR"`
-	InstanceCIDR             string            `yaml:"instanceCIDR"`
-	ControllerIP             string            `yaml:"controllerIP"`
+	ClusterName              string `yaml:"clusterName"`
+	ExternalDNSName          string `yaml:"externalDNSName"`
+	KeyName                  string `yaml:"keyName"`
+	Region                   string `yaml:"region"`
+	AvailabilityZone         string `yaml:"availabilityZone"`
+	ReleaseChannel           string `yaml:"releaseChannel"`
+	ControllerInstanceType   string `yaml:"controllerInstanceType"`
+	ControllerRootVolumeSize int    `yaml:"controllerRootVolumeSize"`
+	WorkerCount              int    `yaml:"workerCount"`
+	WorkerInstanceType       string `yaml:"workerInstanceType"`
+	WorkerRootVolumeSize     int    `yaml:"workerRootVolumeSize"`
+	WorkerSpotPrice          string `yaml:"workerSpotPrice"`
+	VPCID                    string `yaml:"vpcId"`
+	RouteTableID             string `yaml:"routeTableId"`
+	VPCCIDR                  string `yaml:"vpcCIDR"`
+	InstanceCIDR             string `yaml:"instanceCIDR"`
+	ControllerIP             string `yaml:"controllerIP"`
+	EtcdEndpoints            string `yaml:"etcdEndpoints"`
+	EtcdEndpoint             string
+	EtcdSecurityGroupId      string            `yaml:"etcdSecurityGroupId"`
 	PodCIDR                  string            `yaml:"podCIDR"`
 	ServiceCIDR              string            `yaml:"serviceCIDR"`
 	DNSServiceIP             string            `yaml:"dnsServiceIP"`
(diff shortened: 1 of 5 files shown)


colhom:
@orkhanM there are immediate plans to use a separate etcd cluster behind an ELB, instead of running etcd locally on the controller. In that case, we will only use a single etcd endpoint value.

chancez:
It would probably still be useful to support this feature for when someone's already got an existing etcd cluster and wants to not-deploy etcd, but use an existing one.

multi-node/aws/pkg/config/config.go
```diff
@@ -104,7 +107,16 @@ const (
 
 func (c Cluster) Config() (*Config, error) {
 	config := Config{Cluster: c}
-	config.ETCDEndpoints = fmt.Sprintf("http://%s:2379", c.ControllerIP)
```
chancez: this whole thing should probably just take a full url, not just an IP. Being able to specify different ports, and http vrs https is important, so I think you probably should only retain the `etcdEndpoints := strings.Split(c.EtcdEndpoints, ",")` and then parse each URL to make sure it's valid, but not change anything.

Something like:
```
etcdEndpoints := strings.Split(c.EtcdEndpoints, ",")
for i, ep := range etcdEndpoints {
    // Check for a valid url, but leave it as-is otherwise.
    _, err := url.Parse(ep); 
    if err != nil {
        return nil, err
    }

}
config.EtcdEndpoint = etcdEndpoints[0]
config.EtcdEndpoints = c.EtcdEndpoints
```

Then below for the `curl` it could just be
```
{{ .EtcdEndpoint }}/v2/keys/coreos.com/network/config?prevExist=false
```

orkhanM:
Hey, thanks for the feedback. I'll work on getting those changes in. Do you think that I should add the option to expose the etcd clusters security group via http://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html. Right now with the given code it needs to be done manually before the controller and worker instances spin up.

robszumski:
Another thing to note is that even if you are running an etcd proxy locally, it can take in a list of etcd servers to connect to and will manage the health of those endpoints very similar to an ELB. An ELB is nice because you get a stable hostname/IP to code against, but it's not the _only_ way to achieve that outcome.

orkhanM:
Hey @chancez, I've made the changes you suggested, looks much cleaner imo. Let me know if you'd like me to add the ELB portion or the security group ingress rules I mentioned in my previous comment (or anything else =P).

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -55,6 +55,9 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # IP Address for controller in Kubernetes subnet
 # controllerIP: 10.0.0.50
 
+# Comma seperated list of Etcd endpoints defaults to http://10.0.0.50:2379 
+# etcdEndpoints: "http://10.0.0.50:2379"
```
colhom: Can we put a note here that the "default etcd configuration" requires that etcdEndpoints = `http://<controllerIP>:2379`? In the event the controllerIP is changed to non-default, I want the need to change `etcdEndpoints` to be obvious.

orkhanM: Good point, on it.

orkhanM:
@robszumski, Could you review [https://github.com/orkhanM/coreos-kubernetes/pull/4](https://github.com/orkhanM/coreos-kubernetes/pull/4) and let me know if it's something you'd like to get merged into this PR. Currently the security group rules need to be added manually before the worker and controller nodes spin up.

robszumski:
@orkhanM My comments were assuming we were setting up multiple control plane containers, but we're currently not. I think you're ok from my side on that.

Which security groups do you need to set up manually? The one applied to etcd machines that you should have already launched?

orkhanM:
@robszumski yep

orkhanM:
@colhom Anything else here need doing?

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -66,6 +66,13 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # IP Address for controller in Kubernetes subnet
 # controllerIP: 10.0.0.50
 
+# Comma seperated list of Etcd endpoints. By default this needs to be set to "http://<controllerIP>:2379" 
+# if custom etcdEndpoints are specified then a security group id can also be specified which will expose
+# the etcd security group to the worker & controller instances.
+# etcdEndpoints: "http://10.0.0.50:2379"
+# etcdSecurityGroupId: ""
```
cgag: Any reason not to continue using localhost here to avoid accidentally updating controllerIP, but not this value?

orkhanM: @cgag because localhost would prevent the other k8s nodes from being able to reach the etcd endpoint. 

The first endpoint from the EtcdEndpoints list is used to seed both the worker and controller cloud-config.

multi-node/aws/pkg/config/config.go
```diff
@@ -116,7 +120,15 @@ const (
...
+	etcdEndpoints := strings.Split(c.EtcdEndpoints, ",")
+	for i := range etcdEndpoints {
+		_, err := url.Parse(etcdEndpoints[i])
+		if err != nil {
+			return nil, err
+		}
+	}
```
cgag: I think this validation should be in config.valid(), as well as a check that etcdEndpoints isn't empty, since we do do `etcdEndpoints[0]` below.

cgag: I think this validation should be in config.valid(), as well as a check that etcdEndpoints isn't empty, since we do do etcdEndpoints[0] below.

multi-node/aws/pkg/config/templates/cluster.yaml
cgag: Oh, yeah, woops. Lost site of the forest there.

multi-node/aws/pkg/config/config.go
cgag: Somewhat nit-picky but I'd prefer if this loop didn't rely on the index and did something like `for _, endpoint := range etcdEndpoints` as @chancez did in his example.

orkhanM: Indeed, it's neater this way. Thanks!

orkhanM:
Anything else need doing for this to get merged in?

multi-node/aws/pkg/config/config.go
```diff
@@ -141,6 +147,15 @@ func (c Cluster) Config() (*Config, error) {
 	return &config, nil
 }
 
+func (c Cluster) CustomEtcdEndpoint() bool {
```
colhom: I would call this function `EtcdOnController` and then `return host == c.ControllerIP`.

multi-node/aws/pkg/config/config.go
```diff
@@ -116,7 +120,9 @@ const (
 
 func (c Cluster) Config() (*Config, error) {
 	config := Config{Cluster: c}
-	config.ETCDEndpoints = fmt.Sprintf("http://%s:2379", c.ControllerIP)
+
+	config.EtcdEndpoint = strings.Split(c.EtcdEndpoints, ",")[0]
+	config.EtcdEndpoints = c.EtcdEndpoints
```
colhom: this is redundant, as the `Config.EtcdEndpoints` field is inherited from the `Cluster` type

multi-node/aws/pkg/config/config.go
```diff
@@ -81,23 +82,26 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
...
+	WorkerInstanceType       string `yaml:"workerInstanceType"`
+	WorkerRootVolumeSize     int    `yaml:"workerRootVolumeSize"`
+	WorkerSpotPrice          string `yaml:"workerSpotPrice"`
+	VPCID                    string `yaml:"vpcId"`
+	RouteTableID             string `yaml:"routeTableId"`
+	VPCCIDR                  string `yaml:"vpcCIDR"`
+	InstanceCIDR             string `yaml:"instanceCIDR"`
+	ControllerIP             string `yaml:"controllerIP"`
+	EtcdEndpoints            string `yaml:"etcdEndpoints"`
+	EtcdEndpoint             string
```
colhom: This should only be a member of the `Cluster` struct. The `Config` struct is for things that get parsed directly out of the yaml

colhom:
@orkhanM this is a very useful PR, but I think we're going to have to wait on this functionality, as we will be shipping a discrete ETCD cluster in kube-aws in the near future. This definitely increases the config logic complexity, with: 
* the need to determine if we can infer etcd to be on the controller
* manually listing the etcd security group and linking to it
* 'ETCDEndpoints`, 'EtcdEndpoints` and `EtcdEndpoint` all co-existing in the same type, etc.

A lot of this complexity will not be presented for a majority of users once this architectural change is made.

PR #395, opened 2016-04-11
Windows host os support

tetralix:
Added native windows commands support for certificate generation and fixed CRLF issue for temporary files creation in Vagrantfile

### DIFF
--- a/lib/init-ssl-ca.ps1
+++ b/lib/init-ssl-ca.ps1
@@ -1,0 +1,34 @@
+# define location of openssl binary manually since running this
+# script under Vagrant fails on some systems without it
+$OPENSSL="openssl"
+
+function usage {
+	$script=split-path $MyInvocation.PSCommandPath -Leaf
+    write-host "USAGE: $script <output-dir>"
+    write-host "  example: $script .\ssl\ca.pem"
+}
+
+if($args[0] -eq $null){
+	usage
+	return
+}
+
+$OUTDIR=$args[0]
+
+if(! (test-path $OUTDIR)){
+	write-error "ERROR: output directory does not exist:  $OUTDIR"
+	return
+}
+
+$OUTFILE="$OUTDIR\ca.pem"
+
+if(test-path $OUTFILE){
+	return
+}
+
+#Needed to avoid "unable to write 'random state'" error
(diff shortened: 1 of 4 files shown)


colhom:
\cc @aaronlevy

aaronlevy:
@tetralix thank you for this contribution.

My immediate concern with a change like this is that it will be almost impossible for any of the maintainers of this repo to adequately test / review / maintain these changes given we're not using windows directly. 

Somewhat of a side question (that I'm not super familiar with), but with the recent announcement of a full [bash shell in windows 10](https://blogs.windows.com/buildingapps/2016/03/30/run-bash-on-ubuntu-on-windows/), would that be a possibility instead to use these bash scripts natively?

tetralix:
Good question. I must test it on Win10

dpheasant:
@aaronlevy @tetralix ... I should have checked for pull requests first, but I've essentially made the same changes locally. Only difference is instead of using powershell to script certificate generation, I used python. It provides a "quick-and-dirty" os-independent solution at the expense of requiring python. Thoughts? I can submit a PR for if desired.

As for Bash on Windows. I'm not sure how long it will be until MS releases that to the masses. It's currently in beta preview and you must enable developer options/modes that some corporate individuals may not be able to.

michaelmitchell:
The method I use to get this working on windows is to wrap the system calls in git-bash.exe which seems to work perfectly. I also had to add the binmode flag on the Tempfiles in order to get a cluster working properly.

```
system("\"C:\\Program Files\\Git\\git-bash.exe\" -c \"mkdir -p ssl && ./../../lib/init-ssl-ca ssl\"") or abort ("failed generating SSL artifacts")
```

colhom:
@aaronlevy @cgag : do we plan on supporting windows? I'm leaning towards just recommending cygwin for these cases.

colhom:
@tetralix thank you for contributing your work. Unfortunately I'm going to close this PR, as this is not something that we the bandwidth for.

We have little Windows representation both here at CoreOS and in our contributor base for this project. I feel that the effort to build, test and maintain Windows compatibility will not be commensurate with the benefit of supporting a native Windows environment, when well-established alternatives like cygwin and virtual machines already provide a reasonable way to use kube-aws on Windows.

PR #396, opened 2016-04-11
kube-aws: be specific about how long resource creation takes.

cgag:
I've timed it several times with several cluster sizes and it's always
taken very close to 5 minutes.


cgag:
If you don't like this, I can fix the fact that it says "make take" instead of "may take" in another pull request.

colhom:
LGTM

PR #397, opened 2016-04-11
kube-aws: increase heapster memory limit to 650Mi

### LINKED ISSUE #358: Increase heapster/influxdb memory limits
On both of our clusters we were seeing these pods getting restarted a large amount even with a small number of nodes. The restarts were because the pods were being killed via the OOMKiller as they hit their memory limits.

Bumping them both to 650Mb each seemed to have stabilized them. This is for about 60 pods right now.

cgag:
fixes #358

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -510,7 +510,7 @@ write_files:
                     "resources": {
                       "limits": {
                         "cpu": "100m",
-                        "memory": "208Mi"
+                        "memory": "650Mi"
                       },
                       "requests": {
                         "cpu": "100m",


pbx0:
Can you make this change to `single-node/user-data` and `multi-node/generic/controller-install.sh` as well?

Also, any method for arriving at 650Mi? Sounds like that was what worked for chancez. Upstream uses 200Mi + 4Mib per node. That logic never got translated into our scripts so it seems safe to go on the high side here. That would have a cluster covered up to about 100 nodes. We'll have to make this number scale with nodes somehow it sounds like.

For reference: https://github.com/kubernetes/kubernetes/blob/v1.2.2/cluster/addons/cluster-monitoring/standalone/heapster-controller.yaml

pbx0:
Actually looks like the latest heapster controller will scale its resources based on nodes: 
https://github.com/kubernetes/contrib/tree/master/addon-resizer.

So this LGTM to commit now, but I'll soon change it back to 200MiB (maybe 250MiB to be safe) and let its memory automatically scale by 4MiB when we update to v1.2.2.

PR #402, opened 2016-04-12
kube-aws: update docs to reflect route53 integration

colhom:
\cc @robszumski @cgag

### DIFF
--- a/Documentation/kubernetes-on-aws.md
+++ b/Documentation/kubernetes-on-aws.md
@@ -82,6 +82,8 @@ Select a DNS hostname where the cluster's API will be accessible. This informati
 
 When CloudFormation finishes creating your cluster, your controller will expose the TLS-secured API via a public IP address. You will need to create an A record for the DNS hostname which lists the IP address of the API. You can find this IP address later via `kube-aws status`.
 
+`kube-aws` can be optionally be configured to automatically create an A record in an existing route53 hosted zone.
+
 #### KMS Key
 
 [Amazon KMS](http://docs.aws.amazon.com/kms/latest/developerguide/overview.html) keys are used to encrypt and decrypt cluster TLS assets. If you already have a KMS Key that you would like to use, you can skip this step.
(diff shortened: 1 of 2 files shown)


Documentation/kubernetes-on-aws.md
```diff
@@ -156,6 +158,20 @@ You can now customize your cluster by editing asset files:
 
 You can also now check the `my-cluster` asset directory into version control if you desire. The contents of this directory are your reproducible cluster assets. Please take care not to commit the `my-cluster/credentials` directory, as it contains your TLS secrets. If you're using git, the `credentials` directory will already be ignored for you.
 
+#### Route53 Host Record (optional)
+
+`kube-aws` can optionally create an A record for the controller IP in an existing hosted zone.
```
robszumski: Can't there be multiple controllers? What happens?

Documentation/kubernetes-on-aws.md
```diff
@@ -178,6 +194,8 @@ If deploying a production Kubernetes cluster, consider establishing PKI independ
 
 #### Configure your DNS
 
+This step is not necessary if `createRecordSet` is true. In this case, a host record has already been created for you.
```
robszumski: Should the above section appear here? Have a section for "automatic via route53" vs "manual"?

robszumski: Hmm, I guess I can see how this is separated between during the install vs post-install steps...

Documentation/kubernetes-on-aws.md
colhom: As of now, no. When that time comes, there will be a few other things that will have to change as well.

colhom:
@robszumski can I get an LGTM ;) ?

Documentation/kubernetes-on-aws.md
robszumski: Idea: instead of referencing the feature by its config name, we use Route 53. Like:

> If you configured Route 53 settings in your configuration above, a host record has already been created for you.

Probably need an "otherwise" in the line below:

> **Otherwise**, navigate to the...

robszumski:
One last comment then it looks good!

PR #403, opened 2016-04-12
kube-aws: set default recordSetTTL to 300

### LINKED ISSUE #400: default recordSetTTL is 0
If `recordSetTTL` is left commented out in cluster.yaml, `kube-aws up` fails:

```sh
Creating AWS resources. This should take around 5 minutes.
Error: Error creating cluster: TTL must be at least 1 second
```

@cgag in order to fix this, we need to:
* set the default TTL value [here](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/config.go#L31)

* validate the TTL value [here](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/config.go#L329) as part of config validation

* add test cases for default value, invalid value, and valid value to `pkg/config/config_test.go`

cgag:
Validate that recordSetTTL is positive, and that HostedZone is not
blank in config rather than cluster since they don't depend on any
external state.

fixes #400

### DIFF
--- a/multi-node/aws/pkg/config/config_test.go
+++ b/multi-node/aws/pkg/config/config_test.go
@@ -34,6 +34,13 @@ vpcId: vpc-xxxxx
 routeTableId: rtb-xxxxxx
 `, `
 vpcId: vpc-xxxxx
+`, `
+createRecordSet: false
+hostedZone: ""
+`, `
+createRecordSet: true
+recordSetTTL: 400
+hostedZone: core-os.net
 `,
 }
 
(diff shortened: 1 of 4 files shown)


multi-node/aws/pkg/config/config_test.go
```diff
@@ -34,6 +34,11 @@ vpcId: vpc-xxxxx
 routeTableId: rtb-xxxxxx
 `, `
 vpcId: vpc-xxxxx
+`, `
+createRecordSet: false
+hostedZone: ""
+`, `
+recordSetTTL: 400
```
colhom: should this case maybe be invalid?  a non-default `recordSetTTL` with `createRecordSet=false` is non-sensical?

cgag: hmm yeah, there's no good reason to ever have that combo, i'll make it invalid and improve the tests

PR #404, opened 2016-04-12
kube-aws: don't add trailing dot to empty dns names

cgag:
If you add the trailing dot to an empty hostname, the validation
that hostname isn't blank is worthless.

Basically none of the trailing dot stuff was being tested because it was
being added in ClusterFromFile rather than ClusterFromBytes, which is only
used in the live code. Also, the r53 dummy service was returning the
HostedZone and ResourceRecordSet without the final ., which would be
there in the real return values.

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -59,6 +59,16 @@ func ClusterFromFile(filename string) (*Cluster, error) {
 		return nil, fmt.Errorf("file %s: %v", filename, err)
 	}
 
+	return c, nil
+}
+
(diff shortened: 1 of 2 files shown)


multi-node/aws/pkg/config/config.go
```diff
@@ -491,6 +492,9 @@ func cidrOverlap(a, b *net.IPNet) bool {
 }
 
 func withTrailingDot(s string) string {
```
colhom: nit: I think calling this `ensureTrailingDot` would be more descriptive. I felt the need to check for myself that the trailing dot was only added if not there. Not a blocker, maybe separate PR or something.

colhom:
Makes sense to me, LGTM.

PR #406, opened 2016-04-13
kube-aws: cloudformation tags

orkhanM:
Currently there is no way to specify tags to apply to the created cloudformation stack. Many companies use tags to audit their AWS environment and do cost analysis. This pull request adds the ability to specify cloudformation resource tags as a list of key-value pairs (string -> string) in the cluster.yaml file

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -147,12 +147,20 @@ func (c *Cluster) Create(stackBody string) error {
 		return err
 	}
 
+	var tags []*cloudformation.Tag
+	for k, v := range c.StackTags {
+		key := k
+		value := v
+		tags = append(tags, &cloudformation.Tag{Key: &key, Value: &value})
+	}
+
 	cfSvc := cloudformation.New(c.session)
 	creq := &cloudformation.CreateStackInput{
 		StackName:    aws.String(c.ClusterName),
 		OnFailure:    aws.String("DO_NOTHING"),
 		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
 		TemplateBody: &stackBody,
+		Tags:         tags,
 	}
 
 	resp, err := cfSvc.CreateStack(creq)
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -82,33 +82,33 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 }
 
 type Cluster struct {
-	ClusterName              string `yaml:"clusterName"`
-	ExternalDNSName          string `yaml:"externalDNSName"`
-	KeyName                  string `yaml:"keyName"`
-	Region                   string `yaml:"region"`
-	AvailabilityZone         string `yaml:"availabilityZone"`
-	ReleaseChannel           string `yaml:"releaseChannel"`
-	ControllerInstanceType   string `yaml:"controllerInstanceType"`
-	ControllerRootVolumeSize int    `yaml:"controllerRootVolumeSize"`
-	WorkerCount              int    `yaml:"workerCount"`
-	WorkerInstanceType       string `yaml:"workerInstanceType"`
-	WorkerRootVolumeSize     int    `yaml:"workerRootVolumeSize"`
-	WorkerSpotPrice          string `yaml:"workerSpotPrice"`
-	VPCID                    string `yaml:"vpcId"`
-	RouteTableID             string `yaml:"routeTableId"`
-	VPCCIDR                  string `yaml:"vpcCIDR"`
-	InstanceCIDR             string `yaml:"instanceCIDR"`
-	ControllerIP             string `yaml:"controllerIP"`
-	PodCIDR                  string `yaml:"podCIDR"`
-	ServiceCIDR              string `yaml:"serviceCIDR"`
-	DNSServiceIP             string `yaml:"dnsServiceIP"`
-	K8sVer                   string `yaml:"kubernetesVersion"`
-	HyperkubeImageRepo       string `yaml:"hyperkubeImageRepo"`
-	KMSKeyARN                string `yaml:"kmsKeyArn"`
-
-	CreateRecordSet bool   `yaml:"createRecordSet"`
-	RecordSetTTL    int    `yaml:"recordSetTTL"`
-	HostedZone      string `yaml:"hostedZone"`
+	ClusterName              string            `yaml:"clusterName"`
+	ExternalDNSName          string            `yaml:"externalDNSName"`
+	KeyName                  string            `yaml:"keyName"`
+	Region                   string            `yaml:"region"`
+	AvailabilityZone         string            `yaml:"availabilityZone"`
+	ReleaseChannel           string            `yaml:"releaseChannel"`
+	ControllerInstanceType   string            `yaml:"controllerInstanceType"`
+	ControllerRootVolumeSize int               `yaml:"controllerRootVolumeSize"`
+	WorkerCount              int               `yaml:"workerCount"`
+	WorkerInstanceType       string            `yaml:"workerInstanceType"`
+	WorkerRootVolumeSize     int               `yaml:"workerRootVolumeSize"`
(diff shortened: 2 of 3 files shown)


multi-node/aws/pkg/config/config.go
```diff
@@ -82,33 +82,33 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
...
+	PodCIDR                  string              `yaml:"podCIDR"`
+	ServiceCIDR              string              `yaml:"serviceCIDR"`
+	DNSServiceIP             string              `yaml:"dnsServiceIP"`
+	K8sVer                   string              `yaml:"kubernetesVersion"`
+	HyperkubeImageRepo       string              `yaml:"hyperkubeImageRepo"`
+	KMSKeyARN                string              `yaml:"kmsKeyArn"`
+	CreateRecordSet          bool                `yaml:"createRecordSet"`
+	RecordSetTTL             int                 `yaml:"recordSetTTL"`
+	HostedZone               string              `yaml:"hostedZone"`
+	Tags                     []map[string]string `yaml:"tags"`
```
colhom: These tags will only apply to the stack itself (not constituent resources), so this should have a less-generic name, like `StackTags` or something.

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -81,3 +81,8 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 
 # Hyperkube image repository to use.
 # hyperkubeImageRepo: quay.io/coreos/hyperkube
+
+# AWS Tags for cloudformation stack (formatted as a list of key-value pairs)
+#tags:
+#- Name: "Kubernetes" 
+#- Environment: "Production"
```
colhom: Is there any reason this is a list of key/value pair objects, and not just an object?

orkhanM: I decided to go with the list so that an arbitrary amount of tags could be specified. 

Key value pairs because that's what the Tag constructor accepts.
https://godoc.org/github.com/aws/aws-sdk-go/service/cloudformation#Tag

multi-node/aws/pkg/config/config.go
orkhanM: Thats what I originally thought too but actually "stack-level tags" are propagated to resources that AWS CloudFormation supports.

http://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

I can change it though.

multi-node/aws/pkg/config/templates/cluster.yaml
colhom: I'll clarify a little more:

Why not structure `tags` like this:
```yaml
tags:
  Name: "Kubernetes"
  Environment: "Production"
  Owner: "colhom"
```

orkhanM: _ahhh_ I hadn't thought of that. That simplifies [this section](https://github.com/coreos/coreos-kubernetes/pull/406/files#diff-57c0e4221c5add72bb835d26badd2241R151) significantly. Testing locally now.

multi-node/aws/pkg/config/config.go
```diff
@@ -108,7 +108,7 @@ type Cluster struct {
 	CreateRecordSet          bool                `yaml:"createRecordSet"`
 	RecordSetTTL             int                 `yaml:"recordSetTTL"`
 	HostedZone               string              `yaml:"hostedZone"`
-	Tags                     []map[string]string `yaml:"tags"`
+	StackTags                []map[string]string `yaml:"tags"`
```
colhom: change the yaml tag to `stackTags` as well

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -81,3 +80,8 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 
 # Hyperkube image repository to use.
 # hyperkubeImageRepo: quay.io/coreos/hyperkube
+
+# AWS Tags for cloudformation stack (formatted as a list of key-value pairs)
+#stackTags:
```
colhom: @orkhanM looks good. Can you squash up w/ a commit message along the lines of `kube-aws: cloudformation stack tag functionality`

colhom:
@orkhanM looks good. Can you squash up w/ a commit message along the lines of `kube-aws: cloudformation stack tag functionality`

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -147,12 +147,22 @@ func (c *Cluster) Create(stackBody string) error {
 		return err
 	}
 
+	tags := make([]*cloudformation.Tag, len(c.StackTags))
+	i := 0
+	for k, v := range c.StackTags {
+		key := k
+		value := v
```
cgag: Why not  `for key, value := range c.StackTags {`?  Or you could just use k and v below.  I'm generally not a fan of introducing variables unless they're necessary.

Similarly I'd prefer to drop `i` and instead write `tags = append(tags, &cloudFormation..Tag{...}`.

Otherwise this looks great to me, thanks :)

colhom: You need to take the reference of the variables, and loops in golang re-use
variables across iterations. Need to copy them in this case
On Apr 15, 2016 10:43 AM, "Curtis Gagliardi" <notifications@github.com>
wrote:

In multi-node/aws/pkg/cluster/cluster.go
<https://github.com/coreos/coreos-kubernetes/pull/406#discussion_r59912137>:

> @@ -147,12 +147,22 @@ func (c *Cluster) Create(stackBody string) error {
>  		return err
>  	}
>
> +	tags := make([]*cloudformation.Tag, len(c.StackTags))
> +	i := 0
> +	for k, v := range c.StackTags {
> +		key := k
> +		value := v

Why not for key, value := range c.StackTags {? Or you could just use k and
v below. I'm generally not a fan of introducing variables unless they're
necessary.

Similarly I'd prefer to drop i and instead write tags = append(tags,
&cloudFormation..Tag{...}.

Otherwise this looks great to me, thanks :)

—
You are receiving this because you commented.
Reply to this email directly or view it on GitHub
<https://github.com/coreos/coreos-kubernetes/pull/406/files/6442f35d68173d513c36e8be8c6f45e5cd930287#r59912137>

orkhanM: exactly, unfortunately theres no way to instantiate a string inline as there is no copy constructor that I found. I did change the cloudFormation..Tag{...} assignment to use append rather then the assignment operator.

cgag: bah forgot they were string pointers not strings, was thinking they were being copied when tag was created, you're right

colhom:
@orkhanM squash this up and we should be good to merge.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -147,12 +147,20 @@ func (c *Cluster) Create(stackBody string) error {
 		return err
 	}
 
+	tags := make([]*cloudformation.Tag, 0)
```
colhom: get rid of the second 0 argument, it's not necessary if we're using the append approach.

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -81,3 +80,8 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 
 # Hyperkube image repository to use.
 # hyperkubeImageRepo: quay.io/coreos/hyperkube
+
+# AWS Tags for cloudformation stack (formatted as a list of key-value pairs)
```
colhom: It's no longer formatted as a list of key-value pairs

colhom:
@orkhanM we should really have unit tests for this, but there is significant restructuring of how `cluster.Create()` function works in order for that to be possible. I'll \cc you when we have another PR which restructures the code in order to make it testable. (ref #415)

I have two more nits for you to address before we merge. Sorry about that.

Thanks for the contribution!

orkhanM:
@colhom sounds good, I've made the changes you requested and re-squashed. Let me know when you can merge this in.

colhom:
Thanks @orkhanM

PR #414, opened 2016-04-15
kube-aws: don't append trailing dot to externalDNSName

cgag:
We previously automatically appended trailing dots to externalDNSName
in the config struct, which was then reflected in kubeconfig.
This caused some problems with people using /etc/hosts, since they would
add the externalDNSName to their hosts file without the trailing dot,
and /etc/hosts lookups don't append one.

Now we only add trailing dots on the fly, whenever they're necessary to
interact with the AWS API, never anywhere persistent.

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -70,12 +70,9 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 		return nil, fmt.Errorf("failed to parse cluster: %v", err)
 	}
 
-	// HostedZone needs to end with a '.'
-	c.HostedZone = withTrailingDot(c.HostedZone)
-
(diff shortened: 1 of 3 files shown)


multi-node/aws/pkg/config/config.go
```diff
@@ -511,7 +508,7 @@ func cidrOverlap(a, b *net.IPNet) bool {
 	return a.Contains(b.IP) || b.Contains(a.IP)
 }
 
-func withTrailingDot(s string) string {
+func WithTrailingDot(s string) string {
```
colhom: Nit: At some point we should probably move these weird little dns-related helper methods to a separate "utils" type lib package. Another example would be `isSubDomain()`. Non-blocker for now.

colhom:
LGTM

PR #419, opened 2016-04-17
update cluster.yaml

Thermi:
Added necessary information for the sizing of podCIDR.

### DIFF
--- a/multi-node/aws/pkg/config/templates/cluster.yaml
+++ b/multi-node/aws/pkg/config/templates/cluster.yaml
@@ -70,6 +70,9 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # serviceCIDR: "10.3.0.0/24"
 
 # CIDR for all pod IP addresses
+# this subnet must be at least as large as serviceCIDR * workerCount
+# If that is not the case, flannel will refuse to start on some workers,
+# as it can not elect a subnet (it will want at least a /24) for local use.
 # podCIDR: "10.2.0.0/16"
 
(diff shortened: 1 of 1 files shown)


multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -70,6 +70,9 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # serviceCIDR: "10.3.0.0/24"
 
 # CIDR for all pod IP addresses
+# this subnet must be at least as large as serviceCIDR * workerCount
```
aaronlevy: The serviceCIDR shouldn't have anything to do with the size of the podCIDR (they just cannot overlap).

aaronlevy:
I think the intention here is to give some more detail about what might be expected for the podCIDR. The flannel default is to assign a /24 per node, so maybe we can start there.

What about something like:

> The podCIDR will contain the IP addresses assigned to all pods in the cluster. Flannel is used to configure the podNetwork and will by default assign a /24 out of the podCIDR to each node in your cluster.

/cc @joshix @robszumski

Thermi:
IMHO There should be a description of what happens when the subnet is too small. I had things ranging from no working node to 3 of 4 nodes working to all nodes working. The obvious thing to take away from that observation is that nodes where flannel can not pick a subnet can not join the cluster. This should be noted, because it is not obvious.

aaronlevy:
I'm going to close this for now, but please feel free to open a PR (or re-open this one) with updated verbiage about the default subnet sizing.

PR #421, opened 2016-04-18
Testable stack creation

### LINKED ISSUE #415: Stack creation codepath is not testable
Currently `pkg/cluster/Cluster.Create()` is not testable via mocked AWS service interfaces. The logic that parses tags and calls out to `CreateStack` needs to extracted into a private function which receives a mockable cloudformation service interface.

colhom:
Fixes #415 
 
\cc @aaronlevy @cgag @orkhanM

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster_test.go
+++ b/multi-node/aws/pkg/cluster/cluster_test.go
@@ -2,10 +2,12 @@ package cluster
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/config"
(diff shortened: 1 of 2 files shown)


multi-node/aws/pkg/cluster/cluster_test.go
```diff
@@ -281,3 +283,102 @@ hostedZone: staging.core-os.net
...
+	if len(cfSvc.ExpectedTags) != len(req.Tags) {
+		return nil, fmt.Errorf(
+			"expected tag count does not match supplied tag count\nexpected=%v, supplied=%v",
+			cfSvc.ExpectedTags,
+			req.Tags,
+		)
+	}
+
+	matchCnt := 0
+	for _, eTag := range cfSvc.ExpectedTags {
```
aaronlevy: Not sure if it matters, but this would still pass in the case of duplicate tags. Pseudo passing test case:

```
want = []tag{
  {k:foo, v:bar},
  {k:foo, v:bar},
}
got = []tag{k:foo, v:bar}
```

Also, maybe reflect.DeepEqual() instead of manually checking struct fields?

aaronlevy:
Minor comment but otherwise lgtm

multi-node/aws/pkg/cluster/cluster_test.go
colhom: We're going for set equality on the two tag arrays. We could do a stable sort and then DeepEqual() the two, but I feel that's more to go wrong there. Opinions?

aaronlevy: Not strong ones. Up to you.

PR #424, opened 2016-04-19
Add Kubernetes Network Policy

tomdee:
I've been working on a change for the last few weeks to add support for Network Policy to this repo. The policy is implemented by Calico. Some information on what network policy is can be found [here](https://github.com/projectcalico/calico-containers/blob/master/docs/cni/kubernetes/NetworkPolicy.md)  and [here](https://github.com/projectcalico/calico-containers/blob/master/docs/cni/kubernetes/stars-demo/README.md)

The progression of the network policy API in kubernetes is being tracked at https://github.com/kubernetes/kubernetes/issues/22469

I'm aware that the CNI plugins aren't being downloaded in the best way - I'm currently waiting for a CNI release which should be happening very soon. I'll update the PR and retest as soon as binaries become available.

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -33,11 +33,12 @@ $ sudo chmod 600 /etc/kubernetes/ssl/*-key.pem
 $ sudo chown root:root /etc/kubernetes/ssl/*-key.pem
 ```
 
-### flannel Configuration
+### Network Configuration
+Networking is provided by Flannel and Calico.
+* [flannel][flannel-docs] provides a software-defined overlay network for routing traffic to/from the [pods][pod-overview]
+* [Calico][calico-docs] secures the overlay network by restricting traffic to/from the pods based on fine-grained network policy.
 
-[flannel][flannel-docs] provides a key Kubernetes networking capability &mdash; a software-defined overlay network to manage routing of the [Pod][pod-overview] network.
-
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently of flannel, then the flannel parts of this guide can be skipped. In this case, Calico may still be used for providing network policy. See [Kubernetes networking](kubernetes-networking.md) for more detail.
 
 We will configure flannel to source its local configuration in `/etc/flannel/options.env` and cluster-level configuration in etcd. Create this file and edit the contents:
 
@@ -50,7 +51,6 @@ We will configure flannel to source its local configuration in `/etc/flannel/opt
 FLANNELD_IFACE=${ADVERTISE_IP}
 FLANNELD_ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
 ```
-
 Next create a [systemd drop-in][dropins], which is a method for appending or overriding parameters of a systemd unit. In this case we're appending two dependency rules. Create the following drop-in, which will use the above configuration when flannel starts:
 
 **/etc/systemd/system/flanneld.service.d/40-ExecStartPre-symlink.conf**
@@ -60,6 +60,7 @@ Next create a [systemd drop-in][dropins], which is a method for appending or ove
 ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 ```
 
+[calico-docs]: https://github.com/projectcalico/calico-containers/tree/v0.19.0/docs/cni/kubernetes
 [flannel-docs]: https://coreos.com/flannel/docs/latest/
 [pod-overview]: https://coreos.com/kubernetes/docs/latest/pods.html
 [service-overview]: https://coreos.com/kubernetes/docs/latest/services.html
@@ -68,7 +69,7 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 
 In order for flannel to manage the pod network in the cluster, Docker needs to be configured to use it. All we need to do is require that flanneld is running prior to Docker starting.
 
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
 
 We're going to do this, like we proceeded above with flannel, using a [systemd drop-in][dropins]:
 
@@ -88,11 +89,14 @@ The [kubelet](http://kubernetes.io/docs/admin/kubelet.html) is the agent on each
 
 On the master node, the kubelet is configured to communicate with the API server, but not register for cluster work, as shown in the `--register-schedulable=false` line in the YAML excerpt below. This prevents user pods being scheduled on the master nodes, and ensures cluster work is routed only to task-specific worker nodes.
 
+When using Calico, the kubelet is configured to use the Container Networking Interface (CNI) standard for networking. This makes Calico aware of each pod that is created and allows it to network the pods into the flannel overlay. Both flannel and Calico communicate via CNI interfaces to ensure the correct IP range (managed by flannel) is used for each node.
+
 Note that the kubelet running on a master node may log repeated attempts to post its status to the API server. These warnings are expected behavior and can be ignored. Future Kubernetes releases plan to [handle this common deployment consideration more gracefully](https://github.com/kubernetes/kubernetes/issues/14140#issuecomment-142126864).
 
 * Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
 * Replace `${DNS_SERVICE_IP}`
-* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release
+* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release. If using Calico a version that includes CNI binaries should be used. e.g. `v1.2.2_coreos_cni.0`
+* Replace `${NETWORK_PLUGIN}` with `cni` if using Calico. Otherwise just leave it blank.
 
 **/etc/systemd/system/kubelet.service**
 
@@ -103,6 +107,8 @@ ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
 Environment=KUBELET_VERSION=${K8S_VER}
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=http://127.0.0.1:8080 \
+  --network-plugin-dir=/etc/kubernetes/cni/net.d \
+  --network-plugin=${NETWORK_PLUGIN} \
   --register-schedulable=false \
   --allow-privileged=true \
   --config=/etc/kubernetes/manifests \
@@ -154,6 +160,7 @@ spec:
     - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
     - --client-ca-file=/etc/kubernetes/ssl/ca.pem
     - --service-account-key-file=/etc/kubernetes/ssl/apiserver-key.pem
+    - --runtime-config=extensions/v1beta1=true,extensions/v1beta1/thirdpartyresources=true
     ports:
     - containerPort: 443
       hostPort: 443
@@ -270,7 +277,7 @@ spec:
 
 ### Set Up the kube-scheduler Pod
 
-The scheduler is the last major piece of our master node. It monitors the API for unscheduled pods, finds them a machine to run on, and communicates the decision back to the API.
+The scheduler monitors the API for unscheduled pods, finds them a machine to run on, and communicates the decision back to the API.
 
 Create File `/etc/kubernetes/manifests/kube-scheduler.yaml`:
 
@@ -300,6 +307,115 @@ spec:
       initialDelaySeconds: 15
       timeoutSeconds: 1
 ```
+
+### Set Up Calico Node Container
+
+The Calico node container runs on all hosts, including the master node. It performs two functions:
+* Connects containers to the flannel overlay network, which enables the "one IP per pod" concept.
+* Enforces network policy created through the Kubernetes policy API, ensuring pods talk to authorized resources only.
+
+This step can be skipped if not using Calico.
+
+When creating `/etc/systemd/system/calico-node.service`:
+
+* Replace `${ETCD_ENDPOINTS}`
+* Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
+
+**/etc/systemd/system/calico-node.service**
+
+```yaml
+[Unit]
+Description=Calico per-host agent
+Requires=network-online.target
+After=network-online.target
+
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --stage1-from-dir=stage1-fly.aci \
+--volume=modules,kind=host,source=/lib/modules,readOnly=false \
+--mount=volume=modules,target=/lib/modules \
+--trust-keys-from-https quay.io/calico/node:v0.19.0
+
+KillMode=mixed
+Restart=always
+TimeoutStartSec=0
+
+[Install]
+WantedBy=multi-user.target
+```
+
+### Set Up the policy-agent Pod
+
+The policy agent is the last major piece of the master node. It monitors the API for changes related to network policy and configures Calico to implement that policy. This step can be skipped if not using Calico.
+
+When creating `/etc/kubernetes/manifests/policy-agent.yaml`:
+
+* Replace `${ETCD_ENDPOINTS}` with the same endpoints used in calico-node.service, above.
+
+**/etc/kubernetes/manifests/policy-agent.yaml**
+
+```yaml
+apiVersion: v1
+kind: Pod 
+metadata:
+  name: calico-policy-agent
+  namespace: calico-system 
+spec:
+  hostNetwork: true
+  containers:
+    # The Calico policy agent.
+    - name: k8s-policy-agent
+      image: calico/k8s-policy-agent:v0.1.4
+      env:
+        - name: ETCD_ENDPOINTS
+          value: "${ETCD_ENDPOINTS}"
+        - name: K8S_API
+          value: "http://127.0.0.1:8080"
+        - name: LEADER_ELECTION 
+          value: "true"
+    # Leader election container used by the policy agent.
+    - name: leader-elector
+      image: quay.io/calico/leader-elector:v0.1.0
+      imagePullPolicy: IfNotPresent
+      args:
+        - "--election=calico-policy-election"
+        - "--election-namespace=calico-system"
+        - "--http=127.0.0.1:4040"
+```
+
+### Set Up the CNI config
+The kubelet reads the CNI configuration on startup and uses that to determine which CNI plugin to call. Create the following file which tells the kubelet to call the flannel plugin but to then delegate control to the Calico plugin. Using the flannel plugin ensures that the Calico plugin is called with the IP range for the host that was selected by flannel.
+
+* Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
+* Replace `${ETCD_ENDPOINTS}`
+
+**/etc/kubernetes/cni/net.d/10-calico.conf**
+
+```json
+{
+    "name": "calico",
+    "type": "flannel",
+    "delegate": {
+        "type": "calico",
+        "etcd_endpoints": "$ETCD_ENDPOINTS",
+        "log_level": "none",
+        "log_level_stderr": "info",
+        "hostname": "${ADVERTISE_IP}",
+        "policy": {
+            "type": "k8s",
+            "k8s_api_root": "http://127.0.0.1:8080/api/v1/"
+        }
+    }
+}
+```
+
 
 ## Start Services
 
@@ -339,7 +455,22 @@ $ sudo systemctl enable kubelet
 Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
 ```
 
-### Create kube-system Namespace
+### Start Calico
+
+Start Calico if you configured your kubelet to use CNI plugins earlier:
+
+```sh
+$ sudo systemctl start calico-node
+```
+
+Ensure that the Calico node container will start after a reboot:
+
+```sh
+$ sudo systemctl enable calico-node
+Created symlink from /etc/systemd/system/multi-user.target.wants/calico-node.service to /etc/systemd/system/calico-node.service.
+```
+
+### Create Namespaces
 
 The Kubernetes Pods that make up the control plane will exist in their own namespace. We need to create this namespace so these components are discoverable by other hosts in the cluster.
 
@@ -369,11 +500,37 @@ Now we can create the `kube-system` namespace:
 $ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"kube-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
 ```
 
+The Calico policy-agent runs in its own `calico-system` namespace. Create this namespace:
+
+```sh
+$ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"calico-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
+```
+
+### Enable network policy API
+
+Network policy in Kubernetes is currently implemented as a third party resource. To enable it, run the following:
+
+```sh
+$ curl -H "Content-Type: application/json" -XPOST http://127.0.0.1:8080/apis/extensions/v1beta1/namespaces/default/thirdpartyresources --data-binary @- <<BODY
+{
+  "kind": "ThirdPartyResource",
+  "apiVersion": "extensions/v1beta1",
+  "metadata": {
+    "name": "network-policy.net.alpha.kubernetes.io"
+  },
+  "description": "Specification for a network isolation policy",
+  "versions": [
+    {
+      "name": "v1alpha1"
+    }
+  ]
+}
+BODY
+```
+
 Our Pods should now be starting up and downloading their containers. To check the download progress, you can run `docker ps`.
 
 To check the health of the kubelet systemd unit that we created, run `systemctl status kubelet.service`.
-
-If you run into issues with Docker and flannel, check to see that the drop-in was applied correctly by running `systemctl cat docker.service` and ensuring that the drop-in appears at the bottom.
 
 <div class="co-m-docs-next-step">
   <p><strong>Did the containers start downloading?</strong> As long as they started to download, everything is working properly.</p>
--- a/Documentation/deploy-workers.md
+++ b/Documentation/deploy-workers.md
@@ -30,9 +30,9 @@ $ sudo ln -s ${WORKER_FQDN}-worker-key.pem worker-key.pem
 ```
 
 
-### flannel Configuration
+### Networking Configuration
 
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently of flannel, then the flannel parts of this guide can be skipped. It's recommended that Calico is still used for providing network policy. See [kubernetes networking](kubernetes-networking.md) for more detail.
 
 Just like earlier, create `/etc/flannel/options.env` and modify these values:
 
@@ -59,7 +59,7 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 
 ### Docker Configuration
 
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
 
 Require that flanneld is running prior to Docker start.
 
@@ -80,7 +80,8 @@ Create `/etc/systemd/system/kubelet.service` and substitute the following variab
 * Replace `${MASTER_HOST}`
 * Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
 * Replace `${DNS_SERVICE_IP}`
-* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release
+* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release. If using Calico a version that includes CNI binaries should be used. e.g. `v1.2.3_coreos.cni.1`
+* Replace `${NETWORK_PLUGIN}` with `cni` if using Calico. Otherwise just leave it blank.
 
 **/etc/systemd/system/kubelet.service**
 
@@ -91,6 +92,8 @@ ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
 Environment=KUBELET_VERSION=${K8S_VER}
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=https://${MASTER_HOST} \
+  --network-plugin-dir=/etc/kubernetes/cni/net.d \
+  --network-plugin=${NETWORK_PLUGIN} \
   --register-node=true \
   --allow-privileged=true \
   --config=/etc/kubernetes/manifests \
@@ -104,6 +107,38 @@ Restart=always
 RestartSec=10
 [Install]
 WantedBy=multi-user.target
+```
+
+### Set Up the CNI config
+
+The kubelet reads the CNI configuration on startup and uses that to determine which CNI plugin to call. Create the following file which tells the kubelet to call the flannel plugin but to then delegate control to the Calico plugin. Using the flannel plugin ensures that the Calico plugin is called with the IP range for the node that was selected by flannel.
+
+Note that this configuration is different to the one on the master nodes. It includes additional Kubernetes authentication information since the API server isn't available on localhost.
+
+* Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
+* Replace `${ETCD_ENDPOINTS}`
+* Replace `${CONTROLER_IP}`
+
+**/etc/kubernetes/cni/net.d/10-calico.conf**
+
+```json
+{
+    "name": "calico",
+    "type": "flannel",
+    "delegate": {
+        "type": "calico",
+        "etcd_endpoints": "$ETCD_ENDPOINTS",
+        "log_level": "none",
+        "log_level_stderr": "info",
+        "hostname": "${ADVERTISE_IP}",
+        "policy": {
+            "type": "k8s",
+            "k8s_api_root": "https://${CONTROLER_IP}:443/api/v1/",
+            "k8s_client_key": "/etc/kubernetes/ssl/worker-key.pem",
+            "k8s_client_certificate": "/etc/kubernetes/ssl/worker.pem"
+        }
+    }
+}
 ```
 
 ### Set Up the kube-proxy Pod
@@ -182,6 +217,48 @@ contexts:
 current-context: kubelet-context
 ```
 
+### Set Up Calico Node Container
+
+The Calico node container runs on all hosts, including the master node. It performs two functions:
+* Connects containers to the flannel overlay network, which enables the "one IP per pod" concept.
+* Enforces network policy created through the Kubernetes policy API, ensuring pods talk to authorized resources only.
+
+This step can be skipped if not using Calico.
+
+Create `/etc/systemd/system/calico-node.service` and substitute the following variables:
+
+* Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
+* Replace `${ETCD_ENDPOINTS}`
+
+**/etc/systemd/system/calico-node.service**
+
+```yaml
+[Unit]
+Description=Calico node for network policy
+Requires=network-online.target
+After=network-online.target
+
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --stage1-from-dir=stage1-fly.aci \
+--volume=modules,kind=host,source=/lib/modules,readOnly=false \
+--mount=volume=modules,target=/lib/modules \
+--trust-keys-from-https quay.io/calico/node:v0.19.0
+KillMode=mixed
+Restart=always
+TimeoutStartSec=0
+
+[Install]
+WantedBy=multi-user.target
+```
+
 ## Start Services
 
 Now we can start the Worker services.
@@ -194,24 +271,29 @@ Tell systemd to rescan the units on disk:
 $ sudo systemctl daemon-reload
 ```
 
-### Start kubelet
+### Start kubelet, flannel and Calico Node
 
-Start the kubelet, which will start the proxy as well.
+Start the kubelet, which will start the proxy as well as the Calico node (if required).
 
 ```sh
+$ sudo systemctl start flanneld
 $ sudo systemctl start kubelet
+$ sudo systemctl start calico-node
 ```
 
-Ensure that the kubelet starts on each boot:
+Ensure that the services start on each boot:
 
 ```sh
+$ sudo systemctl enable flanneld
+Created symlink from /etc/systemd/system/multi-user.target.wants/flanneld.service to /etc/systemd/system/flanneld.service.
 $ sudo systemctl enable kubelet
 Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
+$ sudo systemctl enable calico-node
+Created symlink from /etc/systemd/system/multi-user.target.wants/calico-node.service to /etc/systemd/system/calico-node.service.
 ```
 
 To check the health of the kubelet systemd unit that we created, run `systemctl status kubelet.service`.
-
-If you run into issues with Docker and Flannel, check to see that the drop-in was applied correctly by running `systemctl cat docker.service` and ensuring that the drop-in appears at the bottom.
+To check the health of the calico-node systemd unit that we created, run `systemctl status calico-node.service`.
 
 <div class="co-m-docs-next-step">
   <p><strong>Is the kubelet running?</strong></p>
--- a/Documentation/kubernetes-networking.md
+++ b/Documentation/kubernetes-networking.md
@@ -2,7 +2,7 @@
 
 ## Network Model
 
-The Kubernetes network model outlines four methods of component communication:
+The Kubernetes network model outlines three methods of component communication:
 
 * Pod-to-Pod Communication
     * Each Pod in a Kubernetes cluster is assigned an IP in a flat shared networking namespace. This allows for a clean network model where Pods, from a networking perspective, can be treated much like VMs or physical hosts.
@@ -13,13 +13,13 @@ The Kubernetes network model outlines four methods of component communication:
 * External-to-Internal Communication
     * Accessing services from outside the cluster is generally implemented by configuring external loadbalancers which target all nodes in the cluster. Once traffic arrives at a node, it is routed to the correct Service backends via the kube-proxy.
 
-See [Kubernetes Networking][kubernetes-network] for more detailed information on the Kubernetes network model and motiviation.
+See [Kubernetes Networking][kubernetes-network] for more detailed information on the Kubernetes network model and motivation.
 
 [kubernetes-network]: http://kubernetes.io/docs/admin/networking.html
 
 ## Port allocation
 
-The information below describes a minimum set of port allocations used by Kubernetes components. Some of these allocations will be optional depending on the deployment (e.g. if flannel is being used). Additionally, there are likely additional ports a deployer will need to open on their infrastructure (e.g. 22/ssh).
+The information below describes a minimum set of port allocations used by Kubernetes components. Some of these allocations will be optional depending on the deployment (e.g. if flannel or Calico is being used). Additionally, there are likely additional ports a deployer will need to open on their infrastructure (e.g. 22/ssh).
 
 Master Node Inbound
 
@@ -33,35 +33,38 @@ Worker Node Inbound
 -----------|-------------|--------------------------------|------------------------------------------------------------------------|
 | TCP      | 10250       | Master Nodes                   | Worker node Kubelet healthcheck port.                                  |
 | TCP      | 30000-32767 | External Application Consumers | Default port range for [external service][external-service] ports. Typically, these ports would need to be exposed to external load-balancers, or other external consumers of the application itself. |
-| TCP      | ALL         | Master & Worker Nodes          | Intra-cluster communication (unnecessary if flannel is used)           |
+| TCP      | ALL         | Master & Worker Nodes          | Intra-cluster communication (unnecessary if `vxlan` is used for networking)           |
 | UDP      | 8285        | Worker Nodes                   | flannel overlay network - *udp backend*. This is the default network configuration (only required if using flannel) |
 | UDP      | 8472        | Worker Nodes                   | flannel overlay network - *vxlan backend* (only required if using flannel) |
+| TCP      | 179         | Worker Nodes                   | Calico BGP network (only required if the BGP backend is used) |
 
 etcd Node Inbound
 
 | Protocol | Port Range | Source        | Purpose                                                  |
 -----------|------------|---------------|----------------------------------------------------------|
 | TCP      | 2379-2380  | Master Nodes  | etcd server client API                                   |
-| TCP      | 2379-2380  | Worker Nodes  | etcd server client API (only required if using flannel). |
+| TCP      | 2379-2380  | Worker Nodes  | etcd server client API (only required if using flannel or Calico). |
 
 [external-service]: http://kubernetes.io/docs/user-guide/services/#publishing-services---service-types
 
 ## Advanced Configuration
 
-The CoreOS Kubernetes documentation will describe either using cloud-provider networking integration (e.g. AWS, GCE), or a software-defined overlay network (e.g. [flannel][coreos-flannel]) to manage the Kubernetes Pod Network. However, in some cases a deployer may want to make use of existing network infrastructure to manage the Kubernetes network themselves.
+The CoreOS Kubernetes documentation describes a software-defined overlay network (i.e. [flannel][coreos-flannel]) to manage the Kubernetes Pod Network. However, in some cases a deployer may want to make use of existing network infrastructure to manage the Kubernetes network themselves e.g. using [Calico][calico]
 
 The following requirements must be met by your existing infrastructure to use Tectonic with a self-managed network.
 
 [coreos-flannel]: https://coreos.com/flannel/docs/latest/flannel-config.html
+[calico]: http://docs.projectcalico.org/en/latest/
 
 ### Pod-to-Pod Communication
 
 Each pod in the Kubernetes cluster will be assigned an IP that is expected to be routable from all other hosts and pods in the Kubernetes cluster.
 
-The general model for achieving this is to first assign an IP range to each host in your cluster.
+An easy way to achieve this is to use Calico. The Calico agent is already running on each node to enforce network policy. Starting it with the `CALICO_NETWORKING` environment variable set to `true` will cause it to run a BGP agent inside the Calico agent pod. These BGP agents will automatically form a full mesh network to exchange routing information. This allows a single large IP range to be used across your whole cluster and IP addresses to be efficiently assigned from it.  To peer with your existing BGP infrastructure follow this [guide][calico-bgp]. If your Kubernetes cluster is hosted on an [L2 network][calico-l2] (e.g. in your own datacenter or on AWS) there is no need to peer with your routers.
+
+An alternative way to achieve this is to first assign an IP range to each host in your cluster.
 Requests to IPs in an assigned range would need to be routed to that host via your network infrastructure.
 Next, the host is configured such that each pod launched on the host is assigned an IP from the host range.
-
 
 For example:
 
@@ -77,6 +80,8 @@ The actual allocation of Pod IPs on the host can be achieved by configuring Dock
 To achieve this network model, there are various methods that can be used. See the [Kubernetes Networking][how-to-achieve] documentation for more detail.
 
 [how-to-achieve]: http://kubernetes.io/docs/admin/networking.html#how-to-achieve-this
+[calico-bgp]: https://github.com/projectcalico/calico-containers/blob/v0.19.0/docs/bgp.md
+[calico-l2]: http://docs.projectcalico.org/en/latest/l2-interconnectFabric.html
 
 ### Pod-to-Service Communication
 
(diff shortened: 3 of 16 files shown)


Documentation/deploy-master.md
```diff
@@ -301,6 +287,69 @@ spec:
...
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: Can we make this multi-line (easier to read options).

aaronlevy: Instead of using the docker:// prefix, we should be able to pull from quay and have it to convert to an ACI on the fly. Then the `--insecure-options=image` flag can be dropped too.

Documentation/deploy-master.md
```diff
@@ -301,6 +287,69 @@ spec:
...
+
+### Set Up the policy-agent Pod
+
+The policy agent is the last major piece of the our master node. It monitors the API for changes related to network policy and configures Calico to implement that policy.
+
+When creating `/srv/kubernetes/manifests/policy-agent.yaml`:
+
+* Replace `${ETCD_ENDPOINTS}`
+
+**/srv/kubernetes/manifests/policy-agent.yaml**
```
aaronlevy: How do we ensure this continues to run (if this pod dies it will never be rescheduled)? Could be a static pod, or daemonset instead?

Documentation/deploy-master.md
```diff
@@ -369,12 +433,38 @@ Now we can create the `kube-system` namespace:
 $ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"kube-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
 ```
 
+The Calico policy-agent runs in it's own `calico-system` namespace:
+
+```sh
+$ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"kube-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
```
aaronlevy: s/kube-system/calico-system

Documentation/deploy-workers.md
```diff
@@ -102,10 +98,40 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
...
+        "type": "calico",
+        "etcd_endpoints": "$ETCD_ENDPOINTS",
+        "log_level": "none",
+        "log_level_stderr": "info",
+        "hostname": "${ADVERTISE_IP}",
+        "policy": {
+            "type": "k8s",
+            "k8s_api_root": "https://${CONTROLER_IP}:443/api/v1/",
+            "k8s_client_key": "/etc/kubernetes/ssl/worker-key.pem",
+            "k8s_client_certificate": "/etc/kubernetes/ssl/worker.pem"
```
aaronlevy: nit: assuming this needs to be configured on master nodes too, calling out "worker" here could be confusing.

Documentation/deploy-workers.md
```diff
@@ -182,6 +208,44 @@ contexts:
...
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: Same comment as above (nit: convert to multi line, and drop docker:// prefix).

Documentation/deploy-workers.md
```diff
@@ -194,24 +258,30 @@ Tell systemd to rescan the units on disk:
...
 Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
+$ sudo systemctl enable calico-agent
+Created symlink from /etc/systemd/system/multi-user.target.wants/calico-agent.service to /etc/systemd/system/calico-agent.service.
+
 ```
 
 To check the health of the kubelet systemd unit that we created, run `systemctl status kubelet.service`.
-
-If you run into issues with Docker and Flannel, check to see that the drop-in was applied correctly by running `systemctl cat docker.service` and ensuring that the drop-in appears at the bottom.
+To check the health of the calico-node systemd unit that we created, run `systemctl status calico-node.service`.
```
aaronlevy: This section intermixes use of `calico-node` and `calico-agent`. Not sure which is valid, but should unify naming.

Documentation/deploy-master.md
tomdee: Good point. I'll put it in /etc/kubernetes/manifests instead and update to the recently release policy agent that support multi master - config here https://github.com/projectcalico/k8s-policy/blob/master/examples/calico-policy-agent.yaml

Documentation/kubernetes-networking.md
```diff
@@ -58,11 +59,12 @@ The following requirements must be met by your existing infrastructure to use Te
 
 Each pod in the Kubernetes cluster will be assigned an IP that is expected to be routable from all other hosts and pods in the Kubernetes cluster.
 
-The general model for achieving this is to first assign an IP range to each host in your cluster.
+An easy way to achieve this is to use Calico. The Calico agent is already running on each node to enforce network policy. Starting it with the `CALICO_NETWORKING` environment variable set to `true` will cause it to run a BGP agent inside the Calico agent pod. These BGP agents will automatically form a full mesh network to exchange routing information. This allows a single large IP range to be used across your whole cluster and IP addresses to be efficiently assigned from it.  To peer with your existing BGP infrastructure follow this [guide][calico-bgp]. If you're Kubernetes clusters is hosted on an [L2 network](calico-l2) (e.g. in your own datacenter or on AWS) there is no need to peer with your routers.
```
aaronlevy: The calico-l2 link here should be `[L2 network][calico-l2]` .. (note the square brackets)

Documentation/kubernetes-upgrade.md
```diff
@@ -20,6 +20,18 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
...
+## Upgrading Calico
+
+The Calico agent runs on both master and worker nodes, and is is distributed as a container image. It runs under rkt using systemd.
+
+To update the image version, change the image tag in the service file (`/etc/systemd/system/calico-node.service`) to reference the new calico-node image.
+
+
+**/etc/systemd/system/calico-node.service**
+```
+ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: make multi-line

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -53,6 +47,32 @@ coreos:
...
+        [Service]
+        Slice=machine.slice
+        Environment=CALICO_DISABLE_FILE_LOGGING=true
+        Environment=HOSTNAME=$private_ipv4
+        Environment=IP=$private_ipv4
+        Environment=FELIX_FELIXHOSTNAME=$private_ipv4
+        Environment=CALICO_NETWORKING=false
+        Environment=NO_DEFAULT_POOLS=true
+        Environment=ETCD_ENDPOINTS={{ .ETCDEndpoints }}
+        ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: make multi-line & drop docker:// prefix & --insecure-options

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -100,6 +121,14 @@ write_files:
       -d @"/srv/kubernetes/manifests/heapster-dc.json" \
       "http://127.0.0.1:8080/apis/extensions/v1beta1/namespaces/kube-system/deployments"
 
+      /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
+      -d @"/srv/kubernetes/manifests/calico-policy-agent.json" \
+      "http://127.0.0.1:8080/api/v1/namespaces/kube-system/replicationcontrollers"
```
aaronlevy: Looks like `calico-policy-agent.json` is being created in `/etc/kubernetes/manifests`. Also, the manifest is a Pod, and this api endpoint is for replication controllers.

aaronlevy: There is also a namespace mismatch (kube-system vs calico-system)

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -42,6 +48,32 @@ coreos:
...
+        [Service]
+        Slice=machine.slice
+        Environment=CALICO_DISABLE_FILE_LOGGING=true
+        Environment=HOSTNAME=$private_ipv4
+        Environment=IP=$private_ipv4
+        Environment=FELIX_FELIXHOSTNAME=$private_ipv4
+        Environment=CALICO_NETWORKING=false
+        Environment=NO_DEFAULT_POOLS=true
+        Environment=ETCD_ENDPOINTS={{ .ETCDEndpoints }}
+        ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: make multi-line & drop docker:// prefix & --insecure-options

multi-node/generic/controller-install.sh
```diff
@@ -100,6 +100,37 @@ WantedBy=multi-user.target
...
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: make multi-line & drop docker:// prefix & --insecure-options

multi-node/generic/worker-install.sh
```diff
@@ -69,6 +79,35 @@ WantedBy=multi-user.target
...
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: make multi-line & drop docker:// prefix & --insecure-options

single-node/user-data
```diff
@@ -93,6 +103,35 @@ WantedBy=multi-user.target
...
+[Service]
+Slice=machine.slice
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --volume=modules,kind=host,source=/lib/modules,readOnly=false --mount=volume=modules,target=/lib/modules --stage1-from-dir=stage1-fly.aci --insecure-options=image docker://quay.io/calico/node:v0.19.0
```
aaronlevy: nit: make multi-line & drop docker:// prefix & --insecure-options

Documentation/deploy-workers.md
```diff
@@ -86,11 +72,21 @@ Create `/etc/systemd/system/kubelet.service` and substitute the following variab
...
+Environment="RKT_OPTS=--volume cni,kind=host,source=/opt/cni/bin --mount volume=cni,target=/opt/cni/bin"
+ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
+ExecStartPre=/usr/bin/mkdir -p /opt/cni/bin
+ExecStartPre=/usr/bin/chmod a+w /opt/cni/bin
+ExecStartPre=-/usr/bin/wget -nc -O /opt/cni/bin/calico https://github.com/projectcalico/calico-cni/releases/download/v1.3.0/calico
+ExecStartPre=-/usr/bin/wget -nc -O /opt/cni/bin/flannel https://f001.backblaze.com/file/calico/flannel
+ExecStartPre=-/usr/bin/wget -nc -O /opt/cni/bin/host-local https://f001.backblaze.com/file/calico/host-local
+ExecStartPre=/usr/bin/chmod +x /opt/cni/bin/calico
+ExecStartPre=/usr/bin/chmod +x /opt/cni/bin/flannel
+ExecStartPre=/usr/bin/chmod +x /opt/cni/bin/host-local
```
aaronlevy: You mention this in the PR, but wget'ing a bunch of binaries to the host isn't an ideal method for distributing these assets. Might need to think about this a bit more, but options might be including blessed versions/assets in the hyperkube image. Or at least distributing these binaries via utility containers. Or? 

/cc @philips

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -16,15 +16,9 @@ coreos:
     - name: etcd2.service
       command: start
 
-    - name: docker.service
```
aaronlevy: Can you describe why this is no longer necessary?

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -144,3 +176,23 @@ write_files:
           name: kubelet-context
         current-context: kubelet-context
 
+  - path: /etc/kubernetes/cni/net.d/10-calico.conf
```
aaronlevy: It doesn't look like this is being created on the master nodes. Is it not necessary for some reason?

Documentation/deploy-master.md
```diff
@@ -60,26 +60,11 @@ Next create a [systemd drop-in][dropins], which is a method for appending or ove
 ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 ```
 
+[calico-docs]: https://github.com/projectcalico/calico-containers/tree/master/docs/cni/kubernetes
+[network-policy]: https://github.com/caseydavenport/kubernetes/blob/network-policy/docs/admin/network-policy.md#network-policy-in-kubernetes
```
aaronlevy: How far off is this from being merged? I don't think we should mention this in docs if people can't use it in the targeted version.

aaronlevy:
@tomdee Done with first review pass. Also, this will need to pass the upstream conformance tests prior to being merged, so might be worthwhile just running that as a sanity check. Let me know if we can be of help there.

/cc @joshix @robszumski for docs changes.

tomdee:
Thanks @aaronlevy for the speedy review. I'll get all the easy stuff fixed tomorrow and pull out a list of things that might need more discussion.

Documentation/deploy-master.md
tomdee: If I run this from the command line (on Ubuntu) it asks me for confirmation:
```
image: remote fetching from URL "https://quay.io/c1/aci/quay.io/calico/node/v0.19.0/aci/linux/amd64/"
pubkey: prefix: "quay.io/calico/node"
key: "https://quay.io/aci-signing-key"
gpg key fingerprint is: BFF3 13CD AA56 0B16 A898  7B8F 72AB F5F6 799D 33BC
	Quay.io ACI Converter (ACI conversion signing key) <support@quay.io>
Are you sure you want to trust this key (yes/no)?
```
How do I skip that when it's scripted? Or does CoreOS automatically trust quay.io?

aaronlevy: Woops, you'll also need to add the rkt flag `--trust-keys-from-https`

multi-node/aws/pkg/config/templates/cloud-config-controller
tomdee: Previously, Kubernetes used docker bridge networking for containers. This worked because flannel  obtains a subnet for the host and configures the docker bridge to use that subnet. Then docker will assign IPs from that range.

Now that Kubernetes is configured to use CNI networking, it doesn't use Docker for networking. Docker creates the containers with `--net=none` then K8s invokes the CNI plugin to do the networking. 

Now the flannel CNI plugin is used which calls through to the Calico CNI plugin to do the actual container networking. The flannel plugin tells Calico that it should call through to the host-local IPAM plugin for IP assignment, and flannels also configures the IP range that is passed to the host-local plugin.

Hence, docker no longer has a dependency on Flannel.

Documentation/deploy-workers.md
tomdee: This should now be fixed. The correct term is `calico-node`

Documentation/deploy-master.md
```diff
@@ -154,6 +138,7 @@ spec:
     - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
     - --client-ca-file=/etc/kubernetes/ssl/ca.pem
     - --service-account-key-file=/etc/kubernetes/ssl/apiserver-key.pem
+    - --runtime-config=extensions/v1beta1/deployments=true,extensions/v1beta1/daemonsets=true,extensions/v1beta1=true,extensions/v1beta1/thirdpartyresources=true
```
caseydavenport: I think that `v1beta/deployments` and `v1beta/daemonsets` are enabled by default in v1.2

https://github.com/kubernetes/kubernetes/blob/release-1.2/pkg/master/master.go#L667

Documentation/deploy-master.md
```diff
@@ -33,11 +33,12 @@ $ sudo chmod 600 /etc/kubernetes/ssl/*-key.pem
...
-### flannel Configuration
+### Network Configuration
+Networking is provided by Flannel and Calico.
+* [flannel][flannel-docs] provides a software-defined overlay network for routing traffic to/from the [pods][pod-overview]
+* [Calico][calico-docs] secures the overlay network by restricting traffic to/from the [Pod]s based on fine-grained network policy.
 
-[flannel][flannel-docs] provides a key Kubernetes networking capability &mdash; a software-defined overlay network to manage routing of the [Pod][pod-overview] network.
-
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently of flannel, then the flannel parts of this guide can be skipped. It's recommended that Calico is still used for providing network policy. See [kubernetes networking](kubernetes-networking.md) for more detail.
```
joshix: replace second sentence: "In this case, Calico may still be used for network policy."
Capitalize "Kubernetes"

Documentation/deploy-master.md
```diff
@@ -301,6 +286,86 @@ spec:
       timeoutSeconds: 1
 ```
 
+### Run the per host calico agent
+The per host Calico agent runs on all hosts including the master node. It is responsible for enforcing networking policy and connecting the pods on that host to the flannel overlay network.
```
joshix: s/per host// (redundant)
s/hosts/hosts,/
s/that/each/

Documentation/deploy-master.md
```diff
@@ -301,6 +286,86 @@ spec:
...
+Restart=always
+TimeoutStartSec=0
+
+[Install]
+WantedBy=multi-user.target
+```
+
+### Set Up the policy-agent Pod
+
+The policy agent is the last major piece of the our master node. It monitors the API for changes related to network policy and configures Calico to implement that policy.
```
joshix: s/our//

Documentation/deploy-master.md
```diff
@@ -301,6 +286,86 @@ spec:
...
+
+When creating `/etc/kubernetes/manifests/policy-agent.yaml`:
+
+* Replace `${ETCD_ENDPOINTS}`
```
joshix: "... with the same endpoints used in `calico-node.service`, above." (Correct?)

Documentation/deploy-master.md
```diff
@@ -369,12 +449,38 @@ Now we can create the `kube-system` namespace:
 $ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"kube-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
 ```
 
+The Calico policy-agent runs in it's own `calico-system` namespace:
```
joshix: s/it's/its/

multi-node/aws/pkg/config/config.go
```diff
@@ -126,6 +127,16 @@ func (c Cluster) Config() (*Config, error) {
 	config.APIServers = fmt.Sprintf("http://%s:8080", c.ControllerIP)
 	config.SecureAPIServers = fmt.Sprintf("https://%s:443", c.ControllerIP)
 	config.APIServerEndpoint = fmt.Sprintf("https://%s", c.ExternalDNSName)
+	if config.UseCalico {
+		config.K8sNetworkPlugin = "cni"
+		config.CalicoEnable = "true"
+		// Use a Hyperkube with CNI binaries
+		config.K8sVer = config.K8sVer + "-cni"
+	}else{
+		config.CalicoEnable = "false"
```
colhom: nit: `config.CalicoEnable` seems redundant w.r.t `cluster.UseCalico`. Is there a desire to decouple the two states in the future?

Documentation/deploy-master.md
tomdee: @caseydavenport Can you raise this as a separate issue on this repo?

Documentation/deploy-master.md
```diff
@@ -301,6 +286,86 @@ spec:
...
+**/etc/kubernetes/manifests/policy-agent.yaml**
+
+```yaml
+apiVersion: v1
+kind: Pod 
+metadata:
+  name: calico-policy-agent
+  namespace: calico-system 
+spec:
+  hostNetwork: true
```
colhom: Why does the agent need to be a static pod on the controller with host networking? It looks to me like this pod should be scheduled via a deployment, and we don't need the host networking.

* `ETCD_ENDPOINTS` is routable from the pod network on any kubelet in the cluster. 

* `policy-agent` pod talk to the "in-cluster" kubernetes service at `kubernetes.default.svc.cluster.local` instead of using the insecure port and host networking. This is the standard way that cluster components communicate with the api server.

multi-node/aws/pkg/config/config.go
tomdee: I was just figuring that CalicoEnable is a yaml string that gets transformed into a Go boolean. I need to have a template value that I can use unchanged as a cloud-config enable flag. I wasn't sure if a good boolean would serialize into the right format (with the right caps) for a cloud-config flag so thought it better to be explicit. I'm happy to test further if you think it's worth it?

colhom: It'll work! Golang boolean serialization is cloud-config compatible!

https://play.golang.org/p/xefEIUcA1M

Documentation/deploy-master.md
aaronlevy: I agree with @chom on this.

Couple clarifying points:

The ETCD_ENDPOINTS aren't on the pod network itself (unless this has changed elsewhere), but they are routable from any node in the cluster.

I'm not sure if the policy-agent knows how to securely speak with the api-server (but ideally it should). The two main options developers generally provide are getting the necessary data from a kubeconfig file, or using "InClusterConfig" (discovers api-server from env var in the pod and service account data mounted at known location).

Documentation/deploy-master.md
```diff
@@ -301,6 +286,86 @@ spec:
...
+          value: "true"
+    # Leader election container used by the policy agent.
+    - name: leader-elector
+      image: gcr.io/google_containers/leader-elector:0.4
+      imagePullPolicy: IfNotPresent
+      args:
+        - "--election=calico-policy-election"
+        - "--http=127.0.0.1:4040"
+      ports:
+        - containerPort: 4040
```
aaronlevy: Does anything external to this pod need this? All containers in the pod should be in same network namespace, so can already communicate over localhost. If this stays + hostNetworking=true means that 4040 is bound on the host too

Documentation/deploy-master.md
colhom: Nope, the etcd endpoint is literally a port on the controller instance. My point was that (due to lack of network policy!!) you can hit that endpoint from any pod.

I've been reading through the source for policy agent- it seems that the default behavior will use service token secret mounted in at the conventional location and the in-cluster kubernetes service. 

[defaults for ca.crt and kuberentes service endpoint](https://github.com/projectcalico/k8s-policy/blob/c62fe7e113b013141eca808980907c0af7b0773f/constants.py#L4)
[default for service token](https://github.com/projectcalico/k8s-policy/blob/a8e439792e5f993dc910561c0ac5f4604b22960c/policy_agent.py#L437)
[how the defaults get applied](https://github.com/projectcalico/k8s-policy/blob/a8e439792e5f993dc910561c0ac5f4604b22960c/policy_agent.py#L35)

tldr; remove `K8S_API` env var and host-networking flag, and the policy-agent pod should automatically use the in-cluster k8s service correctly.

Documentation/deploy-master.md
colhom: 4040 is only used by `policy-agent` container to determine if that pod is the leader. Doesn't need to be exposed at the pod level.

Documentation/deploy-master.md
caseydavenport: Yeah, the policy agent code does support that - it comes from a time when it was run on-cluster as you describe (and is maintained _i think_ for backwards compat with older CNI plugin versions, though it should be removed soon).

That approach won't work any more because of how we've tweaked the responsibility of the policy agent / CNI plugin in terms of configuring network policy.  The policy agent, if run as a pod, won't have the necessary network policy in place to allow it to talk to the API.  It's the responsibility of the agent to configure the network policy that would allow it to reach the API - a bootstrap loop.  @aaronlevy and I were just discussing this.

Documentation/deploy-master.md
tomdee: Scratch that. Now that I've rebased I see they've removed the line in master

Documentation/deploy-workers.md
tomdee: It's now on the master too. It's using worker.pem but only on the worker nodes. LMK if you still think this is problem.

multi-node/aws/pkg/config/templates/cloud-config-worker
tomdee: It is now

multi-node/aws/pkg/config/config.go
tomdee: Cool, I'll make that change.

tomdee:
ping @aaronlevy @colhom @joshix I've addressed all the feedback, what are the next steps to getting this merged?

Documentation/deploy-workers.md
```diff
@@ -194,24 +270,30 @@ Tell systemd to rescan the units on disk:
 $ sudo systemctl daemon-reload
 ```
 
-### Start kubelet
+### Start kubelet, flannel and Calico Node
 
-Start the kubelet, which will start the proxy as well.
+Start the kubelet, which will start the proxy as well and start the Calico node (if required).
```
caseydavenport: "which will start the proxy as well as the Calico node container (if required)"

Documentation/kubernetes-on-aws.md
```diff
@@ -158,6 +158,17 @@ You can now customize your cluster by editing asset files:
 
 You can also now check the `my-cluster` asset directory into version control if you desire. The contents of this directory are your reproducible cluster assets. Please take care not to commit the `my-cluster/credentials` directory, as it contains your TLS secrets. If you're using git, the `credentials` directory will already be ignored for you.
 
+#### Calico network policy (optional)
+
+THe cluster can be configured to use Calico to provide network policy.
```
caseydavenport: Typo: THe

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -264,6 +323,34 @@ write_files:
...
+              - name: K8S_API
+                value: "http://127.0.0.1:8080"
+              - name: LEADER_ELECTION
+                value: "true"
+          # Leader election container used by the policy agent.
+          - name: leader-elector
+            image: quay.io/calico/leader-elector:v0.1.0
+            imagePullPolicy: IfNotPresent
+            args:
+              - "--election=calico-policy-election"
```
caseydavenport: We probably want to pass the `--election-namespace=calico-system` option here.

tomdee: Done

Documentation/deploy-master.md
```diff
@@ -60,6 +60,7 @@ Next create a [systemd drop-in][dropins], which is a method for appending or ove
 ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 ```
 
+[calico-docs]: https://github.com/projectcalico/calico-containers/tree/master/docs/cni/kubernetes
```
aaronlevy: If we can, we should target a particular release for the docs.

Documentation/deploy-master.md
```diff
@@ -68,7 +69,7 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 
 In order for flannel to manage the pod network in the cluster, Docker needs to be configured to use it. All we need to do is require that flanneld is running prior to Docker starting.
 
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently of flannel, or if Calico is being used, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
```
aaronlevy: Saying if Calico is being used this step can be skipped might cause confusion because we *are* using Calico even when we are using flannel...

aaronlevy: Maybe we just don't call out either tool and leave this to:

> If the pod-network is being managed independently, this step can be skipped.

If the admin is deploying their own pod-network (can be quite complex), this should be sufficiently clear. (/cc @joshix )

Documentation/deploy-workers.md
```diff
@@ -59,7 +59,7 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 
 ### Docker Configuration
 
-*Note:* If the pod-network is being managed independently of flannel, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
+*Note:* If the pod-network is being managed independently of flannel, or if Calico is being used, this step can be skipped. See [kubernetes networking](kubernetes-networking.md) for more detail.
```
aaronlevy: Same comment as above in the master deploy

Documentation/deploy-master.md
```diff
@@ -88,11 +89,14 @@ The [kubelet](http://kubernetes.io/docs/admin/kubelet.html) is the agent on each
...
 
+When using Calico, the kubelet is configured to use CNI for networking. This makes Calico aware of each pod that is created and allows it to network the pods into the flannel overlay. The Calico CNI plugin is actually called by the Flannel CNI plugin so that the correct IP range that flannel has selected for the host can be used.
+
 Note that the kubelet running on a master node may log repeated attempts to post its status to the API server. These warnings are expected behavior and can be ignored. Future Kubernetes releases plan to [handle this common deployment consideration more gracefully](https://github.com/kubernetes/kubernetes/issues/14140#issuecomment-142126864).
 
 * Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
 * Replace `${DNS_SERVICE_IP}`
-* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release
+* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release. If using Calico a version that includes CNI binaries should be used. e.g. `v1.2.2_coreos_cni.0`
```
aaronlevy: Can we change this example to v1.2.3? People should never try to use v1.2.2 due to the controller-manager + thirdpartyresource issue (https://github.com/kubernetes/kubernetes/issues/24380)

Documentation/deploy-workers.md
```diff
@@ -80,7 +80,8 @@ Create `/etc/systemd/system/kubelet.service` and substitute the following variab
 * Replace `${MASTER_HOST}`
 * Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
 * Replace `${DNS_SERVICE_IP}`
-* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release
+* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release. If using Calico a version that includes CNI binaries should be used. e.g. `v1.2.2_coreos_cni.0`
```
aaronlevy: s/v1.2.2/v1.2.3/

Documentation/kubernetes-networking.md
```diff
@@ -77,6 +79,8 @@ The actual allocation of Pod IPs on the host can be achieved by configuring Dock
 To achieve this network model, there are various methods that can be used. See the [Kubernetes Networking][how-to-achieve] documentation for more detail.
 
 [how-to-achieve]: http://kubernetes.io/docs/admin/networking.html#how-to-achieve-this
+[calico-bgp]: https://github.com/projectcalico/calico-containers/blob/master/docs/bgp.md
```
aaronlevy: Target a specific release if possible

Documentation/kubernetes-networking.md
```diff
@@ -86,9 +90,10 @@ A requirement of a manually configured network is that the service-ip-range does
...
+IP addresses assigned on the pod network are typically not routable outside of the cluster unless you're using Calico and have [peered with your routers][calico-external]. This isn't an issue since most communication between your applications stays within the cluster, as described above. Allowing external traffic into the cluster is generally accomplished by mapping external load-balancers to specifically exposed services in the cluster. This mapping allows the kube-proxy process to route the external requests to the proper pods using the cluster's pod-network.
 
 In a manually configured network, it may be necessary to open a range of ports to outside clients (default 30000-32767) for use with "external services". See the [Kubernetes Service][kube-service] documentation for more information on external services.
 
+[calico-external]: https://github.com/projectcalico/calico-containers/blob/master/docs/ExternalConnectivity.md
```
aaronlevy: nit: target a specific release if possible

multi-node/generic/controller-install.sh
```diff
@@ -263,6 +311,40 @@ spec:
...
+        - name: K8S_API
+          value: "http://127.0.0.1:8080"
+        - name: LEADER_ELECTION
+          value: "true"
+    # Leader election container used by the policy agent.
+    - name: leader-elector
+      image: quay.io/calico/leader-elector:v0.1.0
+      imagePullPolicy: IfNotPresent
+      args:
+        - "--election=calico-policy-election"
```
aaronlevy: It was noted elsewhere that `--election-namespace=calico-system` should be added

single-node/user-data
```diff
@@ -254,6 +293,40 @@ spec:
...
+        - name: K8S_API
+          value: "http://127.0.0.1:8080"
+        - name: LEADER_ELECTION
+          value: "true"
+    # Leader election container used by the policy agent.
+    - name: leader-elector
+      image: quay.io/calico/leader-elector:v0.1.0
+      imagePullPolicy: IfNotPresent
+      args:
+        - "--election=calico-policy-election"
```
aaronlevy: Noted elsewhere that `--election-namespace=calico-system` should be added

Documentation/deploy-master.md
```diff
@@ -301,6 +308,114 @@ spec:
...
+        - name: K8S_API
+          value: "http://127.0.0.1:8080"
+        - name: LEADER_ELECTION 
+          value: "true"
+    # Leader election container used by the policy agent.
+    - name: leader-elector
+      image: quay.io/calico/leader-elector:v0.1.0
+      imagePullPolicy: IfNotPresent
+      args:
+        - "--election=calico-policy-election"
```
aaronlevy: noted elsewhere that `--election-namespace=calico-system` should be added

multi-node/generic/controller-install.sh
```diff
@@ -5,7 +5,10 @@ set -e
 export ETCD_ENDPOINTS=
 
 # Specify the version (vX.Y.Z) of Kubernetes assets to deploy
-export K8S_VER=v1.2.2_coreos.0
+export K8S_VER=v1.2.3_coreos.0-cni
```
aaronlevy: NOTE: we need to update this prior to merge

multi-node/generic/controller-install.sh
```diff
@@ -27,13 +30,16 @@ export K8S_SERVICE_IP=10.3.0.1
 # This same IP must be configured on all worker nodes to enable DNS service discovery.
 export DNS_SERVICE_IP=10.3.0.10
 
+# Whether to use Calico for Kubernetes network policy.
+export USE_CALICO=true
```
aaronlevy: In the generic scripts we should default to `USE_CALICO=false` for now, as people use these for deployments to cloud-providers not otherwise supported by this tool. I'm fine with the vagrant single/multinode installs flipping this to true, but this file should default to off so behavior doesn't change for existing users.

multi-node/generic/controller-install.sh
```diff
@@ -5,7 +5,10 @@ set -e
 export ETCD_ENDPOINTS=
 
 # Specify the version (vX.Y.Z) of Kubernetes assets to deploy
-export K8S_VER=v1.2.2_coreos.0
+export K8S_VER=v1.2.3_coreos.0-cni
+
+# Hyperkube image repository to use.
+export HYPERKUBE_IMAGE_REPO=quay.io/calico/hyperkube
```
aaronlevy: NOTE: update this prior to merge

aaronlevy: also, naming nit: just "HYPERKUBE_IMAGE"?

multi-node/generic/controller-install.sh
```diff
@@ -49,6 +55,12 @@ function init_config {
             exit 1
         fi
     done
+
+    if [ $USE_CALICO = "true" ]; then
+        export K8S_NETWORK_PLUGIN="cni"
```
aaronlevy: We should also set the HYPERKUBE_IMAGE here if USE_CALICO=true

multi-node/generic/controller-install.sh
```diff
@@ -100,6 +114,40 @@ WantedBy=multi-user.target
 EOF
     }
 
+
+    local TEMPLATE=/etc/systemd/system/calico-node.service
+    [ -f $TEMPLATE ] || {
```
aaronlevy: Shouldn't this be gated on the USE_CALICO flag?

multi-node/generic/controller-install.sh
```diff
@@ -644,14 +762,26 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 EOF
     }
 
-    local TEMPLATE=/etc/systemd/system/docker.service.d/40-flannel.conf
+    local TEMPLATE=/etc/kubernetes/cni/net.d/10-calico.conf
     [ -f $TEMPLATE ] || {
         echo "TEMPLATE: $TEMPLATE"
         mkdir -p $(dirname $TEMPLATE)
         cat << EOF > $TEMPLATE
-[Unit]
```
aaronlevy: This still needs to be gated on USE_CALICO=true

multi-node/generic/worker-install.sh
```diff
@@ -10,19 +10,25 @@ export ETCD_ENDPOINTS=
 export CONTROLLER_ENDPOINT=
 
 # Specify the version (vX.Y.Z) of Kubernetes assets to deploy
-export K8S_VER=v1.2.2_coreos.0
+export K8S_VER=v1.2.3_coreos.0-cni
```
aaronlevy: NOTE: update before merge

multi-node/generic/worker-install.sh
```diff
@@ -10,19 +10,25 @@ export ETCD_ENDPOINTS=
...
+
+# Hyperkube image repository to use.
+export HYPERKUBE_IMAGE_REPO=quay.io/calico/hyperkube
```
aaronlevy: NOTE: update before merge

multi-node/generic/worker-install.sh
```diff
@@ -38,6 +44,12 @@ function init_config {
             exit 1
         fi
     done
+
+    if [ $USE_CALICO = "true" ]; then
+        export K8S_NETWORK_PLUGIN="cni"
```
aaronlevy: Set HYPERKUBE_IMAGE here too

multi-node/generic/worker-install.sh
```diff
@@ -10,19 +10,25 @@ export ETCD_ENDPOINTS=
...
+
+# Hyperkube image repository to use.
+export HYPERKUBE_IMAGE_REPO=quay.io/calico/hyperkube
 
 # The IP address of the cluster DNS service.
 # This must be the same DNS_SERVICE_IP used when configuring the controller nodes.
 export DNS_SERVICE_IP=10.3.0.10
 
+# Whether to use Calico for Kubernetes network policy.
+export USE_CALICO=true
```
aaronlevy: Should default to false, then can set to "true" in the actual vagrant install.

multi-node/generic/worker-install.sh
```diff
@@ -159,14 +205,28 @@ ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 EOF
     }
 
-    local TEMPLATE=/etc/systemd/system/docker.service.d/40-flannel.conf
+    local TEMPLATE=/etc/kubernetes/cni/net.d/10-calico.conf
     [ -f $TEMPLATE ] || {
         echo "TEMPLATE: $TEMPLATE"
         mkdir -p $(dirname $TEMPLATE)
         cat << EOF > $TEMPLATE
-[Unit]
```
aaronlevy: Needs to be gated on USE_CALICO=true

single-node/Vagrantfile
```diff
@@ -21,9 +21,9 @@ Vagrant.configure("2") do |config|
   # always use Vagrant's insecure key
   config.ssh.insert_key = false
 
-  config.vm.box = "coreos-alpha"
+  config.vm.box = "coreos-beta"
```
aaronlevy: I'm assuming this change was due to: https://github.com/coreos/bugs/issues/1246 - but now that it is resolved, can we switch this back to alpha?

I'm not necessarily opposed to changing this, just checking motivation.

single-node/user-data
```diff
@@ -5,7 +5,10 @@ set -e
 export ETCD_ENDPOINTS="http://127.0.0.1:2379"
 
 # Specify the version (vX.Y.Z) of Kubernetes assets to deploy
-export K8S_VER=v1.2.2_coreos.0
+export K8S_VER=v1.2.3_coreos.0-cni
```
aaronlevy: NOTE: update before merge

single-node/user-data
```diff
@@ -5,7 +5,10 @@ set -e
...
+
+# Hyperkube image repository to use.
+export HYPERKUBE_IMAGE_REPO=quay.io/calico/hyperkube
```
aaronlevy: NOTE: update before merge

aaronlevy: I also would still like enabling the calico image/services to be a feature flag given we are using custom images. In the single-node vagrant - I'm not opposed to having it default to "true", but given that it didn't seem super complex from the multi-node install to gate on a flag - it would be helpful during debugging to be able to flip this on/off.

aaronlevy:
@tomdee finished with this review pass

Documentation/deploy-master.md
robszumski: Yes, this would be great

Documentation/deploy-master.md
robszumski: SGTM

Documentation/deploy-master.md
```diff
@@ -88,11 +89,14 @@ The [kubelet](http://kubernetes.io/docs/admin/kubelet.html) is the agent on each
 
 On the master node, the kubelet is configured to communicate with the API server, but not register for cluster work, as shown in the `--register-schedulable=false` line in the YAML excerpt below. This prevents user pods being scheduled on the master nodes, and ensures cluster work is routed only to task-specific worker nodes.
 
+When using Calico, the kubelet is configured to use CNI for networking. This makes Calico aware of each pod that is created and allows it to network the pods into the flannel overlay. The Calico CNI plugin is actually called by the Flannel CNI plugin so that the correct IP range that flannel has selected for the host can be used.
```
robszumski: CNI is never defined in this doc.

"When using Calico, the kubelet is configured to use the Container Networking Interface (CNI) standard for networking."

robszumski: > The Calico CNI plugin is actually called by the Flannel CNI plugin so that the correct IP range that flannel has selected for the host can be used.

This was a little confusing to me. What if we changed it to something similar to:

"Both flannel and Calico communicate via CNI interfaces to ensure the correct IP range (managed by flannel) is used for each node."

- I swapped host for node since that's used elsewhere in this doc (and others).
- flannel is always lowercase. We had mixed case in this sentence before.

Documentation/deploy-master.md
robszumski: Is there a way we can always build in the CNI binaries? And then enable them for Calico users? Or make Calico the default? This choice is going to be confusing for users.

Documentation/deploy-master.md
```diff
@@ -154,6 +160,7 @@ spec:
     - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
     - --client-ca-file=/etc/kubernetes/ssl/ca.pem
     - --service-account-key-file=/etc/kubernetes/ssl/apiserver-key.pem
+    - --runtime-config=extensions/v1beta1=true,extensions/v1beta1/thirdpartyresources=true
```
robszumski: Unsure if we should explain _why_ we're enabling these beta features? What does everyone think?

Documentation/deploy-master.md
```diff
@@ -301,6 +308,114 @@ spec:
       timeoutSeconds: 1
 ```
 
+### Set Up Calico Node Container
+
+The Calico node container runs on all hosts, including the master node. It performs two functions
+* It networks containers created through CNI to connect them to the flannel overlay network.
+* It enforces network policy created through the Kubernetes policy API.
```
robszumski: > It networks containers created through CNI to connect them to the flannel overlay network.

Would be nice to mention what each of these get you:

"Connects containers to the flannel overlay network, which enables the "one IP per pod" concept."

"Enforces network policy created through the Kubernetes policy API, ensuring pods talk to authorized resources only."

Documentation/deploy-master.md
```diff
@@ -301,6 +308,114 @@ spec:
       timeoutSeconds: 1
 ```
 
+### Set Up Calico Node Container
+
+The Calico node container runs on all hosts, including the master node. It performs two functions
```
robszumski: colon after "It performs two functions"

Documentation/deploy-master.md
```diff
@@ -339,7 +454,22 @@ $ sudo systemctl enable kubelet
 Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
 ```
 
-### Create kube-system Namespace
+### Start Calico
+
+Now start Calico (if required)
```
robszumski: can we give you a hint about the test to know if its required or not?

"Start Calico if you configured your kubelet to use CNI plugins earlier:"

Documentation/deploy-master.md
```diff
@@ -369,12 +499,38 @@ Now we can create the `kube-system` namespace:
 $ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"kube-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
 ```
 
+The Calico policy-agent runs in its own `calico-system` namespace:
```
robszumski: > The Calico policy-agent runs in its own `calico-system` namespace:

"The Calico policy-agent runs in its own `calico-system` namespace. **Create this namespace:**"

Documentation/deploy-master.md
```diff
@@ -369,12 +499,38 @@ Now we can create the `kube-system` namespace:
...
+
+```sh
+$ curl -H "Content-Type: application/json" -XPOST -d'{"apiVersion":"v1","kind":"Namespace","metadata":{"name":"calico-system"}}' "http://127.0.0.1:8080/api/v1/namespaces"
+```
+
+### Enable network policy API
+
+Network policy in Kubernetes is currently implemented as a third party resource. To enable it, run the following
```
robszumski: colon after "run the following"

Documentation/deploy-workers.md
```diff
@@ -106,6 +109,37 @@ RestartSec=10
 WantedBy=multi-user.target
 ```
 
+### Set Up the CNI config
+The kubelet reads the CNI configuration on startup and uses that to determine which CNI plugin to call. Create the following file which tells the kubelet to call the flannel plugin but to then delegate control to the Calico plugin. Using the flannel plugin ensures that the Calico plugin is called with the IP range for the host that was selected by flannel.
```
robszumski: host -> node

robszumski: new line before this paragraph

Documentation/deploy-workers.md
```diff
@@ -182,6 +216,48 @@ contexts:
 current-context: kubelet-context
 ```
 
+### Set Up Calico Node Container
+
+The Calico node container runs on each worker node and performs two functions
```
robszumski: echo earlier comments from the master doc

Documentation/deploy-workers.md
```diff
@@ -194,24 +270,30 @@ Tell systemd to rescan the units on disk:
...
+Ensure that the services start on each boot:
 
 ```sh
+$ sudo systemctl enable flanneld
+Created symlink from /etc/systemd/system/multi-user.target.wants/flanneld.service to /etc/systemd/system/flanneld.service.
 $ sudo systemctl enable kubelet
 Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
+$ sudo systemctl enable calico-node
+Created symlink from /etc/systemd/system/multi-user.target.wants/calico-node.service to /etc/systemd/system/calico-node.service.
+
```
robszumski: extra newline here

single-node/user-data
robszumski: @aaronlevy how would this feature flag work? I'd like to remove the choice between hyperkubes with CNI vs those without. Would this allow us to ship one container image?

multi-node/generic/controller-install.sh
tomdee: I'm sticking with the _REPO suffix (unless you feel strongly about it) as that's what's already used for AWS.

multi-node/generic/controller-install.sh
tomdee: To what though?

multi-node/generic/controller-install.sh
tomdee: I don't think so. It's only enabled and started if USE_CALICO is true.

multi-node/generic/controller-install.sh
tomdee: I thought it was OK to write files if they're not bring used?

Documentation/deploy-master.md
tomdee: I would err on the side of not explaining. The need to do this should go away with k8s 1.3. @aaronlevy WDYT?

Documentation/deploy-master.md
aaronlevy: We can, but this becomes a divergence from upstream. For now we are feature-flagging this and building separate images. Once this makes it upstream we can decide to start back-porting the change: https://github.com/kubernetes/kubernetes/pull/24896

multi-node/generic/controller-install.sh
aaronlevy: Sorry, I guess it would be `K8S_VER=v1.2.3_coreos_cni.0` (unless this is being set elsewhere?)

multi-node/generic/controller-install.sh
aaronlevy: sgtm

multi-node/generic/controller-install.sh
```diff
@@ -100,6 +114,40 @@ WantedBy=multi-user.target
...
+Environment=CALICO_DISABLE_FILE_LOGGING=true
+Environment=HOSTNAME=${ADVERTISE_IP}
+Environment=IP=${ADVERTISE_IP}
+Environment=FELIX_FELIXHOSTNAME=${ADVERTISE_IP}
+Environment=CALICO_NETWORKING=false
+Environment=NO_DEFAULT_POOLS=true
+Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
+ExecStart=/usr/bin/rkt run --inherit-env --stage1-from-dir=stage1-fly.aci \
+--volume=modules,kind=host,source=/lib/modules,readOnly=false \
+--mount=volume=modules,target=/lib/modules \
```
aaronlevy: I would like to test / evaluate that this is necessary.

/cc @mjg59

Documentation/deploy-workers.md
aaronlevy: sgtm

Documentation/deploy-master.md
aaronlevy: For posterity: from a previous discussion (and some of my own assumptions based on that discussion), the reason was:

The default policy that the calico-node sets is to deny all egress traffic from pods. Because calico-node starts before kubelet / pods then by the time the agent starts (as a pod) it will not be able to make any requests that would hit iptables rules.

By using host-networking & localhost, the agent pod bypasses any local rules when speaking with the apiserver. @caseydavenport please correct me if any of this is wrong.

multi-node/generic/controller-install.sh
aaronlevy: I meant the docker drop in which requires flannel to be running. I believe that should still be needed if not using CNI.

Documentation/deploy-master.md
aaronlevy: I'm fine with not explaining for now, but we can open up a side-issue if we want to add some docs. We've never had a question about it, short of people asking us to enable them.

multi-node/generic/controller-install.sh
tomdee: I figured that we shouldn't go overwriting things that the user can set, e.g. as a user I set `K8S_VER=v1.2.4_coreos_cni.0` and enable Calico and then it gets changed back to 1.2.3

multi-node/generic/controller-install.sh
tomdee: Ooops, yes! Good spot.

multi-node/generic/controller-install.sh
mjg59: This is currently necessary. The two ways to avoid it are:

1) Ensure that the ip_tables module is loaded before the container is initialised. The other modules should then be auto-loaded by the kernel
2) Add support to the kernel to automatically load the ip_tables module when it receives an iptables request. I have a patch to implement this and will test it shortly.

aaronlevy:
I believe the last thing that needs to be addressed from my end is that when not using CNI I believe we still need the `/etc/systemd/system/docker.service.d/40-flannel.conf` drop-ins.

@colhom any final review notes?

aaronlevy:
This LGTM. I'll hold off on an okay from @colhom, but feel free to merge after that.

colhom:
Passes conformance with and without useCalico=true. lgtm

edevil:
It used to be that the single-node [guide](https://coreos.com/kubernetes/docs/latest/deploy-master.html) could be used for a no-frills installation of Kubernetes with the minimal working configuration. But recently  this PR introduced yet another cog in the Kubernetes machine, Calico. Is this really needed for a simple installation? From the documentation it seems to perform a superset of the tasks of Flannel but they are both included and integrated. I think Calico should at least be an optional addon, or substitute Flannel entirely.

aaronlevy:
@edevil The goal here is still to provide a baseline installation of Kubernetes from which deployers can choose how the rest of their cluster functions.

In this regard, Calico is still an optional component. Per the guide you linked, when you get to the Calico installation section it states "This step can be skipped if not using Calico". But maybe we can improve the documents to be more clear on this point.

The reasoning for adding Calico here is that we wanted to provide a simple method for users to start developing against Kubernetes network policies - which are likely going to become a tier 1 api-object (as of v1.3 the objects are in the beta api): https://github.com/kubernetes/kubernetes/pull/24154

Similarly, we somewhat recently added heapster as a default addon - because it became a requirement for using the horizontal-pod-autoscaler functionality of Kubernetes.

We definitely want to keep these tools simple / no-frills to provide a baseline from which others can build atop. In the network policy case, this is a little difficult to completely move into an addon, because we need some policy enforcement (node agent) before the kubelet even starts. So the middle ground for now, which allows developers to easily build against upstream kubernetes functionality, is to document this step as optional.

edevil:
Thank you for the clarification. Then I think that the documentation can be improved. For example, this section, https://coreos.com/kubernetes/docs/latest/deploy-master.html#set-up-the-cni-config, does not say if it's optional or not. As it is Calico related I assume it is, but it's also related to Flannel.

On another note, from what I've read Calico can also provide an overlay network. If Calico was added but Flannel removed at least the number of components would stay the same.

robszumski:
@edevil What you are wishing for is the exact idea behind the [recently announced Canal project](https://coreos.com/blog/coreos-intel-calico-packet-extend-gifee.html#project-calico-and-coreos-deliver-cloud-native-networking-with-canal).

> Canal unites and integrates the connectivity capabilities of flannel, developed by CoreOS, with the robust network policy management of Project Calico.

Basically, you'll get the best of both as one package. As the project gets underway, look for it to be integrated here.

aaronlevy:
Since this PR is closed @edevil would you mind opening an issue about the documentation improvements (or even better a PR!)

PR #425, opened 2016-04-19
kube-aws: truncate configPath (cluster.yaml) on open

zuercher:
If the cluster.yaml file exists and is longer than the new version being
written, the resulting cluster.yaml will contain portions of the previous
version.


colhom:
@zuercher the desired behavior here is `kube-aws init` failing if `cluster.yaml` exists already. I think that check is simply missing.

zuercher:
OK. I'll close this one and submit another one that fails the open instead.

colhom:
Thanks @zuercher.

PR #427, opened 2016-04-20
kube-aws: configurable release channel

colhom:
\cc @cgag

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -114,6 +114,12 @@ const (
 	vpcLogicalName = "VPC"
 )
 
+var supportedReleaseChannels = map[string]bool{
+	"alpha":  true,
+	"beta":   true,
+	"stable": false,
+}
+
 func (c Cluster) Config() (*Config, error) {
 	config := Config{Cluster: c}
 	config.ETCDEndpoints = fmt.Sprintf("http://%s:2379", c.ControllerIP)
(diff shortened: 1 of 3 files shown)


four43:
Thanks @colhom  :)

multi-node/aws/pkg/config/config.go
```diff
@@ -329,6 +331,18 @@ func (cfg Cluster) valid() error {
 	if cfg.ExternalDNSName == "" {
 		return errors.New("externalDNSName must be set")
 	}
+
+	releaseChannelSupported := false
+	for _, channel := range supportedReleaseChannels {
```
aaronlevy: Could use a map for supportedReleaseChannels and would make this check a little cleaner.

multi-node/aws/pkg/config/config_test.go
```diff
@@ -253,3 +253,59 @@ func TestIsSubdomain(t *testing.T) {
...
+			conf: `
+releaseChannel: beta
+`,
+			channel: "beta",
+		},
+	}
+
+	invalidConfigs := []string{
+		`
+#TODO(chom): move this to validConfigs when stable is supported
```
aaronlevy: Seems weird to have the TODO as part of the test string. Also, as a nit, do these need to be multi-line strings. Can they not just be `[]string{"foo"}` instead?

colhom:
Good idea to use a map.

multi-node/aws/pkg/config/config_test.go
colhom: We need the newlines in the beginning and end, so that's why I have the multi-lines definitions.

I have the TODO in the test case so, if the test ever fails, we'll have a reminder there when it prints out the config block as part of error message.

aaronlevy:
k. lgtm with squash

PR #433, opened 2016-04-22
kube-aws: fail if configPath (cluster.yaml) already exists

zuercher:
If the cluster.yaml file exists and is longer than the new version being
written, the resulting cluster.yaml will contain portions of the previous
version. Open it with os.O_EXCL, causing an error if the file already
exists.

This replaces #425.


colhom:
LGTM. Thanks @zuercher

PR #436, opened 2016-04-26
vagrant: Increase default Virtualbox memory to 2GB

### LINKED ISSUE #311: Increase default Virtualbox memory to at least 4GB
The default of 1 GB is practically useless and leads to confusing problems for first-time users

mumoshu:
Without this change, we would end up with a non-functional k8s cluster set up because of memory exhaustion.

Also made it configurable via a constant defined at top of Vagrantfile according to https://github.com/coreos/coreos-kubernetes/issues/311#issuecomment-199522832

Tested with Virtualbox running `vagrant destroy -f  && vagrant up` and after several minutes running `kubectl --kubeconfig kubeconfig get svc` to confirm it's really up.

Fixes #311

P.S. Just fixing around to decrease num of non-critial issues to make coreos-kubernetes looks really good as it actually is 😄
Thanks for sharing!

### DIFF
--- a/single-node/Vagrantfile
+++ b/single-node/Vagrantfile
@@ -10,6 +10,7 @@ Vagrant.require_version ">= 1.6.0"
 
 CLUSTER_IP="10.3.0.1"
 NODE_IP = "172.17.4.99"
+NODE_MEMORY_SIZE = 2048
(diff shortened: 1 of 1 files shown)


aaronlevy:
lgtm

Issue #100, opened 2015-10-06
Support multi availability zone deployments on AWS


vyshane:
In addition to supporting HA masters by default (#90), it would also be good to have support for multi AZ deployments on AWS. A minimal HA setup would be 3 master nodes on 3 different AZs, with worker nodes also on different AZs.

ivankennethwang:
+1

danielwhatmuff:
+1

four43:
+1 This is a great idea. Having to add subnets and modify the autoscaling group after that fact is kind of a bummer. I'm assuming that's where some of the complexity lies...

arkkanoid:
+1

mumoshu:
Hi, I have sent the PR https://github.com/coreos/coreos-kubernetes/pull/439 for this.
Does it seem to solve your problem?

four43:
What a pull request! It has example snippets and a screenshot. Nice! That looks like it would work for my use case.

tigrawap:
Regarding https://github.com/coreos/coreos-kubernetes/pull/439  
Kubernetes itself does not support multi-AZ   
Provisioning workers alone does not solve the problem, there are more to it, for example AZ-bound EBS

See https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/federation.md

borsboom:
The [changelog for Kubernetes v1.2.0](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#major-themes) says:

> Kubernetes clusters can now span zones within a cloud provider. Pods from a service will be automatically spread across zones, enabling applications to tolerate zone failure.

Does that not apply to AWS?

four43:
We run a multi-az setup with our cluster. It's a fairly small cluster, but Kubernetes will avoid putting pods for the same rc on the same node. There is no guarantee however that those 2 nodes will be in different AZs. It sounds like that's planned for 1.2 however.

brandonweeks:
http://blog.kubernetes.io/2016/03/building-highly-available-applications-using-Kubernetes-new-multi-zone-clusters-a.k.a-Ubernetes-Lite.html

"Ubernetes Lite" is already present in Kubernetes 1.2, however @tigrawap is correct in saying this doesn't support persistent storage across AZs.

four43:
@brandonweeks - Storage was a mess for us, ended up using a different networked solution that resides on a different cluster completely. Glad to see the Kubernetes team addressing these issues.

[...]

PR #439, opened 2016-04-27
kube-aws: Support Multi-AZ workers on AWS

mumoshu:
One step forward to achieve high-availability throughout the cluster.
This change allows you to specify multiple subnets in cluster.yaml to make workers' ASG spread instances over those subnets. Differentiating each subnet's availability zone results in H/A of workers.
Beware that this change itself does nothing with H/A of masters.

Possibly relates to #147, #100

## How I have tested this

I have tested this running `./build && rm -rf cluster.yaml userdata/ credentials/ && bin/kube-aws init *several options* && bin/kube-aws render && bin/kube-aws up` with with/without the newly added `subnets` section in `cluster.yaml`.

For example, given a `cluster.yaml`:

```yaml
*snip*

subnets:
- instanceCIDR: "10.0.0.0/24"
  availabilityZone: ap-northeast-1a
- instanceCIDR: "10.0.1.0/24"
  availabilityZone: ap-northeast-1c
```

After modifying worker ASG's desired capacity to 2 result in 2 worker nodes:

```
$ kubectl --kubeconfig=kubeconfig get nodes
NAME                                            STATUS                     AGE
ip-10-0-0-125.ap-northeast-1.compute.internal   Ready                      10m
ip-10-0-0-50.ap-northeast-1.compute.internal    Ready,SchedulingDisabled   17m
ip-10-0-1-51.ap-northeast-1.compute.internal    Ready                      17m
```

spreading across availability zone(this time they are `ap-northeast-1a` and `ap-northeast-1c`)

image

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -32,7 +32,6 @@ func newDefaultCluster() *Cluster {
 		ClusterName:              "kubernetes",
 		ReleaseChannel:           "alpha",
 		VPCCIDR:                  "10.0.0.0/16",
-		InstanceCIDR:             "10.0.0.0/24",
 		ControllerIP:             "10.0.0.50",
 		PodCIDR:                  "10.2.0.0/16",
 		ServiceCIDR:              "10.3.0.0/24",
@@ -46,6 +45,7 @@ func newDefaultCluster() *Cluster {
 		WorkerRootVolumeSize:     30,
 		CreateRecordSet:          false,
 		RecordSetTTL:             300,
+		Subnets:                  []Subnet{},
 	}
 }
 
@@ -74,8 +74,23 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 	// as it will with RecordSets
 	c.HostedZone = WithTrailingDot(c.HostedZone)
 
+	// If the user specified no subnets, we assume that a single AZ configuration with the default instanceCIDR is demanded
+	if len(c.Subnets) == 0 && c.InstanceCIDR == "" {
+		c.InstanceCIDR = "10.0.0.0/24"
+	}
+
 	if err := c.valid(); err != nil {
 		return nil, fmt.Errorf("invalid cluster: %v", err)
+	}
+
+	// For backward-compatibility
+	if len(c.Subnets) == 0 {
+		c.Subnets = []Subnet{
+			{
+				AvailabilityZone: c.AvailabilityZone,
+				InstanceCIDR:     c.InstanceCIDR,
+			},
+		}
 	}
 	return c, nil
 }
@@ -109,6 +124,12 @@ type Cluster struct {
 	HostedZone               string            `yaml:"hostedZone"`
 	StackTags                map[string]string `yaml:"stackTags"`
 	UseCalico                bool              `yaml:"useCalico"`
+	Subnets                  []Subnet          `yaml:"subnets"`
+}
+
+type Subnet struct {
+	AvailabilityZone string `yaml:"availabilityZone"`
+	InstanceCIDR     string `yaml:"instanceCIDR"`
 }
 
 const (
@@ -160,8 +181,9 @@ type StackTemplateOptions struct {
 
 type stackConfig struct {
 	*Config
-	UserDataWorker     string
-	UserDataController string
+	UserDataWorker        string
+	UserDataController    string
+	ControllerSubnetIndex int
 }
 
 func execute(filename string, data interface{}, compress bool) (string, error) {
@@ -206,6 +228,25 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
 	}
 
 	stackConfig.Config.TLSConfig = compactAssets
+
+	controllerIPAddr := net.ParseIP(stackConfig.ControllerIP)
+	if controllerIPAddr == nil {
+		return nil, fmt.Errorf("invalid controllerIP: %s", stackConfig.ControllerIP)
+	}
+	controllerSubnetFound := false
+	for i, subnet := range stackConfig.Subnets {
+		_, instanceCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
+		if err != nil {
+			return nil, fmt.Errorf("invalid instanceCIDR: %v", err)
+		}
+		if instanceCIDR.Contains(controllerIPAddr) {
+			stackConfig.ControllerSubnetIndex = i
+			controllerSubnetFound = true
+		}
+	}
+	if !controllerSubnetFound {
+		return nil, fmt.Errorf("Fail-fast occurred possibly because of a bug: ControllerSubnetIndex couldn't be determined for subnets (%v) and controllerIP (%v)", stackConfig.Subnets, stackConfig.ControllerIP)
+	}
 
 	if stackConfig.UserDataWorker, err = execute(opts.WorkerTmplFile, stackConfig.Config, compressUserData); err != nil {
 		return nil, fmt.Errorf("failed to render worker cloud config: %v", err)
@@ -372,9 +413,6 @@ func (c Cluster) valid() error {
 	if c.Region == "" {
 		return errors.New("region must be set")
 	}
-	if c.AvailabilityZone == "" {
-		return errors.New("availabilityZone must be set")
-	}
 	if c.ClusterName == "" {
 		return errors.New("clusterName must be set")
 	}
@@ -391,26 +429,78 @@ func (c Cluster) valid() error {
 		return fmt.Errorf("invalid vpcCIDR: %v", err)
 	}
 
-	_, instancesNet, err := net.ParseCIDR(c.InstanceCIDR)
-	if err != nil {
-		return fmt.Errorf("invalid instanceCIDR: %v", err)
-	}
-	if !vpcNet.Contains(instancesNet.IP) {
-		return fmt.Errorf("vpcCIDR (%s) does not contain instanceCIDR (%s)",
-			c.VPCCIDR,
-			c.InstanceCIDR,
-		)
-	}
-
 	controllerIPAddr := net.ParseIP(c.ControllerIP)
 	if controllerIPAddr == nil {
 		return fmt.Errorf("invalid controllerIP: %s", c.ControllerIP)
 	}
-	if !instancesNet.Contains(controllerIPAddr) {
-		return fmt.Errorf("instanceCIDR (%s) does not contain controllerIP (%s)",
-			c.InstanceCIDR,
-			c.ControllerIP,
-		)
+
+	if len(c.Subnets) == 0 {
+		if c.AvailabilityZone == "" {
+			return fmt.Errorf("availabilityZone must be set")
+		}
+		_, instanceCIDR, err := net.ParseCIDR(c.InstanceCIDR)
+		if err != nil {
+			return fmt.Errorf("invalid instanceCIDR: %v", err)
+		}
+		if !vpcNet.Contains(instanceCIDR.IP) {
+			return fmt.Errorf("vpcCIDR (%s) does not contain instanceCIDR (%s)",
+				c.VPCCIDR,
+				c.InstanceCIDR,
+			)
+		}
+		if !instanceCIDR.Contains(controllerIPAddr) {
+			return fmt.Errorf("instanceCIDR (%s) does not contain controllerIP (%s)",
+				c.InstanceCIDR,
+				c.ControllerIP,
+			)
+		}
+	} else {
+		if c.InstanceCIDR != "" {
+			return fmt.Errorf("The top-level instanceCIDR(%s) must be empty when subnets are specified", c.InstanceCIDR)
+		}
+		if c.AvailabilityZone != "" {
+			return fmt.Errorf("The top-level availabilityZone(%s) must be empty when subnets are specified", c.AvailabilityZone)
+		}
+
+		var instanceCIDRs = make([]*net.IPNet, 0)
+		for i, subnet := range c.Subnets {
+			if subnet.AvailabilityZone == "" {
+				return fmt.Errorf("availabilityZone must be set for subnet #%d", i)
+			}
+			_, instanceCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
+			if err != nil {
+				return fmt.Errorf("invalid instanceCIDR for subnet #%d: %v", i, err)
+			}
+			instanceCIDRs = append(instanceCIDRs, instanceCIDR)
+			if !vpcNet.Contains(instanceCIDR.IP) {
+				return fmt.Errorf("vpcCIDR (%s) does not contain instanceCIDR (%s) for subnet #%d",
+					c.VPCCIDR,
+					c.InstanceCIDR,
+					i,
+				)
+			}
+		}
+
+		controllerInstanceCidrExists := false
+		for _, a := range instanceCIDRs {
+			if a.Contains(controllerIPAddr) {
+				controllerInstanceCidrExists = true
+			}
+		}
+		if !controllerInstanceCidrExists {
+			return fmt.Errorf("No instanceCIDRs in Subnets (%v) contain controllerIP (%s)",
+				instanceCIDRs,
+				c.ControllerIP,
+			)
+		}
+
+		for i, a := range instanceCIDRs {
+			for j, b := range instanceCIDRs[i+1:] {
+				if i > 0 && cidrOverlap(a, b) {
+					return fmt.Errorf("CIDR of subnet %d (%s) overlaps with CIDR of subnet %d (%s)", i, a, j, b)
+				}
+			}
+		}
 	}
 
 	_, podNet, err := net.ParseCIDR(c.PodCIDR)
--- a/multi-node/aws/pkg/config/templates/cluster.yaml
+++ b/multi-node/aws/pkg/config/templates/cluster.yaml
@@ -31,7 +31,7 @@ keyName: {{.KeyName}}
 # Region to provision Kubernetes cluster
 region: {{.Region}}
 
-# Availability Zone to provision Kubernetes cluster
+# Availability Zone to provision Kubernetes cluster when placing nodes in a single availability zone (not highly-available) Comment out for multi availability zone setting and use the below `subnets` section instead.
 availabilityZone: {{.AvailabilityZone}}
 
 # ARN of the KMS key used to encrypt TLS assets.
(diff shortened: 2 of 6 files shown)


rafamonteiro:
Just gave it a try and worked perfectly 👍

multi-node/aws/pkg/config/config.go
```diff
@@ -385,26 +398,32 @@ func (cfg Cluster) valid() error {
...
+			return fmt.Errorf("invalid instanceCIDR for subnet #%d: %v", i, err)
+		}
+		if !vpcNet.Contains(instancesNet.IP) {
+			return fmt.Errorf("vpcCIDR (%s) does not contain instanceCIDR (%s) for subnet #%d",
+				cfg.VPCCIDR,
+				cfg.InstanceCIDR,
+				i,
+			)
+		}
+		if i == 0 && !instancesNet.Contains(controllerIPAddr) {
```
aaronlevy: If we have this convention we should document it (maybe in config).

aaronlevy:
In general the code looks good. I'm not familiar enough with the upstream "ubernetes-lite" proposal and how well this would work with that (deferring to @colhom )

colhom:
@mumoshu I've read through the `kube-up` segment that deals with this, and AFAICT the process is the same there as you have it here. I'm going to kick off an e2e test tonight for this.

 [federation-lite proposal](https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/federation-lite.md)

multi-node/aws/pkg/config/config.go
mumoshu: Thanks for your comment!
I agree that we are better documenting it for now.

However we should also remove it when we introduce an auto-scaling group to manage controller(s) and an ELB pointing to it.
This "convention" remains here because we currently have a singleton controller(which has `controllerIP` pointing to it) and it is placed in the first subnet "for now"(choosing one of subnets for the singleton controller doesn't make sense but I have proceeded with this convention for now).

If we had an auto-scaling group managing controller(s), we would configure the group to spread EC2 instances evenly over multiple subnets(as I had done for workers in this PR).
Then, we don't need to choose first/second/last subnet to place the singleton controller or to specify/validate controllerIP(because we don't need it anymore. We will point the master from workers via DNS name of an load balancer or Route53 DNS name pointing to it).

Does this make sense to you? @aaronlevy @colhom

aaronlevy: Yup, longer-term plan sounds good to me.

My concern with documenting this in the `config.yaml` is that it's not immediately clear what is required of the networking configs (e.g. first subnet must container controllerIP).

However, I just checked the existing config.yaml, and looks like we're not being explicit about the any of the networking expectations - so this is probably something I can raise as a separate issue.

multi-node/aws/pkg/config/config.go
```diff
@@ -108,6 +118,12 @@ type Cluster struct {
 	RecordSetTTL             int               `yaml:"recordSetTTL"`
 	HostedZone               string            `yaml:"hostedZone"`
 	StackTags                map[string]string `yaml:"stackTags"`
+	Subnets                  []Subnet          `yaml:"subnets"`
```
colhom: This field (and it's subfields) need to be added to `pkg/config/templates/cluster.yaml`

colhom:
@mumoshu this is a breaking change, as `instanceCIDR` is no longer supported. That information is now expressed as the first element of the `subnets` array. I like this format more, but I hesitate to merge breaking changes.

**If** we were to go the route of breaking the config api, I would suggest that we add `ControllerIP` as a field of the `Subnet` struct. For now `valid()` could enforce that only one `Subnet.ControllerIP` is set, and later when mutli-controller support lands we can remove that restriction.

\cc @aaronlevy

cw-kuoka:
@colhom Thanks for your comments!
First of all, I don't want to make a breaking change as far as possible, too.

Excuse me if i'm missing the point. I've tried to make this not a breaking change by `subnets` settings to be completely optional (see [here](https://github.com/coreos/coreos-kubernetes/pull/439/files#diff-e2d9d007d15d3afa7dab291ed73f2d46R348) and [there](https://github.com/coreos/coreos-kubernetes/pull/439/files#diff-e2d9d007d15d3afa7dab291ed73f2d46R363)).

What I intended is that, even after this change, `instanceCIDR` is still supported so that if you specify `instanceCIDR` in the top-level in `cluster.yaml`, you "can" reference it from templates if you like.

cw-kuoka:
@colhom Let me add something to my previous comment.

Maybe not about config api, but I did make the singleton-controller-part of the template refer to not the top-level `instanceCIDR` but the one in the first subnet [here](https://github.com/coreos/coreos-kubernetes/pull/439/files#diff-46f2be30feea4011eaebbb141603bde0R237) believing it's clearer(just thought that referring to the top-level `instanceCIDR` when you have 2-subnets won't make sense to users).

I guess I can make singleton-controller-part of the template refer to the top-level `instanceCIDR` (also with the according changes to `config.go`).
Then, I guess that both config and template are backward-compatible.
Would it make sense to you?

🙇

mumoshu:
@colhom Thanks for your comments!
First of all, I don't want to make a breaking change as far as possible, too.

Excuse me if i'm missing the point. I've tried to make this not to be a breaking change by making subnets settings to be completely optional (see here and there).

What I intended is that, even after this change, instanceCIDR is still supported so that if you specify instanceCIDR in the top-level in cluster.yaml, you "can" reference it from templates if you like.

mumoshu:
@colhom Let me add something to my previous comment.

Maybe not about config api, but I did make the singleton-controller-part of the template refer to not the top-level instanceCIDR but the one in the first subnet here believing it's clearer(just thought that referring to the top-level instanceCIDR when you have 2-subnets won't make sense to users).

I guess I can make singleton-controller-part of the template refer to the top-level instanceCIDR (also with the according changes to config.go).
Then, I guess that both config and template are backward-compatible.
Would it make sense to you?

🙇

multi-node/aws/pkg/config/config.go
mumoshu: I have the same feeling about your concern .
Also, at this point, a separate issue sounds O.K. for me to keep us proceeding incrementally.

multi-node/aws/pkg/config/config.go
mumoshu: I have added documentation in the commit https://github.com/coreos/coreos-kubernetes/pull/439/commits/7b5813187cf80101e5ad2def653f3b7f19357e84
Would you mind reviewing it? :bow:

multi-node/aws/pkg/config/config.go
```diff
@@ -74,6 +75,15 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 	// as it will with RecordSets
 	c.HostedZone = WithTrailingDot(c.HostedZone)
 
+	if len(c.Subnets) == 0 {
+		c.Subnets = []Subnet{
```
colhom: definitely appreciate the effort to keep backwards compatibility here. Given that this is clearly documented, we can later think about deprecating `instanceCIDR` and `availabilityZone`.

colhom:
@mumoshu I agree with your approach to keep backwards comparability. I'd like to see a few more things.

* If `len(config.Subnets) > 0`, then ensure in `config.Valid()` that `instanceCIDR` and `availbilityZone` are empty. Otherwise error out.

* Add the "this is only for single-AZ mode" warning to `availabilityZone` in `templates/cluster.yaml` as well

* To figure out which subnet the controller goes in, loop through all the subnets and pick the the one with an `instanceCIDR` which contains the `controllerIP`. Assuming all subnets are non-overlapping, the answer should either be a single subnet or an error (no subnets contain controllerIP)

multi-node/aws/pkg/config/config.go
```diff
@@ -385,26 +398,32 @@ func (cfg Cluster) valid() error {
...
 		return fmt.Errorf("invalid controllerIP: %s", cfg.ControllerIP)
 	}
-	if !instancesNet.Contains(controllerIPAddr) {
-		return fmt.Errorf("instanceCIDR (%s) does not contain controllerIP (%s)",
-			cfg.InstanceCIDR,
-			cfg.ControllerIP,
-		)
+
+	for i, subnet := range cfg.Subnets {
+		if subnet.AvailabilityZone == "" {
```
colhom: We should also be checking that non of the subnets overlap with eachother.

mumoshu:
@colhom Thanks! I have added some commits according to your very agreeable comments.
I appreciate it if you could look into those too.

multi-node/aws/pkg/config/config.go
```diff
@@ -385,26 +426,78 @@ func (cfg Cluster) valid() error {
...
 	if controllerIPAddr == nil {
 		return fmt.Errorf("invalid controllerIP: %s", cfg.ControllerIP)
 	}
-	if !instancesNet.Contains(controllerIPAddr) {
-		return fmt.Errorf("instanceCIDR (%s) does not contain controllerIP (%s)",
-			cfg.InstanceCIDR,
-			cfg.ControllerIP,
-		)
+
+	if len(cfg.Subnets) == 0 {
```
colhom: Isn't this case unreachable? [here](https://github.com/mumoshu/coreos-kubernetes/blob/multi-az-workers/multi-node/aws/pkg/config/config.go#L87) you ensure that subnets always has at least one element.

colhom: Oops sorry, scratch that ;)

colhom: Question: If we move [this](https://github.com/mumoshu/coreos-kubernetes/blob/multi-az-workers/multi-node/aws/pkg/config/config.go#L87) block of logic __before__ `c.valid()`, can't we do away with this conditional validation behavior and only validate the `cfg.Subnets` array, as the backwards compatibility measure would have already been taken?

multi-node/aws/pkg/config/config.go
```diff
@@ -403,29 +403,39 @@ func (cfg Cluster) valid() error {
...
+		instanceCIDRs = append(instanceCIDRs, instanceCIDR)
+		if !vpcNet.Contains(instanceCIDR.IP) {
 			return fmt.Errorf("vpcCIDR (%s) does not contain instanceCIDR (%s) for subnet #%d",
 				cfg.VPCCIDR,
 				cfg.InstanceCIDR,
 				i,
 			)
 		}
-		if i == 0 && !instancesNet.Contains(controllerIPAddr) {
+		if i == 0 && !instanceCIDR.Contains(controllerIPAddr) {
```
colhom: We're still enforcing that the controller be a member of the first subnet here.

multi-node/aws/pkg/config/config.go
```diff
@@ -403,29 +403,39 @@ func (cfg Cluster) valid() error {
...
 			return fmt.Errorf("instanceCIDR (%s) does not contain controllerIP (%s)",
 				subnet.InstanceCIDR,
 				cfg.ControllerIP,
 			)
 		}
 	}
 
+	for i, a := range instanceCIDRs {
+		for j, b := range instanceCIDRs[i+1:] {
+			if i > 0 && cidrOverlap(a, b) {
```
colhom: Why do we need to check that `i > 0` here? If I'm reading this code correctly, it will always fail to check that `instanceCIDRs[0]` overlaps with `instanceCIDRs[1]`.

colhom:
@mumoshu thanks for the changes, they look good.

I have a few minor comments I've left.  I feel like the code could be a bit cleaner if `instanceCIDR/availablilityZone` were canonicalized into the `subnets` array __before__ `cluster.valid()` is called in the `ClusterFromBytes()` method. Let me know what you think about this?

multi-node/aws/pkg/config/config.go
```diff
@@ -227,6 +228,25 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
...
+		return nil, fmt.Errorf("invalid controllerIP: %s", stackConfig.ControllerIP)
+	}
+	controllerSubnetFound := false
+	lastSubnetIndex := len(stackConfig.Subnets) - 1
+	for i, subnet := range stackConfig.Subnets {
+		_, instanceCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
+		if err != nil {
+			return nil, fmt.Errorf("invalid instanceCIDR: %v", err)
+		}
+		if instanceCIDR.Contains(controllerIPAddr) {
```
colhom: I see that my prior comment about "still checking controller is in first subnet" has already been fixed! Sorry, I blame github for being weird ;)

multi-node/aws/pkg/config/config.go
mumoshu: @colhom Thanks for your question!

I guess doing that([overwriting `InstanceCIDR`](https://github.com/mumoshu/coreos-kubernetes/blob/multi-az-workers/multi-node/aws/pkg/config/config.go#L78) with a default value and determining [implicit `Subnets`](https://github.com/mumoshu/coreos-kubernetes/blob/multi-az-workers/multi-node/aws/pkg/config/config.go#L86) from the top-level `InstanceCIDR` and `AvailabilityZone`) before validation prevents us doing the validation: `Did the user specified both top-level InstanceCIDR/AvailabilityZone and Subnets?` defined [here](https://github.com/mumoshu/coreos-kubernetes/blob/multi-az-workers/multi-node/aws/pkg/config/config.go#L455) with the [test](https://github.com/mumoshu/coreos-kubernetes/blob/multi-az-workers/multi-node/aws/pkg/config/config_test.go#L415)

It's a bit confusing but, thinking more generally, I took `c.valid()` as a validation of an user input. To structurally validate the user input, I thought it would be better not to structurally touch it before validation.

Does this make sense to you?

Btw, the order of `valid()` call and backward-compatiblity logic, defaults are not very intuitive in my code and I wish I could somehow make it more intuitive, too 👍
Just not coming up with a nicer idea though 😢

mumoshu:
@colhom Thanks for your review! I have replied to you in https://github.com/coreos/coreos-kubernetes/pull/439#discussion_r62424617

multi-node/aws/pkg/config/config.go
mumoshu: Nice catch 👍 I've realized that while testing and force-pushed a squashed commit before your review 😅

multi-node/aws/pkg/config/config.go
```diff
@@ -203,6 +228,25 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
...
+			stackConfig.ControllerSubnetIndex = i
+			controllerSubnetFound = true
+		} else if !controllerSubnetFound && i == lastSubnetIndex {
+			return nil, fmt.Errorf("Fail-fast occurred possibly because of a bug: ControllerSubnetIndex couldn't be determined for subnets (%v) and controllerIP (%v)", stackConfig.Subnets, stackConfig.ControllerIP)
+		}
```
cgag: If I understand this loop correctly, I think it could be made a bit more simple by dropiing `lastSubnetIndex` and moving this final check
outside of the loop:

```
	controllerSubnetFound := false
	for i, subnet := range stackConfig.Subnets {
		_, instanceCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
		if err != nil {
			return nil, fmt.Errorf("invalid instanceCIDR: %v", err)
		}
		if instanceCIDR.Contains(controllerIPAddr) {
			stackConfig.ControllerSubnetIndex = i
			controllerSubnetFound = true
		}
	}
	if !controllerSubnetFound {
		return nil, fmt.Errorf("Fail-fast occurred possibly because of a bug: ControllerSubnetIndex couldn't be determined for subnets (%v) and controllerIP (%v)", stackConfig.Subnets, stackConfig.ControllerIP)
	}
```

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -64,10 +64,17 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # CIDR for Kubernetes VPC. If vpcId is specified, must match the CIDR of existing vpc.
 # vpcCIDR: "10.0.0.0/16"
 
-# CIDR for Kubernetes subnet
+# CIDR for Kubernetes subnet when placing nodes in a single availability zone (not highly-available) Leave commented out for multi availability zone setting and use the below `subnets` section instead.
 # instanceCIDR: "10.0.0.0/24"
 
-# IP Address for controller in Kubernetes subnet
+# Kubernetes subnets with their CIDRs and availability zones. Differentiating availability zone for 2 or more subnets result in high-availability (failures of a single availability zone won't result in immediate downtimes)
```
cgag: Should we make any notes about EBS volumes being zone specific?  From what I can tell, if you're safe unless you depend on an EBS volume in the AZ that goes down.

multi-node/aws/pkg/config/templates/stack-template.json
```diff
@@ -226,7 +234,7 @@
     },
     "InstanceController": {
       "Properties": {
-        "AvailabilityZone": "{{.AvailabilityZone}}",
+        "AvailabilityZone": "{{(index .Subnets 0).AvailabilityZone}}",
```
cgag: Should this be  .ControllerSubnetIndex?

multi-node/aws/pkg/config/config.go
```diff
@@ -74,9 +74,27 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
...
+	}
+
 	if err := c.valid(); err != nil {
 		return nil, fmt.Errorf("invalid cluster: %v", err)
 	}
+
+	// For backward-compatibility
+	if len(c.Subnets) == 0 {
+		if c.InstanceCIDR == "" {
+			c.InstanceCIDR = "10.0.0.0/24"
```
cgag: I think this `if c.InstanceCIDR` block is unnecessary since it will have been done before the call to valid(), and valid shouldn't change the value of `c.InstanceCider`.

multi-node/aws/pkg/config/config.go
```diff
@@ -74,9 +74,27 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
...
+	// If the user specified no subnets, we assume that a single AZ configuration with the default instanceCIDR is demanded
+	if len(c.Subnets) == 0 && c.InstanceCIDR == "" {
+		c.InstanceCIDR = "10.0.0.0/24"
+	}
+
 	if err := c.valid(); err != nil {
 		return nil, fmt.Errorf("invalid cluster: %v", err)
 	}
+
+	// For backward-compatibility
```
cgag: I think this comment should be more clear what the backwards-compatibility is for, it'll be unclear outside
the context of the commit.  It also implies to me that single AZ deployments are being deprecated, is that going to be the case (cc @colhom)?

Perhaps it should just say something like "fall back to single AZ if no subnets are specified"?

multi-node/aws/pkg/config/templates/stack-template.json
mumoshu: Definitely yes 👍 Addressed in 78e067f

multi-node/aws/pkg/config/config.go
mumoshu: Thanks! Addressed in ea9f366

multi-node/aws/pkg/config/config.go
mumoshu: Thanks! Addressed in f9c4f4c

colhom:
@mumoshu sorry for the delay- this looks good. Rebase and squash the commits, and I'll run it through tests.

mumoshu:
@colhom Thanks, I have just rebased/squashed this 👍

colhom:
Thanks @mumoshu !

PR #443, opened 2016-04-28
Fix typo: cloud-config files are in userdata/, not cloud-config/

mahata:


colhom:
@mahata the idea is that cloud-config is a type a of userdata. In the future, we plan to transition from cloud-config to [ignition](https://github.com/coreos/ignition).

mahata:
Yeah, but the thing is that the online document is saying `kube-aws render` creates `cloud-config-worker` and `cloud-config-controller` in `cloud-config/` directory. But these files are actually created in `userdata/` directory. That's what I wanted to fix.

* https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html

colhom:
@mahata my mistake, I see now.

PR #449, opened 2016-05-02
kube-aws: hostedZoneId parameter

### LINKED ISSUE #441: Use HostedZone ID instead of HostedZone Name
For this part of cluster.yaml when using AWS

```yaml
# The name of the hosted zone to add the externalDNSName to, 
# E.g: "google.com".  This needs to already exist, kube-aws will not create
# it for you.
hostedZone:
```

I have a public and private hosted zone with the same name, so providing the name causes the creation to ultimately fail because of it. If we instead provided Hosted Zone ID, I don't think that would be a problem

colhom:
Allows specifying a hosted zone by ID, in the case that multiple hosted zones exist with the same name

fixes #441 

WIP for now, \cc @cgag

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -309,6 +309,7 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
 type r53Service interface {
 	ListHostedZonesByName(*route53.ListHostedZonesByNameInput) (*route53.ListHostedZonesByNameOutput, error)
 	ListResourceRecordSets(*route53.ListResourceRecordSetsInput) (*route53.ListResourceRecordSetsOutput, error)
+	GetHostedZone(*route53.GetHostedZoneInput) (*route53.GetHostedZoneOutput, error)
 }
 
 func (c *Cluster) validateDNSConfig(r53 r53Service) error {
@@ -316,26 +317,63 @@ func (c *Cluster) validateDNSConfig(r53 r53Service) error {
 		return nil
 	}
 
-	zonesResp, err := r53.ListHostedZonesByName(&route53.ListHostedZonesByNameInput{
-		DNSName: aws.String(c.HostedZone),
-	})
-	if err != nil {
-		return fmt.Errorf("Error validating HostedZone: %s", err)
+	if c.HostedZoneID == "" {
+		//TODO(colhom): When HostedZone parameter is gone, this block can be removed
+		//Config will gaurantee that HostedZoneID is set from the get-go
+		listHostedZoneInput := route53.ListHostedZonesByNameInput{
+			DNSName: aws.String(c.HostedZone),
+		}
+
+		zonesResp, err := r53.ListHostedZonesByName(&listHostedZoneInput)
+		if err != nil {
+			return fmt.Errorf("Error validating HostedZone: %s", err)
+		}
+
+		zones := zonesResp.HostedZones
+
+		if len(zones) == 0 {
+			return fmt.Errorf("hosted zone %s does not exist", c.HostedZone)
+		}
+
+		var matchingZone *route53.HostedZone
+		for _, zone := range zones {
+			if aws.StringValue(zone.Name) == c.HostedZone {
+				if matchingZone != nil {
+					//This means we've found another match, and HostedZone is ambiguous
+					return fmt.Errorf("multiple hosted-zones found for DNS name \"%s\"", c.HostedZone)
+				}
+				matchingZone = zone
+			} else {
+				/* Weird API semantics: if we see a zone which doesn't match the name
+				   we've exhausted all zones which match the name
+				  http://docs.aws.amazon.com/cli/latest/reference/route53/list-hosted-zones-by-name.html#options */
+
+				break
+			}
+		}
+		if matchingZone == nil {
+			return fmt.Errorf("hosted zone %s does not exist", c.HostedZone)
+		}
+		c.HostedZoneID = aws.StringValue(matchingZone.Id)
 	}
 
-	zones := zonesResp.HostedZones
-	if len(zones) == 0 || (*zones[0].Name != c.HostedZone) {
-		return fmt.Errorf(
-			"HostedZone %s does not exist.  You'll need to create it manually",
-			c.HostedZone,
-		)
+	hzOut, err := r53.GetHostedZone(&route53.GetHostedZoneInput{Id: aws.String(c.HostedZoneID)})
+	if err != nil {
+		return fmt.Errorf("error getting hosted zone %s: %v", c.HostedZoneID, err)
+	}
+
+	if !isSubdomain(c.ExternalDNSName, aws.StringValue(hzOut.HostedZone.Name)) {
+		return fmt.Errorf("externalDNSName %s is not a sub-domain of hosted-zone %s", c.ExternalDNSName, aws.StringValue(hzOut.HostedZone.Name))
 	}
 
 	recordSetsResp, err := r53.ListResourceRecordSets(
 		&route53.ListResourceRecordSetsInput{
-			HostedZoneId: zones[0].Id,
+			HostedZoneId: hzOut.HostedZone.Id,
 		},
 	)
+	if err != nil {
+		return fmt.Errorf("error listing record sets for hosted zone id = %s: %v", c.HostedZoneID, err)
+	}
 
 	if len(recordSetsResp.ResourceRecordSets) > 0 {
 		for _, recordSet := range recordSetsResp.ResourceRecordSets {
@@ -372,4 +410,25 @@ func stackEventErrMsgs(events []*cloudformation.StackEvent) []string {
 	}
 
 	return errMsgs
+}
+
+func isSubdomain(sub, parent string) bool {
+	sub, parent = config.WithTrailingDot(sub), config.WithTrailingDot(parent)
+	subParts, parentParts := strings.Split(sub, "."), strings.Split(parent, ".")
+
+	if len(parentParts) > len(subParts) {
+		return false
+	}
+
+	subSuffixes := subParts[len(subParts)-len(parentParts):]
+
+	if len(subSuffixes) != len(parentParts) {
+		return false
+	}
+	for i := range subSuffixes {
+		if subSuffixes[i] != parentParts[i] {
+			return false
+		}
+	}
+	return true
 }
--- a/multi-node/aws/pkg/cluster/cluster_test.go
+++ b/multi-node/aws/pkg/cluster/cluster_test.go
@@ -220,7 +220,7 @@ type dummyR53Service struct {
 func (r53 dummyR53Service) ListHostedZonesByName(input *route53.ListHostedZonesByNameInput) (*route53.ListHostedZonesByNameOutput, error) {
 	output := &route53.ListHostedZonesByNameOutput{}
 	for _, zone := range r53.HostedZones {
-		if zone.DNS == config.WithTrailingDot(*input.DNSName) {
+		if zone.DNS == config.WithTrailingDot(aws.StringValue(input.DNSName)) {
 			output.HostedZones = append(output.HostedZones, &route53.HostedZone{
 				Name: aws.String(zone.DNS),
 				Id:   aws.String(zone.Id),
@@ -232,7 +232,7 @@ func (r53 dummyR53Service) ListHostedZonesByName(input *route53.ListHostedZonesB
 
 func (r53 dummyR53Service) ListResourceRecordSets(input *route53.ListResourceRecordSetsInput) (*route53.ListResourceRecordSetsOutput, error) {
 	output := &route53.ListResourceRecordSetsOutput{}
-	if name, ok := r53.ResourceRecordSets[*input.HostedZoneId]; ok {
+	if name, ok := r53.ResourceRecordSets[aws.StringValue(input.HostedZoneId)]; ok {
 		output.ResourceRecordSets = []*route53.ResourceRecordSet{
 			&route53.ResourceRecordSet{
 				Name: aws.String(name),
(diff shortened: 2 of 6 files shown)


multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -302,18 +302,26 @@ func (c *Cluster) validateDNSConfig(r53 r53Service) error {
...
+
+	zonesResp, err := r53.ListHostedZonesByName(&listHostedZoneInput)
 	if err != nil {
 		return fmt.Errorf("Error validating HostedZone: %s", err)
 	}
 
 	zones := zonesResp.HostedZones
-	if len(zones) == 0 || (*zones[0].Name != c.HostedZone) {
+
+	zoneNameMismatch := func() bool { return *zones[0].Name != c.HostedZone }
```
colhom: @cgag should we be using the trailing dot form of `c.HostedZone` for this check?

colhom:
also \cc @rwillard

multi-node/aws/pkg/cluster/cluster.go
cgag: The trailing dot should always be there on `c.HostedZone`, it gets added in `config.ClusterFromBytes`

aaronlevy: You're no longer checking len(zones) before referencing the underlying object. Is this safe? I didn't dig deep enough in the sdk to see how it's implemented.

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -24,6 +24,10 @@ externalDNSName: {{.ExternalDNSName}}
 # it for you.
 #hostedZone: ""
 
+# The ID of hosted zone to add the externalDNSName to.
+# Only necessary if there are multiple hosted zones with the same name
+#hostedZoneId: ""
```
aaronlevy: Should we use this exclusively rather than `hostedZone`?

colhom: I'd say no for two reasons:
* We'd break a released API, though we are pre-1.0
* In most cases, I would say that there won't be two hosted zones with the same name on a single AWS account. I think it's reasonable to save 9/10 people a minute each at the cost of making things every so slightly more complicated for the 1/10. 

I could be very wrong about the second point, in which case we should break the API and just use the ID.

multi-node/aws/pkg/cluster/cluster.go
cgag: I made the same comment offline, turns out it's still checked before it's invoked.

multi-node/aws/pkg/config/templates/cluster.yaml
aaronlevy: Seems reasonable not to break this configurable, but may be worth commenting `hostedZone` as deprecated and to use the ID instead (if that's our plan) - then at some release point drop the option.

multi-node/aws/pkg/cluster/cluster.go
aaronlevy: Woops, it is... 2 lines down..

multi-node/aws/pkg/config/templates/cluster.yaml
aaronlevy: Actually, I misread. So you want to keep both configurables so that the common case (by name) is easier to specify.

I would lean toward trying to get people to use the ID exclusively. This could change over time (a new hosted zone with same name is added later) - and then how does this tool react on re-use (e.g. upgrades)? 

Is it difficult to extract the hostedZoneID?

cgag: It's not hard to find find the hostedZoneID, it's just very easy to remember the name.  It's nice to not have to go digging around in console or looking up aws-cli commands to use this feature.  It seems like requiring hostedZoneID adds a decent amount of friction.  Maybe we could list hosted zones and check if it's ambiguous as part of the validations in `kube-aws up` and then tell people they need to use the ID?

The wording on this seems to imply that hostedZoneID is only usable to disambiguate the hostedZoneName, but you could use hostedZoneID exclusively if you wanted.  Not sure if explaining that is worth the complexity it adds to figuring out how to configure cluster.yaml.

If we don't change the scope of this PR, the code LGTM.

colhom:
@aaronlevy thoughts on this?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -316,24 +316,42 @@ func (c *Cluster) validateDNSConfig(r53 r53Service) error {
...
-		return fmt.Errorf(
-			"HostedZone %s does not exist.  You'll need to create it manually",
-			c.HostedZone,
-		)
+
+	if len(zones) == 0 {
+		return fmt.Errorf("hosted zone %s does not exist", c.HostedZone)
+	}
+
+	var matchingZone *route53.HostedZone = nil
```
aaronlevy: nit: unnecessary `= nil`

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -316,24 +316,42 @@ func (c *Cluster) validateDNSConfig(r53 r53Service) error {
...
+	for _, zone := range zones {
+		if aws.StringValue(zone.Name) == c.HostedZone {
+			//TODO:(chom) for v1, we will require HostedZoneID be
+			//set, as HostedZone will be going away
```
aaronlevy: If the plan is to deprecate `HostedZone`, then we should allow a person to specify *only* a `HostedZoneID`.  Looks like this still requires someone to provide `HostedZone`. I would probably expect the steps to be:

1. If `HostedZoneID` is specified: use it (no ambiguity). Fail on no zones found.
1. If only `HostedZone` is specified, fail on len(zones)>1 or no zones found.

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -24,6 +24,10 @@ externalDNSName: {{.ExternalDNSName}}
 # it for you.
 #hostedZone: ""
```
aaronlevy: Per my previous comment, if we are deprecating this flag, we should allow the user to *not* specify it. Then also add a comment here noting that this flag is deprecated in favor of `hostedZoneID`

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -391,3 +402,24 @@ func stackEventErrMsgs(events []*cloudformation.StackEvent) []string {
 
 	return errMsgs
 }
+
+func isSubdomain(sub, parent string) bool {
```
colhom: moved from config package

multi-node/aws/pkg/config/config_test.go
```diff
@@ -203,68 +201,6 @@ dnsServiceIP: 10.6.142.100
 
 }
 
-func TestIsSubdomain(t *testing.T) {
```
colhom: moved to cluster package

colhom:
@aaronlevy PTAL at the 2nd commit. `hostedZone` is now treated as an "alternate way" of specifying `hostedZoneId`, and will easily be removed once it comes time.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -309,51 +309,62 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
...
-	var matchingZone *route53.HostedZone = nil
-	for _, zone := range zones {
-		if aws.StringValue(zone.Name) == c.HostedZone {
-			//TODO:(chom) for v1, we will require HostedZoneID be
-			//set, as HostedZone will be going away
-			if c.HostedZoneID == "" || aws.StringValue(zone.Id) == c.HostedZoneID {
+		var matchingZone *route53.HostedZone = nil
+		for _, zone := range zones {
+			if aws.StringValue(zone.Name) == c.HostedZone {
 				matchingZone = zone
```
aaronlevy: Is it safe to assume that we can use the first matching zone, or should we fail on multiple matches?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -309,51 +309,62 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
...
+			return fmt.Errorf("hosted zone %s does not exist", c.HostedZone)
+		}
 
-	var matchingZone *route53.HostedZone = nil
-	for _, zone := range zones {
-		if aws.StringValue(zone.Name) == c.HostedZone {
-			//TODO:(chom) for v1, we will require HostedZoneID be
-			//set, as HostedZone will be going away
-			if c.HostedZoneID == "" || aws.StringValue(zone.Id) == c.HostedZoneID {
+		var matchingZone *route53.HostedZone = nil
```
aaronlevy: nit: unnecessary `= nil`

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -309,51 +309,62 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
...
+		c.HostedZoneID = aws.StringValue(matchingZone.Id)
+	}
+
+	hzOut, err := r53.GetHostedZone(&route53.GetHostedZoneInput{Id: aws.String(c.HostedZoneID)})
+	if err != nil {
+		return fmt.Errorf("error getting hosted zone %s: %v", c.HostedZoneID, err)
+	}
+
+	if !isSubdomain(c.ExternalDNSName, aws.StringValue(hzOut.HostedZone.Name)) {
+		return fmt.Errorf("externalDNSName %s is not a sub-domain of hosted-zone %s", c.ExternalDNSName, aws.StringValue(hzOut.HostedZone.Name))
```
aaronlevy: I guess this would catch a potential mismatch of "first hosted zone found" from comment above. But do we want this implicit functionality? e.g. this could work for a while, then it just so happens that it stops working because the hostedzone is no longer the first (and therefor matching) zone.

I'll defer to you on this. I'm not sure if order is fixed or if we should just hard-fail on multiple.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -309,51 +309,62 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
...
 	}
 
 	recordSetsResp, err := r53.ListResourceRecordSets(
 		&route53.ListResourceRecordSetsInput{
-			HostedZoneId: matchingZone.Id,
+			HostedZoneId: hzOut.HostedZone.Id,
 		},
 	)
+	if err != nil {
+		return fmt.Errorf("error listing record sets for hosted zone id = %s", c.HostedZoneID)
```
aaronlevy: do we want to include the error text here?

multi-node/aws/pkg/cluster/cluster_test.go
```diff
@@ -300,6 +321,7 @@ hostedZoneId: /hostedzone/staging_id_3  #non-existant hostedZoneId
 		clusterConfig, err := config.ClusterFromBytes([]byte(configBody))
 		if err != nil {
 			t.Errorf("could not get valid cluster config: %v", err)
+			continue
```
aaronlevy: Should we just use `Fatalf()` here instead?

multi-node/aws/pkg/cluster/cluster_test.go
```diff
@@ -313,6 +335,7 @@ hostedZoneId: /hostedzone/staging_id_3  #non-existant hostedZoneId
...
```
aaronlevy: Fatalf()?

multi-node/aws/pkg/cluster/cluster_test.go
aaronlevy: Hmm. Maybe not (if we assume we should continue testing rest of configs).

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -25,7 +25,7 @@ externalDNSName: {{.ExternalDNSName}}
...
```
aaronlevy: Add deprecation note to comment comment above (people shouldn't need to wait until running to find out it's deprecated)

multi-node/aws/pkg/cluster/cluster_test.go
colhom: Yep. This error indicates an error parsing a single config, doesn't necessarily mean any others will fail.

multi-node/aws/pkg/cluster/cluster.go
colhom: I wanted there to be an air of mystery ;) Will fix

multi-node/aws/pkg/cluster/cluster.go
colhom: Good point. We should definitely fail if it's ambiguous.

multi-node/aws/pkg/cluster/cluster.go
colhom: hard fail on multiple is the answer here.

colhom:
@aaronlevy PTAL ;)

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -309,33 +309,71 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
...
+		zones := zonesResp.HostedZones
+
+		if len(zones) == 0 {
+			return fmt.Errorf("hosted zone %s does not exist", c.HostedZone)
+		}
+
+		var matchingZone *route53.HostedZone
+		for _, zone := range zones {
+			if aws.StringValue(zone.Name) == c.HostedZone {
+				if matchingZone != nil {
```
aaronlevy: Why not just `if len(zones) > 1` -- then you don't need this loop at all - you got 1 matching zone by name.

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -309,33 +309,71 @@ func (c *Cluster) validateKeyPair(ec2Svc ec2Service) error {
...
+		if err != nil {
+			return fmt.Errorf("Error validating HostedZone: %s", err)
+		}
+
+		zones := zonesResp.HostedZones
+
+		if len(zones) == 0 {
+			return fmt.Errorf("hosted zone %s does not exist", c.HostedZone)
+		}
+
```
aaronlevy: Why not just check `if len(zones) > 1` and fail here. Then you don't need the loop below because you only have a single hostedzone which matches by name.

aaronlevy:
One more review comment

multi-node/aws/pkg/cluster/cluster.go
colhom: Unfortunately `ListHostedZonesByNameInput.DNSName` only controls lexical ordering of results. The way this api call works is very weird, but tldr for this use case is: 

you'll get a list that has two contiguous segments. It starts with all the hosted zones that match `DNSName` exactly followed by hosted zones which don't. Either or both segments may be zero length.

I get the feeling they just wrapped their fuzzy search api for this functionality or something crazy like that.

aaronlevy: Ah. k

aaronlevy:
lgtm. Can we set a target version for dropping legacy support?

colhom:
@aaronlevy ref #511

PR #454, opened 2016-05-03
kube-aws: Improve stack creation failed error messaging

### LINKED ISSUE #430: kube-aws: Provide more contexts on stack creation failed?
Hi, Thanks for sharing this!

I have just encountered the following error on `kube-aws up`:

```
$ kube-aws up
Creating AWS resources. This should take around 5 minutes.
Error: Error creating cluster: Stack creation failed: CREATE_FAILED : The following resource(s) failed to create: [VPC, IAMRoleController, InternetGateway, IAMRoleWorker].
```

Looking into AWS console revealed that its cause was that CloudFormation failed to create a `AWS::EC2::InternetGateway`:
image

It was not very obvious just reading the message from `kube-aws`.
I guess if we could provide more context one can deal with errors more quickly.

Without thinking much, something like the following seems helpful for me:

> Stack creation failed.
> Printing recent events from Cfn which may or may not indicate what was problem:
> *snip*
> CREATE_FAILED	AWS::EC2::InternetGateway	InternetGateway	The maximum number of internet gateways has been reached.

Any thoughts?

cgag:
Previously you were only told which resources cloudformation failed to
create.  We now print the event details associated with each failure.

sample output:

2016-05-03-134713_959x209_scrot

fixes #430

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"strings"
 	"text/tabwriter"
 	"time"
 
@@ -102,7 +103,11 @@ func (c *Cluster) validateExistingVPCState(ec2Svc ec2Service) error {
 
 	if *existingVPC.CidrBlock != c.VPCCIDR {
 		//If this is the case, our network config validation cannot be trusted and we must abort
-		return fmt.Errorf("configured vpcCidr (%s) does not match actual existing vpc cidr (%s)", c.VPCCIDR, *existingVPC.CidrBlock)
+		return fmt.Errorf(
+			"configured vpcCidr (%s) does not match actual existing vpc cidr (%s)",
+			c.VPCCIDR,
+			*existingVPC.CidrBlock,
+		)
 	}
 
 	describeSubnetsInput := ec2.DescribeSubnetsInput{
@@ -138,7 +143,6 @@ func (c *Cluster) Create(stackBody string) error {
 	}
 
 	ec2Svc := ec2.New(c.session)
-
 	if err := c.validateKeyPair(ec2Svc); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


mumoshu:
@cgag I'm impressed with the informative error output.
I believe it will keep saving time for the rest of our lives 👍

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -337,3 +352,25 @@ func (c *Cluster) validateDNSConfig(r53 r53Service) error {
...
 	return nil
 }
+
+func stackEventErrMsgs(events []*cloudformation.StackEvent) []string {
+	var errMsgs []string
+
+	for _, event := range events {
+		if aws.StringValue(event.ResourceStatus) == cloudformation.ResourceStatusCreateFailed {
+			// Only show actual failures, not cancelled dependent resources.
+			if aws.StringValue(event.ResourceStatusReason) != "Resource creation cancelled" {
```
aaronlevy: Why suppress cancellation errors?

aaronlevy:
One question, but otherwise LGTM

multi-node/aws/pkg/cluster/cluster.go
cgag: A failure to create a resource causes cancellation failures for the resources that depend on it, which don't help with debugging. I wanted to only show root causes.

If you or @colhom or @mumoshu think showing the cancellations is worth while, I can remove that check.

aaronlevy: That makes sense. I'm fine with ignoring them. And if a user triggers a cancellation, they should likely know why it failed :)

colhom: Agreed on that. I can't say I've ever tripped on that one.

colhom:
@cgag rebase please.

cgag:
rebased

colhom:
on behalf of first-time users of kube-aws, thank you @cgag ;)

PR #455, opened 2016-05-03
kube-aws: fix shellcheck warnings in build script

cgag:
https://github.com/koalaman/shellcheck

Quote variales to prevent word splitting/globbing.
Prefer $() to ``

Enable "strict mode": http://redsymbol.net/articles/unofficial-bash-strict-mode/


aaronlevy:
LGTM

PR #456, opened 2016-05-03
kube-aws: add script to run tests

cgag:
easier than always using ctrl+r


aaronlevy:
LGTM. One nit would be to update the travis tests to use this too (unify how we run tests in case scope of this changes).

PR #457, opened 2016-05-03
kube-aws: golint fixes

cgag:
Error messages shouldn't end with punctuation.
Comment formatting for exported symbols.
cluster.ClusterInfo is redundant, rename to cluster.Info.


aaronlevy:
LGTM

Issue #447, opened 2016-04-30
install-kube-system is not robust to some failures


daveey:
It looks like /opt/bin/install-kube-system got a

```
Apr 30 03:30:01 ip-10-0-5-235.ec2.internal install-kube-system[4659]: {"ErrStatus":{"metadata":{},"status":"Failure","message":"The  operation against  could not be completed at this time, please try again.","reason":"ServerTimeout","details":{},"code":500}}
```

which curl considers a success. so the cluster doesn't end up running kube-dns

colhom:
@daveey this curl nonsense can be replaced by `kubectl` via the hyperkube container image. That will allow us to analyze exit codes meaningfully and use higher-level constructs like `kubectl apply` for this.

@daveey are you interested in doing a PR for this?

daveey:
@colhom it actually looks like curl just needs a `-f` flag to treat a 500 as a failure. i'll submit a PR for that

daveey:
https://github.com/coreos/coreos-kubernetes/pull/450

colhom:
@daveey unfortunately it's not that easy. When the controller reboots, for instance, `curl -f` will return non-0 if we try to create resources that already exist, and `install-kube-system.service` will keep trying to create stuff that's already there (forever in this case).

I want to use `kubectl apply`, as that will do exactly what we want.

daveey:
ah, that makes sense. why can't these go into /etc/kubernetes/manifests/

colhom:
We don't want to them to be static pods.

aaronlevy:
I'd rather not use apply on each reboot. You could stomp on user-provided changes. That being said, there's lots of room to make this smarter...

colhom:
As to making this smarter, agree completely. We should start by using kubectl for this instead of curl.

As far as user-supplied changes- the mantra is that the cluster assets on disk are the source of truth. Given that we don't have a supported way to change cluster assets and apply them to a running cluster, that's kind of a moot point at present.

Once we can support rolling asset changes out to an existing cluster, I feel that nuking any user-supplied changes on reboot that aren't done via asset files becomes a "feature".

daveey:
I tried using kubectl apply, but ran into https://github.com/kubernetes/kubernetes/issues/24088

daveey:
[...]

PR #462, opened 2016-05-04
Use kubectl instead of curl to load kube-system manifests

daveey:
curl can silently fail, leaving the kube system only partially set up

ideally we would use the quay.io/coreos/hyperkube:v1.2.2_coreos.0 image and run /hyperkube kubectl, however this doesn't work due to

https://github.com/kubernetes/kubernetes/issues/24088

this PR uses a 3rd party image for kubectl, which means it shouldn't be merged. 

i am hoping someone can help me add kubectl to the hyperkube image directly, or suggest other alternatives

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -81,7 +81,7 @@ coreos:
         Type=simple
         StartLimitInterval=0
         Restart=on-failure
-        ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
+        ExecStartPre=/usr/bin/curl -f http://127.0.0.1:8080/version
         ExecStart=/opt/bin/install-kube-system
 
 write_files:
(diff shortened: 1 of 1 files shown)


mumoshu:
@daveey Great PR!

I had also needed to depend on a 3rd party image in my PR #465 

What I have done there was to make image/command passed to `docker run` fully configurable via `cluster.yaml` and noted that it may change in the future. See https://github.com/coreos/coreos-kubernetes/pull/465/files#diff-071e0ee4ae8c6adf4865f25d851f30c9R74 for code regarding it.

With that, I believe that there will be small chances we can get our PRs merged as is, because we won't break things in fragile/backward-incompatible ways.

If that is not enough, I guess we can contribute a Dockerfile for `kubectl` and ask CoreOS guys if they could docker build/push it in a Docker repo something like `quay.io/coreos/kubectl`.

@colhom @aaronlevy Excuse me for being obtrusive but would you mind looking into this? We like to contribute but the same issue seems to be blocking us.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -90,21 +90,17 @@ write_files:
     owner: root:root
     content: |
       #!/bin/bash -e
-      /usr/bin/curl -H "Content-Type: application/json" -XPOST -d @"/srv/kubernetes/manifests/kube-system.json" "http://127.0.0.1:8080/api/v1/namespaces"
+      docker run -v /etc/kubernetes:/etc/kubernetes \
```
aaronlevy: I'd probably prefer this was run via `rkt`

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -90,21 +90,17 @@ write_files:
...
-      for manifest in {kube-dns,heapster}-svc.json;do
-          /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
-          -d @"/srv/kubernetes/manifests/$manifest" \
-          "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
-      done
+      docker run -v /etc/kubernetes:/etc/kubernetes \
+                 -v /srv/kubernetes:/srv/kubernetes \
+                 --net host \
+                 --rm lachlanevenson/k8s-kubectl \
+                 apply -f /srv/kubernetes/manifests/
```
aaronlevy: I don't think we can safely use apply, as this is run during every boot. The api objects could have changed during their lifecycle and it's not safe for us to assume we can stomp on those changes.

aaronlevy: Also, I still think we need to call out each object by name. Otherwise each boot could create objects which happened to be stored in `/srv/kubernetes/manifests` and that may not be expected by end-users.

aaronlevy:
Couple things to cover here:

- As far as the underlying issue, can you give more detail on how curl is silently failing in a way that kubectl would surface more info? If curl fails, I would expect the apiserver response to be in the journal output for the unit - if not we can likely resolve this.

- As far as kubectl in the hyperkube container, I would prefer that the flag issue be fixed upstream, rather than trying to work around the bug by including a separate binary or needing to package a separate image. We try to keep our images as close to upstream as possible (and others in the community could also benefit from this fix).

Once the flag bug is resolved, I wouldn't necessarily be opposed to running kubectl via a container (although I'd likely prefer that it was run via `rkt`). But it would be slightly slower on the installation process (vs curl).

daveey:
@aaronlevy thanks for the review, I should have posted a link to the issue. 

https://github.com/coreos/coreos-kubernetes/issues/447

Take a look at the discussion there which addresses the "apply" issue, as well as why curl is problematic.

I agree with the preference for the original hyperkube image, however the flag bug has been open for a while and not addressed so I don't know when we can expect it.

aaronlevy:
Commented on the linked issue.

Would you be interested in taking a stab at fixing the upstream issue? On a quick scan it looks like it would mean improving flag parsing here: https://github.com/kubernetes/kubernetes/blob/master/cmd/hyperkube/hyperkube.go#L117

daveey:
@aaronlevy i won't have time to work on the upstream issue for a while

colhom:
@daveey I'll be looking into the upstream hyperkube issue soon to see if we can get this resolved.

colhom:
@daveey update on things. kubernetes/kubernetes#25512 is merged, fixing the hyperkube flag issue.

we're going to wait until v1.3 (with that PR in it) is released before going forward with this PR.

aaronlevy:
@daveey would this be something you're still interested in tackling? kubectl is now part of hyperkube, so we could now do something like `rkt run quay.io/coreos/hyperkube --exec /hyperkube -- kubectl create -f foo.yaml`

aaronlevy:
I'm going to close this due to inactivity - but please re-open if this is something you would like to see merged.

PR #463, opened 2016-05-04
Add flag to point Kubernetes volume_plugin_dir to a subdir mounted in…

barakmich:
…to kubelet_wrapper. 

Since `/usr/libexec/...` is read-only on CoreOS, the default for this flag is completely useless. Let's make it useful.


colhom:
@barakmich could you do  the same for `mutli-node/aws/pkg/config/templates/user-data-{controller,worker}`?

aaronlevy:
@barakmich what is status of this? Do you still want these flags?

aaronlevy:
Closing for now, but @barakmich let me know if you want it re-opened

PR #465, opened 2016-05-05
kube-aws: Drain nodes before shutting them down

### LINKED ISSUE #340: Production Quality Deployment
The goal is to offer a "production ready solution" for provisioning a coreos kubernetes cluster. These are the major functionality blockers that I can think of.
- [ ] Cluster upgrade path
  - [ ] Enable decommissioning of kubelets when instances are rotated out of the ASG
  - [ ] Automatically remove nodes when instances are rotated out of ASG
- [ ] Put controllers into ASG, behind an ELB
- [x] Spread workers across AZs (federation-lite) -- thanks @mumoshu! ref https://github.com/coreos/coreos-kubernetes/pull/439
- [ ] Dedicated etcd cluster in an ASG, behind an ELB
  - [ ] Set up etcd tls
- [ ] Set up ~~controller~~ and worker AutoscalingGroups to recover from ec2 instance failures
- [ ] Secure etcd peer/client connections with TLS
- [x] Route53 integration for APIServerEndpoint. Automatically create hosted-zone and/or A record for controller EIP on `kube-aws up` -- DONE #389 (requires that the hosted zone already exist)
- [ ] Provision AWS ElasticSearch cluster
  - [ ] Kibana/elasticsearch/fluentd addons. (ELK logging)
  - [ ] Enable heapster elasticsearch sink functionality (https://github.com/kubernetes/heapster/pull/733)
- [x] Support deploying to existing VPC (and maybe existing subnet as well?) -- DONE #346
- [ ] Cluster PKI infrastructure (ref #420)
  - [ ] Kubelet TLS bootstrapping [upstream proposal](https://github.com/kubernetes/kubernetes/pull/20439)
  - [ ] Figure out what we're going to do about automated CSR signing in kube-aws (necessary for self-healing and autoscaling)
  - [x] Provide option to use pre-existing CA certificate and key to sign component certs (integrate with existing PKI systems)

mumoshu:
Drain nodes before shutting them down to give running pods time to gracefully stop.

This change basically achieves it by running `docker run IMAGE kubectl drain THE_NODE --force` on the to-be-shut-down node before the kubelet gets SIGTERM in a CoreOS' shutdown process.

Depends #608
Refs #340

Without this change, kubelets getting SIGTERM without a prior `drain` results in unfunctional(actually not schedulable) pods to have statuses `Ready`.

With this change, when an ASG's desired cap decreased, a node's status changes over time as follows:

On desired cap change:
=> STATUS=Ready

On shut-down started:
=> STATUS=Ready,SchedulingDisabled (<- Pods are stopped and status is changed by `kubectl drain`)

On shut-down finished:
=> Status=NotReady,SchedulingDisabled (<- It's `NotReady` but it won't result in a down-time because we already stopped both further scheduling and pods)

After a minute:
=> The node disappears from the output of `kubectl get nodes`

Note that:
- This applies to manual shutdowns(via running `sudo systemctl shutdown` for example) and automated shutdowns(triggered by AWS AutoScaling when nodes get rotated out of a group)
- We currently depend on the community docker image `mumoshu/kubectl` because `kubectl` included in the official `coreos/hyperkube` image doesn't work due to the issue https://github.com/kubernetes/kubernetes/issues/24088 in Kubernetes. Once the issue is fixed and the CoreOS team published the new hyperkube image with the updated Kubernetes, we can remove that dependency.
- I consider this an experimental feature. So you shouldn't expect configuration API regarding this stable. It may change in the future.

### DIFF
--- a/multi-node/aws/pkg/config/config_test.go
+++ b/multi-node/aws/pkg/config/config_test.go
@@ -307,7 +307,7 @@ func TestMultipleSubnets(t *testing.T) {
 
 	validConfigs := []struct {
 		conf    string
-		subnets []Subnet
+		subnets []*Subnet
 	}{
 		{
 			conf: `
@@ -320,7 +320,7 @@ subnets:
   - availabilityZone: ap-northeast-1c
     instanceCIDR: 10.4.4.0/24
 `,
-			subnets: []Subnet{
+			subnets: []*Subnet{
 				{
 					InstanceCIDR:     "10.4.3.0/24",
 					AvailabilityZone: "ap-northeast-1a",
@@ -339,7 +339,7 @@ controllerIP: 10.4.3.50
 availabilityZone: ap-northeast-1a
 instanceCIDR: 10.4.3.0/24
 `,
-			subnets: []Subnet{
+			subnets: []*Subnet{
 				{
 					AvailabilityZone: "ap-northeast-1a",
 					InstanceCIDR:     "10.4.3.0/24",
@@ -355,7 +355,7 @@ availabilityZone: ap-northeast-1a
 instanceCIDR: 10.4.3.0/24
 subnets: []
 `,
-			subnets: []Subnet{
+			subnets: []*Subnet{
 				{
 					AvailabilityZone: "ap-northeast-1a",
 					InstanceCIDR:     "10.4.3.0/24",
@@ -368,7 +368,7 @@ subnets: []
 availabilityZone: "ap-northeast-1a"
 subnets: []
 `,
-			subnets: []Subnet{
+			subnets: []*Subnet{
 				{
 					AvailabilityZone: "ap-northeast-1a",
 					InstanceCIDR:     "10.0.0.0/24",
@@ -380,7 +380,7 @@ subnets: []
 # Missing subnets field fall-backs to the single subnet with the default az/cidr.
 availabilityZone: "ap-northeast-1a"
 `,
-			subnets: []Subnet{
+			subnets: []*Subnet{
 				{
 					AvailabilityZone: "ap-northeast-1a",
 					InstanceCIDR:     "10.0.0.0/24",
--- a/multi-node/aws/pkg/config/config_test.go
+++ b/multi-node/aws/pkg/config/config_test.go
@@ -456,7 +456,7 @@ subnets:
 		confBody := minimalConfigYaml + conf
 		_, err := ClusterFromBytes([]byte(confBody))
 		if err == nil {
-			t.Errorf("expected error parsing invalid config: %s", confBody)
+			t.Errorf("expected error parsing invalid config:\n%s", confBody)
 		}
 	}
 
--- a/multi-node/aws/pkg/config/config_test.go
+++ b/multi-node/aws/pkg/config/config_test.go
@@ -1,9 +1,13 @@
 package config
 
 import (
+	"bytes"
+	"gopkg.in/yaml.v2"
 	"net"
 	"reflect"
+	"strings"
 	"testing"
+	"text/template"
 )
 
 const minimalConfigYaml = `externalDNSName: test.staging.core-os.net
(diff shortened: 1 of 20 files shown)


colhom:
@mumoshu this looks awesome, very exciting!

I don't know that we need the `experimental` bit. I would say that this behavior is "universally desirable", so we may just want to run it through conformance tests and node-draining on shutdown non-configurable. If for some odd reason (testing, etc) somebody wants to disable this, that should be done at a lower level (disabling the service in cloud-config, etc)

\cc @aaronlevy @cgag

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -59,6 +59,29 @@ coreos:
...
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/bin/true
+        ExecStop=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.Experimental.NodeDrainer.KubectlImage}} \
+          {{.Experimental.NodeDrainer.KubectlCommand}} \
+          --server=https://{{.ControllerIP}}:443 \
+          --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
+          drain $$(hostname) \
+          --force'
+
```
cgag: Does this also need --ignore-daemonsets=true?  The man page implies to me that you'd need both if you had a daemonset and an unmanaged pod, even though the name "force" makes me think "force" would be enough.

aaronlevy:
@mumoshu can you provide a little more detail regarding: 

> Without this change, kubelets getting SIGTERM without a prior drain results in unfunctional(actually not schedulable) pods to have statuses Ready

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -59,6 +59,29 @@ coreos:
         [Install]
         RequiredBy=kubelet.service
 
+    {{ if .Experimental.NodeDrainer.Enabled }}
+    - name: kube-node-drainer.service
+      enable: true
+      content: |
+        [Unit]
+        Description=drain this k8s node to make running pods time to gracefully shut down before stopping kubelet
```
aaronlevy: I'm not seeing how this unit keeps the kubelet running until drained.

cgag: Looks like systemd shutdown down services in the reverse of the boot order, so `After=kubelet.service` should mean this should finish shut down before the kubelet is.

aaronlevy: ah, thanks @cgag

colhom:
To recap in-person discussions between @aaronlevy, @philips and myself:

Summary- we will leave this node-draining behavior gated on experimental now as @mumoshu has it in this PR, defaulting to false.

Node draining is certainly a good thing to do if you're taking a node out of commission forever. If you're just rebooting the machine, it's not clear if draining it is the desirable thing to do. Factors in this decision include number of pods on that node, how long it takes to migrate them, and whether or not those pods have any affitinty for resources on that host. (An example discussed would be a local-cache implemented via host mount).

 In the future:

* [Taints and tolerations](https://github.com/kubernetes/kubernetes/blob/master/docs/design/taint-toleration-dedicated.md) could provide a means to express which pods/nodes will "be drained on shutdown", and which ones won't.

* Integrate node-draining more with events more specific than "we're shutting down." The most salient additional detail there is "is this node coming back soon, or is it gone?".

@mumoshu: could we get a rebase on this PR please ;)

mumoshu:
@colhom @aaronlevy @cgag @philips Thanks for your review 🙇
I have just rebased this.

Regarding the experimental bit, I agree with leaving it as is.

The context is that, I believe my tests around this functionality does not cover edge cases I can't even imagine of, hence experimental. As I'm planning to deploy CoreOS+K8s onto my production environment, I wished I could end up with a battle-tested solution which we can have some confidence marking it non-experimental. Googling the whole internet and consulting myself does not reveal something like that for me though 😃 

The example of a local-cache implemented via host mount you've mentioned is very insightful for me. I personally think of it as one of those (not even edge?) cases.

Also, the integration based on taints-and-tolerations and more fine-grained events seems very interesting/promising to me 👍

mumoshu:
@cgag Thanks for the good catch https://github.com/coreos/coreos-kubernetes/pull/465#discussion_r62251019!

I have confirmed that it was actually failing when draining a node running DaemonSet-managed pods and fixed it in ae43f64

mumoshu:
https://github.com/coreos/coreos-kubernetes/pull/465#issuecomment-217290868

@aaronlevy Sure. I had wished if a kubelet could stop gracefully when in shutting-down process CoreOS tries to stop it(by sending SIGTERM to it via systemd via docker). So I have tested it. My test showed that kubelet doesn't gracefully stop while CoreOS is shutting down.

I have confirmed it by repeating `kubectl get po,no` and looking into its output after I had triggered shutdown on a node.
After the node is terminated, the node has shown as the status `Ready` in `kubectl getpo,no` output which indicates it is recognized as schedulable in k8s but actually it isn't (as the node is already terminated).
After 1 or 2 minutes, it went `NotReady` and finally disappeared from the output.

multi-node/aws/pkg/config/config_test.go
```diff
@@ -309,3 +314,165 @@ releaseChannel: non-existant #this release channel will never exist
...
+		conf        string
+		nodeDrainer NodeDrainer
+	}{
+		{
+			conf: `
+`,
+			nodeDrainer: NodeDrainer{
+				Enabled:        false,
+				KubectlImage:   "mumoshu/kubectl:v1.2.2",
+				KubectlCommand: "kubectl",
```
colhom: When the default Kubectlmage and KubectlCommand values are changed when the `/hyperkube kubectl` issue is fixed, we'll have to edit tests like to reflect the changes to defaults. 

I would like to see constants `defaultHyperkubeImage` and `defaultKubectlCommand`:
* defined in `config.go`
* applied as defaults to the Cluster struct [here](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/config.go#L30)
* used in the config tests like this which infer the default value

colhom:
@mumoshu I talked with @aaronlevy and we have decided to take a stab a the upstream issue hyperkube issue (kubernetes/kubernetes#24088) that's blocking this PR. Give us a day or two to look into this.

multi-node/aws/pkg/config/config_test.go
mumoshu: Sounds nice! Just added 3171bd6 for it 👍

philips:
There is also this issue which takes a different approach for CoreOS autoupdates. They are somewhat orthogonal but highly related: https://github.com/coreos/bugs/issues/1274

colhom:
kubernetes/kubernetes#24088 has been merged. @aaronlevy could we cherry-pick this into the next hyperkube image?

aaronlevy:
@mumoshu what you describe is more or less the expected behavior. The actual window for marking the node as unhealthy / pod eviction can be set via flags on the controller-manager (e.g. `PodEvictionTimeout, NodeMonitorGracePeriod, NodeStartupGracePeriod, NodeMonitorPeriod`)

@colhom are you sure that's the right issue?

colhom:
Ooops I meant kubernetes/kubernetes#25512. 

@mumoshu we're going to wait until v1.3 (with that PR in it) is released before going forward wit this PR.

mumoshu:
@colhom I'm happy to work on it now as you've merged the 1.3-updates branch into master.

mumoshu:
@aaronlevy Thanks for confirming!

Yes, it might be an expected behavior.
The point is that I want to avoid generating errors/delays whenever possible.

As it's been a quite long time since we've talked, let me recap for clarification.

I believe a node shut-down in (1)CoreOS's auto updating process (2)an ASG scale-in (3)a manual maintenance (4)an EC2 failure or etc will result in some down-time/decreased functional pods until Kubernetes marks the node not-ready/unschedulable. Time to not-ready/unschedulable seems to be configurable via flags you've mentioned.

What I wanted to do here is, mark the node before kubernetes does, to completely avoid errors/delays in this case, so that kubernetes can react to decreased pods immediately.

Also, even with changes in this PR, in some cases I believe you'll need additional steps to make a node actually out of service without returning errors to customers. E.g. when the node running pods or services exposing your web service through node/host ports for external load balancing. I guess the whole procedure would be as follows:

* Remove the node from ELB (done by k8s? or outbound according to your setup. If you've set up your ASG to manage ELB, and the node is going down in process of auto-scaling, it's done by ASG implicitly. But in other cases?)
* Then `kubectl drain` the node
* Do some maintenance work, including a shut-down.
* If the node comes up, add it to ELB(done by k8s if you've made k8s to manage it via k8s services(type: LoadBalancer)? I haven't tried this case yet)

It is not addressed in this PR and I don't even know how I can do the whole thing cleanly enough.

colhom:
@mumoshu any interest in resurrecting this from the dead?

`hyperkube kubectl` is fully functional and we now have cluster upgrades coming down the pipeline in #608. Wouldn't be a graceful upgrade without some node draining!

mumoshu:
@colhom Definitely yes! Can I just rebase this branch onto the latest `master` (with of course some manual testing)?

colhom:
@mumoshu could you rebase on `colhom:cluster-upgrades` and list #608 as a dependency?

mumoshu:
@colhom sure 👍

mumoshu:
@colhom Rebased and fixed a failing test!

felixbuenemann:
Does this really depend on #608? I think this would be a great feature to have with spot instances, which can go away due to price fluctuations, but prices usually don't fluctuate in all AZs at once, so there are usually healthy nodes on other AZs to quickly take over the load.

aaronlevy:
kube-aws development has moved to its own top-level repository. If this is still something you would like to be merged, please re-open this PR under the new repository at: https://github.com/coreos/kube-aws

PR #470, opened 2016-05-10
docs: add link to generic userdata scripts

robszumski:


aaronlevy:
I think we need to add some additional info about setting up TLS assets (these script expect them to be in certain locations and we don't say anything about it).

robszumski:
Maybe the scope of this needs to expanded into a new guide that is basically the opposite of the manual instructions.

1. Do this for TLS (link to existing guide) and put them at X location
2. Use these cloud configs, deploy any number of controllers and workers
3. Configure kubectl (link to existing guide)

robszumski:
I updated this link to the new guide that was written. I think all issues with this PR are now solved.

PR #477, opened 2016-05-12
lib: Add kubernetes.default.svc as alt_name in openssl config

chancez:
Currently if a pod attempts to connect to kubernetes.default.svc using the ca
provided as a service account it will fail due to that name not being trusted

### DIFF
--- a/lib/init-ssl
+++ b/lib/init-ssl
@@ -45,6 +45,8 @@ subjectAltName = @alt_names
 [alt_names]
 DNS.1 = kubernetes
 DNS.2 = kubernetes.default
+DNS.3 = kubernetes.default.svc
+DNS.4 = kubernetes.default.svc.cluster.local
 "
 echo "Generating SSL artifacts in $OUTDIR"
 


aaronlevy:
We should probably add `kubernetes.default.svc.cluster.local` to that list as well. Technically it is a configurable, but we assume `cluster.local` in pretty much everything in this repo (I believe, might be worth a grep).

chancez:
@aaronlevy ah yes, i agree. What about `kubernetes.default.svc.cluster` ? That should work too I'd guess.

chancez:
Would this suffice?
```
[alt_names]
DNS.1 = kubernetes
DNS.2 = kubernetes.default
DNS.3 = kubernetes.default.svc
DNS.4 = kubernetes.default.svc.cluster
DNS.5 = kubernetes.default.svc.cluster.local
```

chancez:
@aaronlevy All updated

aaronlevy:
lgtm

PR #487, opened 2016-05-17
*: Remove unused K8S_SERVICE_IP from scripts

dghubble:
`K8S_SERVICE_IP` does not seem to be used within the scripts, vagrant examples work fine without it.

* Last use seems to be https://github.com/coreos/coreos-kubernetes/commit/c6ef8ddeb5dae1c2ab778350783f443a395eebba#diff-349be742aedc3ca1d5af305a74e1c4bcL63
* Removed in coreos-baremetal https://github.com/coreos/coreos-baremetal/pull/150

cc @aaronlevy

### DIFF
--- a/multi-node/generic/controller-install.sh
+++ b/multi-node/generic/controller-install.sh
@@ -21,10 +21,6 @@ export POD_NETWORK=10.2.0.0/16
 # Routing to these IPs is handled by a proxy service local to each node, and are not required to be routable between nodes.
 export SERVICE_IP_RANGE=10.3.0.0/24
 
-# The IP address of the Kubernetes API Service
-# If the SERVICE_IP_RANGE is changed above, this must be set to the first IP in that range.
-export K8S_SERVICE_IP=10.3.0.1
-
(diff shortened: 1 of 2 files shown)


aaronlevy:
IIRC this was used for cert generation at some point (because I was too lazy to figure out the first IP in a cidr range in bash).

Just checked the single node install and it looks like it's defined both in the Vagrant file and in the user-data (which isn't ideal given the install would fail if the service-ip-range is changed).

We can remove it for now, but might need to add it back to properly support ssl-generation step (and I want to add this back to the docs for multinode/generic as the script expect certs to be available).

aaronlevy:
Ideally we could close the configuration bug that allows for a broken installation (change service-ip-range, but Vagrant file always uses 10.3.0.1).

This might mean making the user-data script have a method of exporting it's configurables to external tools.

dghubble:
Up to you. Its just a few lines I ported and later removed so thought I'd offer it as a cleanup.

PR #491, opened 2016-05-18
`stack-template.json`: use spaces only for indenting

flenter:
The whole file was using spaces for indentation except for a few
lines where tabs were used. Using spaces here keeps things also
visually consistent when a tab is not displayed as 8 spaces


aaronlevy:
lgtm

PR #500, opened 2016-05-22
kube-aws: Fix two bugs with subnets & existing VPCs

michaeldiscala:
Bug #1: Before this patch, when specifying a route table ID in the
cluster.yml file, the substitution on line 618 fails because the
RouteTableID is not available in that scope. This patch modifies the
substitution to read from the correct scope.

Bug #2: Before this patch, the validation code that is run during
'kube-aws up' expects ``c.InstanceCIDR`` to be a valid CIDR even when
the Subnets parameter is set instead. This patch modifies the validation
to avoid raising errors in this case.

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -565,7 +565,7 @@ func (c *Cluster) ValidateExistingVPC(existingVPCCIDR string, existingSubnetCIDR
 		}
 	}
 	_, instanceNet, err := net.ParseCIDR(c.InstanceCIDR)
-	if err != nil {
+	if len(c.Subnets) == 0 && err != nil {
 		return fmt.Errorf("error parsing instances cidr %s : %v", c.InstanceCIDR, err)
 	}
 	_, vpcNet, err := net.ParseCIDR(c.VPCCIDR)
--- a/multi-node/aws/pkg/config/templates/stack-template.json
+++ b/multi-node/aws/pkg/config/templates/stack-template.json
@@ -615,7 +615,7 @@
     ,
     "{{$subnetLogicalName}}RouteTableAssociation": {
       "Properties": {
-        "RouteTableId": "{{.RouteTableID}}",
+        "RouteTableId": "{{$.RouteTableID}}",
         "SubnetId": {
           "Ref": "{{$subnetLogicalName}}"
         }


michaeldiscala:
I'm not at all sure that this is the best way to solve these issues, but they worked for my local testing and I figured I'd pass them along. In particular, I think the check that I'm modifying in  multi-node/aws/pkg/config/config.go may actually be redundant and can just be deleted.

mumoshu:
As the author of the PR and commits introduced these bugs, LGTM, and thanks @michaeldiscala !
Personally, I'd like these cases to be tested in `TestMultipleSubnets` in `config_test.go`.

Forwarding further review to @colhom as I'm not an owner of this repository 🙇

multi-node/aws/pkg/config/templates/stack-template.json
```diff
@@ -615,7 +615,7 @@
     ,
     "{{$subnetLogicalName}}RouteTableAssociation": {
       "Properties": {
-        "RouteTableId": "{{.RouteTableID}}",
+        "RouteTableId": "{{$.RouteTableID}}",
```
colhom: This is inside a `with` block, and consequently needs to use `$.` to access the global scope. Only arises when `routeTableId` cluster.yaml option is set, which we do not test.

multi-node/aws/pkg/config/config.go
```diff
@@ -565,7 +565,7 @@ func (c *Cluster) ValidateExistingVPC(existingVPCCIDR string, existingSubnetCIDR
 		}
 	}
 	_, instanceNet, err := net.ParseCIDR(c.InstanceCIDR)
-	if err != nil {
+	if len(c.Subnets) == 0 && err != nil {
```
colhom: This arises when `subnets` and `existingVpc` fields are both properly defined, which we do not test.

colhom:
@michaeldiscala could you back out your fix for bug #2 (subnets/existingVPC). The first patch is LGTM, the second is unfortunately going to require some additional finagling to validate all subnets against the VPC instead of just one.

colhom:
@michaeldiscala thanks for reporting not one, but two show-stopping bugs! I've got your first fix up in #514, so I'm going to close this one out.

PR #507, opened 2016-05-25
Fix vpc validation for multiple subnets

colhom:
I have a kind-of-off-topic commit which introduces omitempty tags for the Cluster yaml fields. This is subtly desirable behavior I would think, in that defining an empty value in `cluster.yaml` should probably imply the preset default rather than empty.

Fixes bug 2 in #500

\cc @aaronlevy @michaeldiscala

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster_test.go
+++ b/multi-node/aws/pkg/cluster/cluster_test.go
@@ -11,16 +11,44 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/config"
+	yaml "gopkg.in/yaml.v2"
 )
 
-const minimalConfigYaml = `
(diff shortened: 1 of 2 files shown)


multi-node/aws/pkg/cluster/cluster_test.go
```diff
@@ -11,16 +11,44 @@ import (
...
+	if len(c.Subnets) > 0 {
+		for i := range c.Subnets {
+			c.Subnets[i].AvailabilityZone = fmt.Sprintf("dummy-az-%d", i)
+		}
+	} else {
+		//Legacy behavior
+		c.AvailabilityZone = "dummy-az-0"
+	}
+
+	out, err := yaml.Marshal(&c)
```
colhom: This is why 5341bf33e13e589399444846506d431c5894e512 exists. We're using the `config.Cluster` to automate pre-processing of test yaml data.

aaronlevy:
Code lgtm

Are you keeping track / setting target versions for deprecation of "legacy" features?

PR #508, opened 2016-05-25
DO NOT MERGE - work where Fabian can see it

joeatwork:
Here are a preliminary set of manifests I've been using to experiment with Prometheus and Grafana


aaronlevy:
If there is no intention of merging this and this is just scratch space, would you mind keeping this in your fork? If there is intention of getting this merged eventually - it's fine to leave here.

joeatwork:
Gotcha - getting rid of it. Thanks!

Issue #497, opened 2016-05-19
Remove SecurityContextDeny admission controller


aaronlevy:
See: http://kubernetes.io/docs/admin/admission-controllers

Recommended set is now:
`--admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota`

PR #513, opened 2016-05-26
*: Update apiserver admission-control

dghubble:
* Update to the k8s 1.2.0 recommended admission-control settings http://kubernetes.io/docs/admin/admission-controllers/
* Remove SecurityContextDeny. This resolved "ConfigMap [It] should be consumable from pods in volume..." conformance failures ([on bare metal](https://github.com/coreos/coreos-baremetal/issues/71#issuecomment-221932670))
* Closes coreos/coreos-kubernetes#497

@colhom @aaronlevy


aaronlevy:
LGTM

PR #514, opened 2016-05-26
stack-template: global scope for subnet routetableid

colhom:
thanks @michaeldiscala for the bug report and fix.

\cc @cgag 

ref #500


cgag:
LGTM

PR #519, opened 2016-05-31
docs: fix up inconsistent variables

robszumski:
We haven't used the `${CONTROLER_IP}` anywhere else, and it wasn't introduced on the first page. Switch it to `${MASTER_HOST}` for consistency.


joshix:
lgtm

colhom:
@robszumski note on this- internally all of our code still refers to controller instead of master.

joshix:
Right on, @colhom -- and eventually we want to say "controller" everywhere. @robszumski was quick-fixing this so it was at least internally consistent in the meantime.

PR #522, opened 2016-06-01
kube-aws: add maxWorkerCount param

kalbasit:
I had an issue updating my cluster: `MinInstancesInService must be less than the autoscaling group's MaxSize`. This PR allows me to customize the MaxSize to avoid this issue.

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -41,6 +41,7 @@ func newDefaultCluster() *Cluster {
 		ControllerInstanceType:   "m3.medium",
 		ControllerRootVolumeSize: 30,
 		WorkerCount:              1,
+		MaxWorkerCount:           1,
 		WorkerInstanceType:       "m3.medium",
 		WorkerRootVolumeSize:     30,
 		CreateRecordSet:          false,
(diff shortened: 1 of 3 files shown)


cgag:
I an into this issue before and was considering making MaxSize always be WorkerCount + 1 so that update will always work.  Currently batchsize is hardcoded to 1, so I believe there's no reason for MaxSize to every be more than WorkerCount +1.  I like this because I want to keep cluster.yaml as simple as possible.

However, this could be a problem if you had a large cluster to update and wanted a larger batchsize, so maybe it should be configurable.  Maybe BatchSize should be the configurable, and MaxSize would be `WorkerCount + BatchSize`?  Maybe if this is a problem for you, you need to customize the generated stack-template.json?

Any thoughts on the experience here @kalbasit, @danielfm, @colhom?

danielfm:
@cgag

> Currently batchsize is hardcoded to 1, so I believe there's no reason for MaxSize to every be more than WorkerCount +1. I like this because I want to keep cluster.yaml as simple as possible.

I like that idea too.

> However, this could be a problem if you had a large cluster to update and wanted a larger batchsize, so maybe it should be configurable.

Having a configurable `BatchSize` (so that people could potentially specify an arbitrary number there) might cause problems for some users depending on the current cluster saturation: if the cluster is heavily saturated, taking out more than one instance at a time might increase the odds of some pods not being able to be scheduled, thus causing service disruption. Otherwise, if the cluster operates with enough spare resources, this might make stack updates much faster without service disruption. It's difficult to tell.

Since the cluster saturation is something that changes over time, a `BatchSize` chosen upfront might cause more problems than it might solve.

aaronlevy:
ping @colhom @cgag -- any updates here? Is this something we want to land?

colhom:
@aaronlevy this was rectified in #608 as part of making the stack updatable.

https://github.com/coreos/coreos-kubernetes/pull/608/commits/a3c772a5308fe39bbcf5b6539ee9d01fbb1acbab#diff-e5ad8b828a5de3fe2229bee5c64ab2a0R170

\cc @kalbasit @danielfm

kalbasit:
@colhom works for me. Please add to your PR `closes #522` so I know when it's ready :) Thanks.

colhom:
@kalbasit done.

aaronlevy:
I'm going to close this so we're not tracking an issue in a PR if it's decided this won't be merged. @colhom or @kalbasit feel free to open an issue to track the particular functionality needed / issue being resolved.

PR #523, opened 2016-06-02
config: clean up ami region/channel resolution code

colhom:
was needlessly complex due to oversight in the previous refactor.

also removed redundant and unecessary state about valid regions and supported channels. this information is now provided canonically by the coreos release ami metadata and the config package respectively.

\cc @cgag

### DIFF
--- a/multi-node/aws/pkg/config/region_map.go
+++ b/multi-node/aws/pkg/config/region_map.go
@@ -6,57 +6,22 @@ import (
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/coreosutil"
 )
 
-var regions = []string{
-	"ap-northeast-1",
-	"ap-southeast-1",
-	"ap-southeast-2",
-	"eu-central-1",
-	"eu-west-1",
-	"sa-east-1",
-	"us-east-1",
-	"us-gov-west-1",
-	"us-west-1",
-	"us-west-2",
-}
-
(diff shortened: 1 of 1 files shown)


multi-node/aws/pkg/config/region_map.go
```diff
@@ -6,57 +6,22 @@ import (
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/coreosutil"
 )
```
cgag: Maybe we should just move this function into config.go?  Not sure naming this "region_map.go" makes sense anymore.

cgag: Maybe we should just move this function into config.go? Not sure naming this "region_map.go" makes sense anymore.

cgag:
LGTM, much easier to follow.

Some complaints outside the scope of this PR: 
 
I don't like how stringly typed all the AMI stuff is.  GetAMIData returning `map[string]map[string]string` is nasty, but sort of forced by the bad json structure returned from `https://coreos.com/dist/aws/aws-alpha.json` (the AMIs by region should be in an array).   

If we could change that in the future, and have GetAMIData return a slice of structs, and then get getAMI return a struct rather than a string, that'd be great. 

getAMI` returning the ID of the HVM AMI is kind of confusing as is.  I can't come up with a better name though than `getHVMAMIID` and that's horrible.

PR #524, opened 2016-06-02
Stable channel is supported and default

colhom:
Depends on #523 . the second commit is the content of this PR.

\cc @aaronlevy @cgag @pbx0


cgag:
Can you remind me why we didn't have stable supported before?

Code LGTM.

colhom:
@cgag we need rkt >= 1.0 for kubelet-wrapper to work.

aaronlevy:
rkt 1.0 has been in stable for a while... but the kubelet-wrapper just made it to stable.

aaronlevy:
Just want to note that we're going to need to use alpha again for supporting rkt+kubernetes (need newer versions of rkt).

PR #539, opened 2016-06-09
docs: use correct hyperkube version tag

robszumski:
This should be referring to [v1.2.4_coreos_cni.1](https://quay.io/repository/coreos/hyperkube?tag=v1.2.4_coreos_cni.1), right?

Side note: evidence that this versioning scheme may be semantically correct, but also confusing.

### DIFF
--- a/Documentation/deploy-workers.md
+++ b/Documentation/deploy-workers.md
@@ -80,7 +80,7 @@ Create `/etc/systemd/system/kubelet.service` and substitute the following variab
 * Replace `${MASTER_HOST}`
 * Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
 * Replace `${DNS_SERVICE_IP}`
-* Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release. If using Calico a version that includes CNI binaries should be used. e.g. `v1.2.4_coreos.cni.1`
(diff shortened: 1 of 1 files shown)


pbx0:
`v1.2.4_coreos.cni.1` I believe is correct. Its the valid semver `v1.2.4+coreos.cni.1` with the build metadata tag `+` replaced with an `_` to be a valid docker image tag. The kubernetes binaries will report `v1.2.3+coreos.cni.1` as the version.

I'd love to just use plain semvers so people didn't have to look up our arcane versions, but, semver doesn't really provide a way to version upstream software that may carry additional patches. The semver build metadata tag seems like the closest available option for us.  We essentially need a fourth number to bump if we update an image so that the first 3 still correspond to upstream, but, this isn't valid semantic versioning. Though, since we've already divorced the tag we use for the kubernetes binaries from the image tag we could potentially go with a scheme like `v1.2.4.0` or something for image tags.

robszumski:
One of our docs is wrong, either the master or worker guides, as they refer to different tags.

https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-master.md

Does the master doc need to change instead?

pbx0:
Looks like master is wrong. You can always double check the tags available here: https://quay.io/repository/coreos/hyperkube?tab=tags

robszumski:
Closing in favor of #540

PR #540, opened 2016-06-09
docs: use correct hyperkube version tag

robszumski:
Correct fix from https://github.com/coreos/coreos-kubernetes/pull/539

https://quay.io/repository/coreos/hyperkube


pbx0:
LGTM

PR #542, opened 2016-06-11
Update releaseChannel in cluster.yaml template

jgmize:
stable is now the default in [v0.7.1](https://github.com/coreos/coreos-kubernetes/releases/tag/v0.7.1)


colhom:
Thanks @jgmize! Forgot to change that.

PR #543, opened 2016-06-12
docs: Deprecate HostedZone, use HostedZoneId instead

so0k:
cluster.yaml comments indicate HostedZoneId's should be used when asking `kube-aws` to create DNS records. I added this change as well as provided a method for retrieving the HostedZoneId from a HostedZone using AWS cli.


colhom:
@so0k we're going wait to deprecate hostedZone until we release v1. There are a bunch of other blocks of weird logic that can be pruned away as well once we deprecate `hostedZone` as well.

PR #544, opened 2016-06-13
Discrete etcd cluster

colhom:
Work in progress- no review necessary yet (unless you're bored ;)

### DIFF
--- a/multi-node/aws/pkg/cluster/cluster.go
+++ b/multi-node/aws/pkg/cluster/cluster.go
@@ -2,8 +2,10 @@ package cluster
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -232,12 +234,84 @@ func (c *Cluster) createStack(cfSvc cloudformationService, stackBody string) (*c
 	return cfSvc.CreateStack(creq)
 }
 
+/*
+Makes sure that etcd resource definitions are not upgrades by cloudformation stack update.
+Fetches resource defintions from existing stack and splices them into the updated resource defintions.
+
+TODO(chom): etcd controller + dynamic cluster management will obviate need for this function
+*/
+type cfStackResources struct {
+	Resources map[string]map[string]interface{} `json:"Resources"`
+	Mappings  map[string]map[string]interface{} `json:"Mappings"`
+}
+
+func (c *Cluster) lockEtcdResources(cfSvc *cloudformation.CloudFormation, stackBody string) (string, error) {
+
+	//Unmarshal incoming stack resource defintions
+	var newStack cfStackResources
+	if err := json.Unmarshal([]byte(stackBody), &newStack); err != nil {
+		return "", fmt.Errorf("error unmarshaling new stack json: %v", err)
+	}
+
+	instanceEtcdExpr := regexp.MustCompile("^InstanceEtcd[0-9]+$")
+	//Remove all etcdInstance resource defintions from incoming stack
+	for name, _ := range newStack.Resources {
+		if instanceEtcdExpr.Match([]byte(name)) {
+			fmt.Printf("[lockEtcdResources: REMOVE %s\n", name)
+			delete(newStack.Resources, name)
+		}
+	}
+
+	//Fetch and unmarshal existing stack resource defintions
+	res, err := cfSvc.GetTemplate(&cloudformation.GetTemplateInput{
+		StackName: aws.String(c.ClusterName),
+	})
+	if err != nil {
+		return "", fmt.Errorf("error getting stack template: %v", err)
+	}
+	var existingStack cfStackResources
+	if err := json.Unmarshal([]byte(*res.TemplateBody), &existingStack); err != nil {
+		return "", fmt.Errorf("error unmarshaling existing stack json: %v", err)
+	}
+
+	//splice in existing resource defintions for etcd into new stack
+	for name, definition := range existingStack.Resources {
+		if instanceEtcdExpr.Match([]byte(name)) {
+			fmt.Printf("[lockEtcdResources: ADD %s\n", name)
+			newStack.Resources[name] = definition
+		}
+	}
+	newStack.Mappings["EtcdInstanceParams"] = existingStack.Mappings["EtcdInstanceParams"]
+
+	//create generic json document of outgoing stack
+	var outgoingStack map[string]interface{}
+	if err := json.Unmarshal([]byte(stackBody), &outgoingStack); err != nil {
+		return "", fmt.Errorf("error unmarshaling outgoing stack json: %v", err)
+	}
+
+	//splice in special copies of resources,mappings with etcd-related fields masked
+	outgoingStack["Resources"] = newStack.Resources
+	outgoingStack["Mappings"] = newStack.Mappings
+	// ship off new stack to cloudformation api for an update
+	out, err := json.MarshalIndent(&outgoingStack, "", "")
+	if err != nil {
+		return "", fmt.Errorf("error marshaling stack json: %v", err)
+	}
+
+	return string(out), nil
+}
+
 func (c *Cluster) Update(stackBody string) (string, error) {
+
 	cfSvc := cloudformation.New(c.session)
+	var err error
+	if stackBody, err = c.lockEtcdResources(cfSvc, stackBody); err != nil {
+		return "", err
+	}
 	input := &cloudformation.UpdateStackInput{
 		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
 		StackName:    aws.String(c.ClusterName),
-		TemplateBody: &stackBody,
+		TemplateBody: aws.String(stackBody),
 	}
 
 	updateOutput, err := cfSvc.UpdateStack(input)
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -50,9 +50,13 @@ func newDefaultCluster() *Cluster {
 		WorkerRootVolumeType:     "gp2",
 		WorkerRootVolumeIOPS:     0,
 		WorkerRootVolumeSize:     30,
+		EtcdCount:                1,
+		EtcdInstanceType:         "m3.medium",
+		EtcdRootVolumeSize:       30,
+		EtcdDataVolumeSize:       30,
 		CreateRecordSet:          false,
 		RecordSetTTL:             300,
-		Subnets:                  []Subnet{},
+		Subnets:                  []*Subnet{},
 	}
 }
 
@@ -94,7 +98,7 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 
 	// For backward-compatibility
 	if len(c.Subnets) == 0 {
-		c.Subnets = []Subnet{
+		c.Subnets = []*Subnet{
 			{
 				AvailabilityZone: c.AvailabilityZone,
 				InstanceCIDR:     c.InstanceCIDR,
@@ -122,6 +126,11 @@ type Cluster struct {
 	WorkerRootVolumeIOPS     int               `yaml:"workerRootVolumeIOPS,omitempty"`
 	WorkerRootVolumeSize     int               `yaml:"workerRootVolumeSize,omitempty"`
 	WorkerSpotPrice          string            `yaml:"workerSpotPrice,omitempty"`
+	EtcdCount                int               `yaml:"etcdCount"`
+	EtcdInstanceType         string            `yaml:"etcdInstanceType,omitempty"`
+	EtcdRootVolumeSize       int               `yaml:"etcdRootVolumeSize,omitempty"`
+	EtcdDataVolumeSize       int               `yaml:"etcdDataVolumeSize,omitempty"`
+	EtcdDataVolumeEphemeral  bool              `yaml:"etcdDataVolumEphemeral,omitempty"`
 	VPCID                    string            `yaml:"vpcId,omitempty"`
 	RouteTableID             string            `yaml:"routeTableId,omitempty"`
 	VPCCIDR                  string            `yaml:"vpcCIDR,omitempty"`
@@ -140,12 +149,13 @@ type Cluster struct {
 	HostedZoneID             string            `yaml:"hostedZoneId,omitempty"`
 	StackTags                map[string]string `yaml:"stackTags,omitempty"`
 	UseCalico                bool              `yaml:"useCalico,omitempty"`
-	Subnets                  []Subnet          `yaml:"subnets,omitempty"`
+	Subnets                  []*Subnet         `yaml:"subnets,omitempty"`
 }
 
 type Subnet struct {
-	AvailabilityZone string `yaml:"availabilityZone,omitempty"`
-	InstanceCIDR     string `yaml:"instanceCIDR,omitempty"`
+	AvailabilityZone  string `yaml:"availabilityZone,omitempty"`
+	InstanceCIDR      string `yaml:"instanceCIDR,omitempty"`
+	lastAllocatedAddr *net.IP
 }
 
 const (
@@ -160,7 +170,7 @@ var supportedReleaseChannels = map[string]bool{
 
 func (c Cluster) Config() (*Config, error) {
 	config := Config{Cluster: c}
-	config.ETCDEndpoints = fmt.Sprintf("http://%s:2379", c.ControllerIP)
+
 	config.APIServers = fmt.Sprintf("http://%s:8080", c.ControllerIP)
 	config.SecureAPIServers = fmt.Sprintf("https://%s:443", c.ControllerIP)
 	config.APIServerEndpoint = fmt.Sprintf("https://%s", c.ExternalDNSName)
@@ -199,6 +209,77 @@ func (c Cluster) Config() (*Config, error) {
 		config.VPCRef = fmt.Sprintf("%q", config.VPCID)
 	}
 
+	config.EtcdInstances = make([]etcdInstance, config.EtcdCount)
+	var etcdEndpoints, etcdInitialCluster bytes.Buffer
+	for etcdIndex := 0; etcdIndex < config.EtcdCount; etcdIndex++ {
+
+		//Round-robbin etcd instances across all available subnets
+		subnetIndex := etcdIndex % len(config.Subnets)
+		subnet := config.Subnets[subnetIndex]
+
+		_, subnetCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing subnet instance cidr %s: %v", subnet.InstanceCIDR, err)
+		}
+
+		if subnet.lastAllocatedAddr == nil {
+			ip := subnetCIDR.IP
+			//TODO:(chom) this is sloppy, but "soon-ish" etcd with be self-hosted so we'll leave this be
+			for i := 0; i < 3; i++ {
+				ip = incrementIP(ip)
+			}
+			subnet.lastAllocatedAddr = &ip
+		}
+
+		nextAddr := incrementIP(*subnet.lastAllocatedAddr)
+		subnet.lastAllocatedAddr = &nextAddr
+		instance := etcdInstance{
+			IPAddress:   *subnet.lastAllocatedAddr,
+			SubnetIndex: subnetIndex,
+		}
+
+		//TODO(chom): validate we're not overflowing the address space
+		//This is complicated, must also factor in DHCP addresses
+		//for ASG components
+
+		//Punt on this- we're going to have an answer for dynamic etcd clusters at some point. Then we can either throw
+		//the instances in an ASG and use DHCP like all other instances, or simply self-host on cluster
+
+		config.EtcdInstances[etcdIndex] = instance
+
+		//TODO: ipv6 support
+		if len(instance.IPAddress) != 4 {
+			return nil, fmt.Errorf("Non ipv4 address for etcd node: %v", instance.IPAddress)
+		}
+
+		//http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#concepts-private-addresses
+
+		var dnsSuffix string
+		if config.Region == "us-east-1" {
+			// a special DNS suffix for the original AWS region!
+			dnsSuffix = "ec2.internal"
+		} else {
+			dnsSuffix = fmt.Sprintf("%s.compute.internal", config.Region)
+		}
+
+		hostname := fmt.Sprintf("ip-%d-%d-%d-%d.%s",
+			instance.IPAddress[0],
+			instance.IPAddress[1],
+			instance.IPAddress[2],
+			instance.IPAddress[3],
+			dnsSuffix,
+		)
+
+		fmt.Fprintf(&etcdEndpoints, "https://%s:2379", hostname)
+		fmt.Fprintf(&etcdInitialCluster, "%s=https://%s:2380", hostname, hostname)
+		if etcdIndex < config.EtcdCount-1 {
+			fmt.Fprintf(&etcdEndpoints, ",")
+			fmt.Fprintf(&etcdInitialCluster, ",")
+		}
+	}
+	config.EtcdEndpoints = etcdEndpoints.String()
+	config.EtcdInitialCluster = etcdInitialCluster.String()
+
 	return &config, nil
 }
 
@@ -232,6 +313,7 @@ type StackTemplateOptions struct {
 	TLSAssetsDir          string
 	ControllerTmplFile    string
 	WorkerTmplFile        string
+	EtcdTmplFile          string
 	StackTemplateTmplFile string
 }
 
@@ -239,6 +321,7 @@ type stackConfig struct {
 	*Config
 	UserDataWorker        string
 	UserDataController    string
+	UserDataEtcd          string
 	ControllerSubnetIndex int
 }
 
@@ -289,6 +372,7 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
 	if controllerIPAddr == nil {
 		return nil, fmt.Errorf("invalid controllerIP: %s", stackConfig.ControllerIP)
 	}
+
 	controllerSubnetFound := false
 	for i, subnet := range stackConfig.Subnets {
 		_, instanceCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
@@ -309,6 +393,9 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
 	}
 	if stackConfig.UserDataController, err = execute(opts.ControllerTmplFile, stackConfig.Config, compressUserData); err != nil {
 		return nil, fmt.Errorf("failed to render controller cloud config: %v", err)
+	}
+	if stackConfig.UserDataEtcd, err = execute(opts.EtcdTmplFile, stackConfig.Config, compressUserData); err != nil {
+		return nil, fmt.Errorf("failed to render etcd cloud config: %v", err)
 	}
 
 	return &stackConfig, nil
@@ -333,6 +420,10 @@ func (c Cluster) ValidateUserData(opts StackTemplateOptions) error {
 		{
 			Content: stackConfig.UserDataController,
 			Name:    "UserDataController",
+		},
+		{
+			Content: stackConfig.UserDataEtcd,
+			Name:    "UserDataEtcd",
 		},
 	} {
 		report, err := validate.Validate([]byte(userData.Content))
@@ -413,10 +504,18 @@ func getContextString(buf []byte, offset, lineCount int) string {
 	return string(buf[leftLimit:rightLimit])
 }
 
+type etcdInstance struct {
+	IPAddress   net.IP
+	SubnetIndex int
+}
+
 type Config struct {
 	Cluster
 
-	ETCDEndpoints     string
+	EtcdEndpoints      string
+	EtcdInitialCluster string
+	EtcdInstances      []etcdInstance
+
 	APIServers        string
 	SecureAPIServers  string
 	APIServerEndpoint string
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -2,20 +2,15 @@
 coreos:
   update:
     reboot-strategy: "off"
+
   flannel:
     interface: $private_ipv4
-    etcd_endpoints: {{ .ETCDEndpoints }}
-  etcd2:
-    name: controller
-    advertise-client-urls: http://$private_ipv4:2379
-    initial-advertise-peer-urls: http://$private_ipv4:2380
-    listen-client-urls: http://0.0.0.0:2379
-    listen-peer-urls: http://0.0.0.0:2380
-    initial-cluster: controller=http://$private_ipv4:2380
-  units:
-    - name: etcd2.service
-      command: start
+    etcd_endpoints: {{ .EtcdEndpoints }}
+    etcd_cafile: /etc/kubernetes/ssl/ca.pem
+    etcd_certfile: /etc/kubernetes/ssl/etcd-client.pem
+    etcd_keyfile: /etc/kubernetes/ssl/etcd-client-key.pem
 
+  units:
     - name: docker.service
       drop-ins:
         - name: 40-flannel.conf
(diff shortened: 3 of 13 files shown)


wolfeidau:
Would love to test this once your happy with it, if your looking for volunteers.

colhom:
@wolfeidau the TLS stuff still isn't done yet, not usable. I've been slacking on this one but will circle back very soon- if you have some time, I'm accepting sub-PRs on `colhom:discrete-etcd-cluster` fork branch ;)

mumoshu:
@colhom What a PR! Great job. Btw, would you mind sharing me which part of TLS stuff isn't done?(I've read through your code and couldn't figure it)

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -60,6 +60,8 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # Price (Dollars) to bid for spot instances. Omit for on-demand instances.
 # workerSpotPrice: "0.05"
 
+# Instance type for etcd node
+
```
mumoshu: nit: Missing descriptions and examples for newly added parameters

```
#etcdInstanceType: "m3.medium"

# Number of etcd nodes(should be 1, 3, 5 and so on to prevent split-brain)
# etcdCount: 1

# Disk size (GiB) for etcd node
#etcdRootVolumeSize: 30

# Disk size (GiB) for etcd node. This one is user by Etcd to store its key-value data and write-ahead logs
#etcdDataVolumeSize: 30
```

multi-node/aws/pkg/config/config.go
```diff
@@ -181,20 +191,84 @@ func (c Cluster) Config() (*Config, error) {
...
+
+		//http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#concepts-private-addresses
+
+		var dnsSuffix string
+		if config.Region == "us-east-1" {
+			// a special DNS suffix for the original AWS region!
+			dnsSuffix = "ec2.internal"
+		} else {
+			dnsSuffix = fmt.Sprintf("%s.compute.internal", config.Region)
+		}
```
mumoshu: Encountered this yesterday and have patched exactly the same as you did 😄  Great.

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+
+        [Install]
+        RequiredBy=var-lib-etcd2.mount
+
+    - name: decrypt-tls-assets.service
+      enable: true
+      content: |
+        [Unit]
+        Description=decrypt kubelet tls assets using amazon kms
+        Before=kubelet.service
```
mumoshu: Probably unnecessary because we don't have kubelets in etcd nodes?

colhom: we have etcd key/cert and the ca cert.

mumoshu: Ah, sorry. I meant about the `Before=kubelet.service` part.

colhom: oh yes, i'm blind ;) that should be changed to etcd2.

colhom:
@wolfeidau go ahead a give it a whirl! I recommend jumping ahead to #596 and grabbing the HA control plane stuff as well on top of this PR- should all be functional.

colhom:
Hi everyone- update on things.

I've been experimenting with full cluster upgrades and frankly- statically defined etcd clusters (ie: this PR) do not work well with CloudFormation upgrades. 

The only strategy that really works well "in general" with CloudFormation updates is doing a rolling replacement on an autoscaling group- and currently we don't have a good solution for running etcd robustly in an ASG.

We're currently investigating various possibilities for implementing a dynamic/managed etcd cluster- maybe in an ASG, or more optimally hosting etcd on kubernetes and greatly simplifying the CloudFormation stack as a result. The eventual goal is to entirely decouple Kubernetes cluster level upgrades (control plane, kubelet, etcd) from updating the underlying AWS virtual hardware. Etcd on Kubernetes would be a step in the right direction.

Input?

colhom:
The band-aide for now would be to manually ensure that the etcd instances are never updated by cloudformation after the are created.  This is crappy for a lot of reasons, but the only workable solution until we can have etcd nodes that aren't pets. At least CoreOS can take care of doing the OS upgrades :+1: 

On `kube-aws up --upgrade`, we could pull the existing resource definitions for the etcd instances and splice that in to our outgoing template body in the update stack api call- ensuring no doomed-to-fail update action is taken on those instances.

It's ugly, but it's a grokable hack that would allow us to limp along and take care of our named etcd pets until we have etcd on Kubernetes ready for prime time. 

\cc @mumoshu @pieterlange

pieterlange:
@colhom i'm personally looking into implementing https://crewjam.com/etcd-aws/ as my backing cluster. I actually had a bad experience **today** losing etcd state across a autoscaling group managed by [monsanto's solution](https://github.com/MonsantoCo/etcd-aws-cluster).

This is after 3 months+ of running perfectly etcd-aws-cluster btw. There's a lot of work involved in getting software production ready and @crewjam's solution seems the most fleshed out, so after tryouts chances are i'll be going with that. The integrated backup (recovery from complete cluster failure) is the selling point for me after today.

colhom:
@pieterlange sorry to hear about your data loss. Out of morbid curiosity- did your etcd instances fail health checks due to load, causing the ASG to replace them one-by-one until your data was gone?

pieterlange:
Nothing like that, but i did run them (3) on t2.nano's. (with 20 workers and 1 master) I spun up the etcd cluster separately and never really "managed" it.. it's a small miracle i got away with it as long as i did :)

crewjam:
@pieterlange and everyone -- I'd be happy to work with you on this. We did a bit of research and planning on production use of etcd on AWS which resulted in the repo & blog describing it. One caveat though -- for reasons not at all related to the work, our focus shifted a bit and so we haven't *actually* put it into production. We got close, and we are still planning to, but it hasn't happened just yet.

colhom:
@dghubble @aaronlevy @pbx0 PTAL

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -232,12 +234,78 @@ func (c *Cluster) createStack(cfSvc cloudformationService, stackBody string) (*c
 	return cfSvc.CreateStack(creq)
 }
 
+/*
+Makes sure that etcd resource definitions are not upgrades by cloudformation stack update.
+Fetches resource defintions from existing stack and splices them into the updated resource defintions.
+
+TODO(chom): etcd controller + dynamic cluster management will obviate need for this function
```
aaronlevy: What are the benefits of introducing new complexity if we're already planning on replacing it?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -232,12 +234,78 @@ func (c *Cluster) createStack(cfSvc cloudformationService, stackBody string) (*c
...
+*/
+type cfStackResources struct {
+	Resources map[string]map[string]interface{} `json:"Resources"`
+}
+
+func (c *Cluster) lockEtcdResources(cfSvc *cloudformation.CloudFormation, stackBody string) (string, error) {
```
aaronlevy: Can you add a bit more documentation explaining why this is necessary / how the functionality will be replaced (for those of us unfamiliar with cloudformation).

multi-node/aws/pkg/config/config.go
```diff
@@ -199,6 +209,77 @@ func (c Cluster) Config() (*Config, error) {
...
+		_, subnetCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing subnet instance cidr %s: %v", subnet.InstanceCIDR, err)
+		}
+
+		if subnet.lastAllocatedAddr == nil {
+			ip := subnetCIDR.IP
+			//TODO:(chom) this is sloppy, but "soon-ish" etcd with be self-hosted so we'll leave this be
+			for i := 0; i < 3; i++ {
+				ip = incrementIP(ip)
```
aaronlevy: What is the purpose of this?

multi-node/aws/pkg/cluster/cluster.go
colhom: **Problem**: We must ensure that a cloudformation update never affects the etcd cluster instances.

**Reason**: When update ec2 instances, cloudformation may choose a destroy and replace strategy, which has the potential to corrupt the etcd cluster.

**Solution**: `lockEtcdResources` is a 50-line function which only runs on cluster update- it queries the current state of the etcd cluster and ensures the definition does not change across CloudFormation updates.

**Concerns over fragility**: This is a simple, mechanical operation- but even those can go wrong. 

Assuming:

* _lockEtcdCluster()  errors out_: the update will fail

* _lockEtcdClusteR() munges the etcd cluster definition_: In that case, Cloudformation will attempt to update the etcd cluster instances and fail because they are marked as "non-updatable" when first created. The CloudFormation update will then roll-back, and the user will be left with a non-updatable cluster

In either case, the failure mode is "cluster non-updatable"- which is exactly where every user is today anyways.

multi-node/aws/pkg/config/config.go
```diff
@@ -199,6 +209,77 @@ func (c Cluster) Config() (*Config, error) {
...
+		//This is complicated, must also factor in DHCP addresses
+		//for ASG components
+
+		//Punt on this- we're going to have an answer for dynamic etcd clusters at some point. Then we can either throw
+		//the instances in an ASG and use DHCP like all other instances, or simply self-host on cluster
+
+		config.EtcdInstances[etcdIndex] = instance
+
+		//TODO: ipv6 support
+		if len(instance.IPAddress) != 4 {
```
aaronlevy: `if instance.IPAddress.To4() == nil`?

multi-node/aws/pkg/cluster/cluster.go
colhom: ^^ Something to that effect?

multi-node/aws/pkg/config/config.go
colhom: Assuming a `10.0.0.0/24`, we put the first etcd instance at `10.0.0.3`. That's saying start 3 addresses from the lowest address in the CIDR.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -133,17 +135,18 @@ coreos:
       content: |
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
-        Before=kubelet.service
-        After=docker.service
-        Requires=docker.service
+        Before=kubelet.service flanneld.service
+        After=early-docker.service
+        Requires=early-docker.service
```
dghubble: why is this needed now?

dghubble:
Is the separate etcd optional, doesn't look like it? How are you planning to remove/merge this feature with self-hosted Kubernetes? Why is the switch the TLS added in this commit as well?

pieterlange:
Self-hosted would be really nice, but speaking as a lowly sysadmin: there's enough moving parts in this system already. I need to be able to rely on/trust a stable etcd. I'm even proposing to leave this outside of the management of kube-aws altogether in #629!

multi-node/aws/pkg/cluster/cluster.go
aaronlevy: sgtm

multi-node/aws/pkg/config/config.go
aaronlevy: Any particular reason?

multi-node/aws/pkg/config/config.go
```diff
@@ -199,6 +209,77 @@ func (c Cluster) Config() (*Config, error) {
...
+
+		var dnsSuffix string
+		if config.Region == "us-east-1" {
+			// a special DNS suffix for the original AWS region!
+			dnsSuffix = "ec2.internal"
+		} else {
+			dnsSuffix = fmt.Sprintf("%s.compute.internal", config.Region)
+		}
+
+		hostname := fmt.Sprintf("ip-%d-%d-%d-%d.%s",
```
aaronlevy: Why do we need the hostname? If we know the IP can't that just be used directly?

multi-node/aws/pkg/config/config.go
```diff
@@ -413,10 +504,18 @@ func getContextString(buf []byte, offset, lineCount int) string {
 	return string(buf[leftLimit:rightLimit])
 }
 
+type etcdInstance struct {
+	IPAddress   net.IP
+	SubnetIndex int
```
aaronlevy: Is it possible for the subnets to change externally to the etcd configuration?

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: I'm assuming this is because this is being executed via docker run, but I'd prefer we used rkt

multi-node/aws/pkg/config/config.go
colhom: Not really- needed to pick a small offset, so 3 seemed like a good one.

multi-node/aws/pkg/config/config.go
colhom: yep

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+
+            Environment=ETCD_INITIAL_CLUSTER_STATE=new
+            Environment=ETCD_INITIAL_CLUSTER={{.EtcdInitialCluster}}
+            Environment=ETCD_DATA_DIR=/var/lib/etcd2
+            Environment=ETCD_LISTEN_CLIENT_URLS=https://%H:2379
+            Environment=ETCD_ADVERTISE_CLIENT_URLS=https://%H:2379
+            Environment=ETCD_LISTEN_PEER_URLS=https://%H:2380
+            Environment=ETCD_INITIAL_ADVERTISE_PEER_URLS=https://%H:2380
+            PermissionsStartOnly=true
+            ExecStartPre=/usr/bin/chown -R etcd:etcd /var/lib/etcd2
```
aaronlevy: Seems a bit heavy-handed. Why is this necessary?

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+            Environment=ETCD_PEER_CERT_FILE=/etc/etcd2/ssl/etcd.pem
+            Environment=ETCD_PEER_KEY_FILE=/etc/etcd2/ssl/etcd-key.pem
+
+            Environment=ETCD_CLIENT_CERT_AUTH=true
+            Environment=ETCD_TRUSTED_CA_FILE=/etc/etcd2/ssl/ca.pem
+            Environment=ETCD_CERT_FILE=/etc/etcd2/ssl/etcd.pem
+            Environment=ETCD_KEY_FILE=/etc/etcd2/ssl/etcd-key.pem
+
+            Environment=ETCD_INITIAL_CLUSTER_STATE=new
+            Environment=ETCD_INITIAL_CLUSTER={{.EtcdInitialCluster}}
```
aaronlevy: If we're already templating this file, we don't necessarily need the %H hostname replacement and can use IPs directly.

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+    - name: etcd2.service
+      drop-ins:
+        - name: 20-etcd2-aws-cluster.conf
+          content: |
+            [Unit]
+            Requires=decrypt-tls-assets.service
+            After=decrypt-tls-assets.service
+
+            [Service]
+            Environment=ETCD_NAME=%H
```
aaronlevy: Why change from default: https://github.com/coreos/coreos-overlay/blob/master/dev-db/etcd/files/etcd2.service#L9 ?

multi-node/aws/pkg/config/config.go
colhom: If given the option, it's generally considered preferable to list a DNS name (either CN or SUBJECT_ALT) rather than using IP SANS.

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
+#cloud-config
+coreos:
+  update:
+    reboot-strategy: etcd-lock
+  units:
+    - name: etcd2.service
+      drop-ins:
```
aaronlevy: Any specific reason to use a drop-in vs letting cloud-config do the translation? https://coreos.com/os/docs/latest/cloud-config.html#etcd2

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+        [Unit]
+        Description=Formats etcd2 ebs volume
+        After=dev-xvdf.device
+        Requires=dev-xvdf.device
+        Before=var-lib-etcd2.mount
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=/opt/bin/ext4-format-volume-once /dev/xvdf
```
aaronlevy: This scares me a bit. And probably a much better fit for ignition -- have you considered using that instead?

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+
+  - path: /opt/bin/decrypt-tls-assets
+    owner: root:root
+    permissions: 0700
+    content: |
+      #!/bin/bash -e
+
+      for encKey in $(find /etc/etcd2/ssl/*.pem.enc);do
+        tmpPath="/tmp/$(basename $encKey).tmp"
+        docker run --net host --rm -v /etc/etcd2/ssl:/etc/etcd2/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext | base64 --decode > $tmpPath
```
aaronlevy: would prefer that we use rkt here

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -72,6 +72,28 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # Price (Dollars) to bid for spot instances. Omit for on-demand instances.
 # workerSpotPrice: "0.05"
 
+## Etcd Cluster config
```
aaronlevy: Probably need documentation going over how the nodes will be distributed across subnets / how that logic works. E.g. we recommend X nodes, with Y subnets, and it will be distributed in this way.

multi-node/aws/pkg/config/config.go
colhom: Cluster upgrades can never support updating the Subnet or VPC parameters, unfortunately. I've had users ask about it- there are complex manual migration steps which cannot currently be automated :(

Now that you've asked, I'll make sure to make that explicit rather than just letting cloudformation roll back the update. https://github.com/coreos/coreos-kubernetes/pull/608#issuecomment-246045559

multi-node/aws/pkg/config/tls_config.go
```diff
@@ -85,10 +93,23 @@ func (c *Cluster) NewTLSAssets() (*RawTLSAssets, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	workerConfig := tlsutil.ClientCertConfig{
 		CommonName: "kube-worker",
 		DNSNames: []string{
-			"*.*.compute.internal",
+			fmt.Sprintf("*.%s.compute.internal", c.Region),
 			"*.ec2.internal",
```
aaronlevy: nit: you could do the same region check here to only use this if in us-east-1

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Yeah, that would be preferable. This PR or follow-up?

colhom:
> Is the separate etcd optional?

No , it's very much mandatory. If etcd is not on separate machines, Kubernetes ec2 instances cannot be updated, which is a big deal

It's a pre-req for dynamically scaling compute resources for both control plane and worker pool- something AWS users (including me!) are very excited about.

 Until self-hosted _etcd_ is stable and production ready, this is a requirement.


> Are you planning to remove/merge this feature with self-hosted Kubernetes, they seem at odds?

Bootkube currently provisions a static single-node etcd instance and points the control plane at it- and that's really it. A single point of non-fault-tolerant failure for your otherwise fault-tolerant control plane.

If anything, this would complement self-hosted by making etcd more robust.( > 1 nodes, multi-zone, etc) and really not change anything about how self-hosted works currently, as neither bootkube nor this PR deal at all with self-hosting etcd.

> Also, the addition of etcd TLS here seems orthogonal

If an etcd interface is exposed to any sort of network, TLS encryption is a must. We have just exposed etcd to the subnet, hence encyption becomes mandatory. If it's only listening on localhost, not as big of a deal.

@dghubble

colhom:
Update- just did some fishing around it bootkube- it appears the multi-node bootkube example actually uses a discrete/external multi-node etcd cluster- just like this PR! So, in reality, this would make bootkube on AWS look almost exactly like the multi-node vagrant setup, just plus TLS!

multi-node/aws/pkg/config/templates/cloud-config-etcd
colhom: Yeah, we have plans to switch over. Ignition has great support for doing this conditional type of disk formatting operation for you.

aaronlevy:
@colhom that review pass is done. Mostly questions at this stage.

@dghubble @pieterlange I'm personally on the fence about this functionality as well - but @colhom is championing for this to be included, and I won't explicitly block on "I feel iffy about this" if @colhom is willing to support long-term.

My general concerns:

This feels somewhat like a big workaround. If cloud-formation does not support this type of deployment, why are we fighting it to try? Parsing out the etcd-sections to keep them from being changed really does seem fragile to me - but to be fair I'm not very familiar with cloud-formation.

We have near term plans to moving toward self-hosted installations and this diverges from that work. This may be fine, it just means we will likely end up with a separate tool (which to be fair, may have happened anyway).

kube-aws could generate multiple cloud-formations in the case that you want a larger etcd cluster. I'm still unclear what we gain from trying to force this all into a single deployment (when we have to work around the fact that it doesn't actually work natively).

The single master likely works well for a lot of people (kube-up, gke, kops, kube-aws (until now)). I also still question that in-place updates cannot occur - can you remind me what the issue is here? Something about ebs volumes will not be reattached after a cloud-formation update?

aaronlevy:
Also, fwiw etcd was always on the worker nodes (for flannel) -- so the TLS stuff isn't just now necessary due to running multiple etcd nodes -- it's always been a need. But along those same lines - as of v1.4 (and self-hosted) we're moving toward no worker etcd access at all - so another temporary divergence here.

multi-node/aws/pkg/config/templates/cloud-config-etcd
aaronlevy: Potentially formatting a volume on every boot based on |grep output is something I would personally not be okay with on systems I was managing. How confident in this are you?

multi-node/aws/pkg/config/config.go
aaronlevy: What is the reason for the small offset? What are you trying to accomplish here and why? I'm trying to understand why we need this code - right now it seems completely arbitrary.

multi-node/aws/pkg/config/config.go
aaronlevy: We have to use IP sans all over the place -- why is this particular use-case a concern?

multi-node/aws/pkg/config/config.go
aaronlevy: Where is this indexed? The config file? There needs to be an explicit "order cannot ever change in this config file" if that's the case.

aaronlevy: Is it possible to use explicit IDs instead?

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: This pr

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+        RequiredBy=etcd2.service
+
+
+write_files:
+  - path: /opt/bin/ext4-format-volume-once
+    permissions: 0700
+    owner: root:root
+    content: |
+      #!/bin/bash -e
+      if [[ "$(wipefs -n -p $1 | grep ext4)" == "" ]];then
```
aaronlevy: If you actually want to do this (it scares me) use full paths to binaries and long flag names

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,122 @@
...
+        mkfs.ext4 $1
```
aaronlevy: Quote: "$1"

multi-node/aws/pkg/config/config.go
colhom: Yeah that would be a more robust way of going about it- I'll change that. The re-ordering of subnets thing would be kind of a stupid constraint.

multi-node/aws/pkg/config/templates/cloud-config-etcd
colhom: I don't like it either- I'm soliciting alternatives from the OS team. The plan to move over to ignition is imminent though.

multi-node/aws/pkg/config/templates/cluster.yaml
colhom: Agreed. Will put that in there, along with links to AWS ec2 SLA.

multi-node/aws/pkg/config/templates/cloud-config-etcd
colhom: I'm trying to get a less sketchy solution in place.

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -0,0 +1,133 @@
...
+      blkid -t TYPE=ext4 "$1"
+      exitStatus=$?
+      set -e
+
+      if [[ $exitStatus -eq 0 ]];then
+        echo "volume $1 is already ext4 formatted"
+        exit 0
+      elif [[ $exitStatus -eq 2 ]];then
+        echo "No ext4 fs detected- will format"
+        mkfs.ext4 -p "$1"
```
crawford: If you are using `-p`, why bother with blkid?

colhom: Just to be explicit? I guess we should just keep this as simple as possible... I'll change it

multi-node/aws/pkg/config/templates/cloud-config-etcd
```diff
@@ -61,7 +61,7 @@ coreos:
         [Service]
         Type=oneshot
         RemainAfterExit=yes
-        ExecStart=/opt/bin/ext4-format-volume-once /dev/xvdf
+        ExecStart=/usr/sbin/mkfs.ext4 -p /dev/xvdf
```
colhom: @aaronlevy final answer- wdyt?

dghubble:
@colhom Sure thing. It sounds perfectly reasonable for kube-aws to focus on the needs of this particular cluster design, platform, and users. My notes just highlight conflicts between different cluster designs, but it sounds like that's a non-goal, please disregard.

multi-node/aws/pkg/config/templates/cloud-config-etcd
colhom: disregard my previous comment- after talking to crawford I've reverted to first checking with `blkid`, and then leaving the `-p` arg to `mkfs` for extra safety.

colhom:
@aaronlevy any more review items? I've addressed the conditional `ext4` formatting logic in a significantly more robust way in the latest two commits, and am running a conformance test on this now.

pieterlange:
I do not wish to delay this feature because i think the work is just fine. But maybe it _would_ be better to spin up etcd as a separate cloudformation template - this is exactly what i've been doing and what's reflected in my PR (#629). I guess it kind of depends on if you can live with having to spin up mutliple cloudformation templates. I can :-)

colhom:
@pieterlange I have experimented with the separate cf stack approach for etcd- there is too much added complexity in cross-correlating the resources and really not much to be gained (in my opinion).

I've discussed with @aaronlevy  and this will not merge into master until after the v1.4 hyperkube and associated changes are in (weeks).

To unblock further development in this area, we will be merging this PR, HA control plane, cluster upgrades and @mumoshu 's `kubectl drain` PR on a separate experimental branch. After master is updated to v1.4, we will look at merging the experimental stuff back in.

I'll update the community once that branch is ready and passing conformance/e2e tests.

aaronlevy:
I am still wondering about this:

> I also still question that in-place updates cannot occur - can you remind me what the issue is here? Something about ebs volumes will not be reattached after a cloud-formation update?

aaronlevy:
Also, can you expand on:

> there is too much added complexity in cross-correlating the resources

Is it not just etcd network addressability (hostnames/ips) that needs to be shared?

pieterlange:
>Is it not just etcd network addressability (hostnames/ips) that needs to be shared?

that's basically all i'm doing in the PoC PR. Obviously the situation is a little more hairy than that and my PR is just a complete hack. But it worked fine for me.

colhom:
> I also still question that in-place updates cannot occur - can you remind me what the issue is here? Something about ebs volumes will not be reattached after a cloud-formation update?

If a CloudFormation update decides it needs to replace a machine, there is no way to ensure the new machine to receives the same EBS volume as the old machine. So our single controller/ebs backed etcd combo is non-updatable via CloudFormation. More generally, any machine running etcd is non-updatable, whether or not it's EBS backed. That's why this PR separates etcd out and quarantines the instances off from updates, until we can host on Kubernetes.

> Also, can you expand on:
>> there is too much added complexity in cross-correlating the resources

Beyond what @pieterlange mentioned, additional points of difficulty:
 * security group references for ingress rules between two stacks would have to be done manually by making read calls to cloudformation API.
 * having the stacks share subnets is quite tricky, so in reality we'd need separate subnets for etcd instances and k8s instances, each managed by their respective stack. Then there would be ACL's in the middle, along with shared VPC. Basically means you have to create the stacks in a specific order and tear them down in a specific order, with some manual steps ahead of time to "detach" them from eachother. Right now we have a single API call to create, and a single api call to destroy.

In general, breaking the "everything is in this single cloudformation stack" invariant greatly changes the nature of kube-aws, and is out of scope for a problem that is will be solved in a better way within the next 6 months (or so :tm:). 

The [hackery](https://github.com/coreos/coreos-kubernetes/pull/544/files#diff-57c0e4221c5add72bb835d26badd2241R247) necessary to quarantine etcd instances from cloudformation updates boils down to masking updates to a few select subtrees of a cloudformation json document. It fits neatly within the confines of a single function which is referenced in [one place](https://github.com/coreos/coreos-kubernetes/pull/544/files#diff-57c0e4221c5add72bb835d26badd2241R302). You can literally comment out the function and it's one reference, and it's gone.

\cc @aaronlevy

multi-node/aws/pkg/config/config.go
colhom: As we move towards using cross-zone load balancers for the various interfaces (etcd, k8s api mainly), DNS names will need to supplant IP addresses. The only places where I see IP SANS sticking around is maybe the kubelet's certificate (validated by api server when exec'ing), but won't even that go away with bootkube-flavor self-hosted?

pieterlange:
Against my best intentions i feel like my comment has started to derail/undermine the good work being done here. I trust @colhom's judgment on this item.

I suggest discussing the pro's and con's of managing etcd under a separate cloudformation under my PR or a new issue so we don't further pollute this thread. I like the idea of hosting etcd [separately and robustly](https://crewjam.com/etcd-aws/) but that's only because of 1 nasty prior experience.

mumoshu:
@colhom Thanks for your great work here and there. I'm really looking forward to see this get merged 👍
Sorry for bothering you but did you have any chance to set the experimental branch up?
 I'd like to experiment #667 on top of it.

dzavalkinolx:
We created our fork of kube-aws and merged this requested. After ~3 months of experience with I have to admit that this is not a way to go in production. Because 3 etcd nodes have static IPs there is no option to update them without some manual interaction. Since there is no backup of etcd state to somewhere outside cluster (for example S3) manual interactions are quite risky. I decided not to upgrade etcd nodes (by manually editing generated stack template json file). As a result now controller and worker nodes in one of the clusters I manage are running latest CoreOS AMI with Dirty COW bug fixed but etcd nodes run old AMI.

I believe the right approach is described in this article: https://crewjam.com/etcd-aws/, i.e. separate ASG for etcd nodes and etcd state mirroring/backup to S3. In this case etcd nodes can be also upgraded by running cloudformation stack update without any manual interaction.

camilb:
@dzavalkinolx I also forked `kube-aws`and merged #608 and #629 plus some changes. Also started to integrate crewjam's `etcd-aws `. Got it working but without TLS for now. I see more people are looking on this solution and I think the best place to track the progress is here: https://github.com/coreos/kube-aws/issues/9

pieterlange:
@dzavalkinolx @camilb i made a new issue for this at https://github.com/coreos/kube-aws/issues/27. I think we're not far off from a solution but it's nice to hear everyone's input on this.

PR #547, opened 2016-06-16
docs: kube2sky needs to know kube-master-url

Jiaweizdev:
Kube2sky will not work if we do not specify `--kube-master-url` here.


aaronlevy:
If run in-cluster it should be able to determine the location of the api-server via environmental variables injected into the pod. See: http://kubernetes.io/docs/user-guide/environment-guide

So if your cluster is configured correctly, the pod should be able to inspect its local environment, get the kubernetes service IP, then the kube-proxy will route from that IP to the correct location.

Issue #501, opened 2016-05-24
rkt: add a flag to enable rkt support


philips:
Kubernetes v1.3 will have full rkt support; see the [rktnetes v1.3 milestone](https://github.com/kubernetes/kubernetes/milestones/rktnetes-v1.0).

We need to make this as easy to enable as the Calico support or anything else. It should just be a single configuration option on vagrant, aws, etc.

We should be able to start experimenting with this ASAP as much of the support is in k8s master. So, we should start tracking k8s v1.3 pre-releases and getting this into shape.

Quentin-M:
Hi,

This issue just became even more important as it is a hard requirement for Stackanetes. Fully supporting OpenStack services on top of Rktnetes, following the coreos-kubernetes docs is expected by the end of July.

Right now, we can't simply swap _docker_ with _rkt_ on a Kubernetes worker node because rkt isn't available in hyperkube and thus, the kuberlet-wrapper can't run kubelet in rkt-fly. I understand that @pbx0 has started some work around that [[1]] but faced some issues with rkt (permissions) while trying to run kube-proxy [[2]]. Intel apparently suffers from the same issue since April [[3]]. It has been worked around by making it a systemd service though. I tried quickly to run the kubelet directly on CoreOS 1068, without kubelet-wrapper, but it simply won't run: breaking in the very first steps (getting rkt's version). Not exactly sure yet if the version is too old or if this is an actual bug - I am sure @yifan-gu knows.

What's the current status on this? Is there anyone working on this actively? I believe that it is quite important to get that done ASAP as it will significantly increase the user base of rktnetes and rkt - and thus, we'll get a lot more feedbacks.

It is also blocking our very first tasks on Stackanetes as it is complicated to run any k8s manifests with rktnetes. Is there any recommendation right now?

```
$ /usr/bin/kubelet \
>   --container-runtime=rkt \
>   --api-servers=https://XXX.YYY.ZZZZ.WWW \
>   --network-plugin-dir=/etc/kubernetes/cni/net.d \
>   --network-plugin= \
>   --register-node=true \
>   --allow-privileged=true \
>   --config=/etc/kubernetes/manifests \
>   --hostname-override=XXX.YYY.ZZZZ.WWW \
>   --cluster-dns=10.3.0.10 \
>   --cluster-domain=cluster.local \
>   --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
[...]

PR #551, opened 2016-06-21
runtime toggle

pbx0:
This adds the ability to toggle container runtime in coreos-kubernetes.

This is still a WIP but mostly complete and blocked on some external factors. To use rktnetes you just set the runtime flag to `rkt` and the release channel to `alpha` if using kube-aws.

One temporary workaround is escaping the container chroot for invocations of rkt. I expect to merge this for now though.

Updates https://github.com/coreos/coreos-kubernetes/issues/501


yifan-gu:
cc @tmrts @s-urbaniak @euank

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -23,13 +23,17 @@ coreos:
...
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf"
+        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf --volume rktbin,kind=host,source=/usr/bin/rkt --mount volume=rktbin,target=/usr/bin/rkt --volume var-lib-rkt,kind=host,source=/var/lib/rkt --mount volume=var-lib-rkt,target=/var/lib/rkt --volume=stage,kind=host,source=/usr/lib/rkt --mount volume=stage,target=/usr/lib/rkt --volume=mp,kind=host,source=/usr/sbin/modprobe --mount volume=mp,target=/usr/bin/modprobe"
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers={{.SecureAPIServers}} \
         --network-plugin-dir=/etc/kubernetes/cni/net.d \
         --network-plugin={{.K8sNetworkPlugin}} \
+        --container-runtime=rkt \
+        --rkt-path=/usr/bin/rkt \
+        --rkt-stage1-image=/usr/lib/rkt/stage1-images/stage1-coreos.aci \
```
yifan-gu: Will need to change this to the image name in the next beta: `coreos.com/rkt/stage1-coreos` to enable stage1 caching.

multi-node/generic/controller-install.sh
```diff
@@ -96,12 +99,17 @@ function init_templates {
...
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
+ExecStartPre=/usr/bin/touch /etc/hosts
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=http://127.0.0.1:8080 \
   --register-schedulable=false \
   --network-plugin-dir=/etc/kubernetes/cni/net.d \
   --network-plugin=${K8S_NETWORK_PLUGIN} \
+  --container-runtime=${RUNTIME} \
+  --rkt-path=/usr/bin/rkt \
+  --rkt-stage1-image=/usr/lib/rkt/stage1-images/stage1-coreos.aci \
```
yifan-gu: ditto

multi-node/generic/worker-install.sh
```diff
@@ -62,11 +65,16 @@ function init_templates {
...
+Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf --volume rktbin,kind=host,source=/usr/bin/rkt --mount volume=rktbin,target=/usr/bin/rkt --volume var-lib-rkt,kind=host,source=/var/lib/rkt --mount volume=var-lib-rkt,target=/var/lib/rkt --volume=stage,kind=host,source=/usr/lib/rkt --mount volume=stage,target=/usr/lib/rkt --volume=mp,kind=host,source=/usr/sbin/modprobe --mount volume=mp,target=/usr/bin/modprobe"
+ExecStartPre=/usr/bin/touch /etc/hosts
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=${CONTROLLER_ENDPOINT} \
   --network-plugin-dir=/etc/kubernetes/cni/net.d \
   --network-plugin=${K8S_NETWORK_PLUGIN} \
+  --container-runtime=${RUNTIME} \
+  --rkt-path=/usr/bin/rkt \
+  --rkt-stage1-image=/usr/lib/rkt/stage1-images/stage1-coreos.aci \
```
yifan-gu: ditto

multi-node/generic/controller-install.sh
```diff
@@ -840,6 +893,17 @@ systemctl stop update-engine; systemctl mask update-engine
 systemctl daemon-reload
 systemctl enable flanneld; systemctl start flanneld
 systemctl enable kubelet; systemctl start kubelet
+
+# TODO(pb): these options together have not been tested. They shouldn't be mutually exclusive options in the future.
```
yifan-gu: We could test this after https://github.com/kubernetes/kubernetes/pull/27756

single-node/user-data
```diff
@@ -104,10 +116,64 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
...
+    local TEMPLATE=/etc/systemd/system/install-kproxy.service
+    [ -f $TEMPLATE ] || {
+        echo "TEMPLATE: $TEMPLATE"
+        mkdir -p $(dirname $TEMPLATE)
+        cat << EOF > $TEMPLATE
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStartPre=/usr/bin/mkdir -p /opt/
+ExecStartPre=/usr/bin/curl --fail --silent --show-error --location --create-dirs --output /opt/kube-proxy https://storage.googleapis.com/kubernetes-release/release/v1.3.0-alpha.5/bin/linux/amd64/kube-proxy
```
yifan-gu: Should we config the proxy's version as well?

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: What do you mean by next beta? Should just changing that path to `https://github.com/coreos/rkt/blob/master/Documentation/configuration.md` work? I don't have this working locally yet.

single-node/user-data
pbx0: Probably, but i'm hoping the remove this work-around still. Noted that if we end up leaving this in, I'll make sure we configure the version.

multi-node/aws/pkg/config/templates/cloud-config-worker
yifan-gu: @pbx0 I mean since now https://github.com/kubernetes/kubernetes/pull/27707 is merged, so instead of using the path to the aci, we will change to using the stage1's name (to enable caching instead of unpacking the image every time)

yifan-gu: @pbx0 This will take effect after we build another k8s release

yifan-gu:
@pbx0 Could you point me a guide on how to use this?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -37,13 +37,17 @@ coreos:
...
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf"
+        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf --volume rktbin,kind=host,source=/usr/bin/rkt --mount volume=rktbin,target=/usr/bin/rkt --volume var-lib-rkt,kind=host,source=/var/lib/rkt --mount volume=var-lib-rkt,target=/var/lib/rkt --volume=stage,kind=host,source=/usr/lib/rkt --mount volume=stage,target=/usr/lib/rkt --volume=mp,kind=host,source=/usr/sbin/modprobe --mount volume=mp,target=/usr/bin/modprobe"
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers=http://localhost:8080 \
         --network-plugin-dir=/etc/kubernetes/cni/net.d \
         --network-plugin={{.K8sNetworkPlugin}} \
+        --container-runtime={{.ContainerRuntime}} \
+        --rkt-path=/usr/bin/rkt \
+        --rkt-stage1-image=/usr/lib/rkt/stage1-images/stage1-coreos.aci \
```
yifan-gu: Needs to be changed to coreos.com/rkt/stage1-coreos

single-node/user-data
```diff
@@ -88,14 +91,20 @@ function init_templates {
...
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
-ExecStart=/usr/lib/coreos/kubelet-wrapper \
+ExecStart=/share/kubelet-wrapper \
   --api-servers=http://127.0.0.1:8080 \
   --network-plugin-dir=/etc/kubernetes/cni/net.d \
   --network-plugin=${K8S_NETWORK_PLUGIN} \
+  --v=10 \
+  --container-runtime=${RUNTIME} \
+  --rkt-path=/share/rkt/rkt \
+  --rkt-stage1-image=/share/rkt/stage1-coreos.aci \
```
yifan-gu: Needs to be changed to coreos.com/rkt/stage1-coreos

single-node/user-data
```diff
@@ -88,14 +91,20 @@ function init_templates {
...
 [Service]
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
+ExecStartPre=/usr/bin/touch /etc/hosts
 
 Environment=KUBELET_VERSION=${K8S_VER}
 Environment=KUBELET_ACI=${HYPERKUBE_IMAGE_REPO}
+Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf --volume rktbin,kind=host,source=/usr/bin/rkt --mount volume=rktbin,target=/usr/bin/rkt --volume var-lib-rkt,kind=host,source=/var/lib/rkt --mount volume=var-lib-rkt,target=/var/lib/rkt --volume=stage,kind=host,source=/usr/lib/rkt --mount volume=stage,target=/usr/lib/rkt --volume=mp,kind=host,source=/usr/sbin/modprobe --mount volume=mp,target=/usr/bin/modprobe --volume share,kind=host,source=/share --mount volume=share,target=/share"
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
-ExecStart=/usr/lib/coreos/kubelet-wrapper \
+ExecStart=/share/kubelet-wrapper \
```
euank: I assume the path to these files is part of the WIP bit.

Perhaps for single-node we should have an easy way to pre-populate the share folder or remove it entirely for now and try adding it in a separate commit (since it's really not related to rktnetes)

pbx0: Yes, this hack is already in a separate commit. Github re-ordered them a little. The last few commits are temporary.

Quentin-M:
Bump!

Quentin-M:
Bump @pbx0

pbx0:
@Quentin-M updated the description to match the current state of this PR

Quentin-M:
Thanks.

Quentin-M:
As a side note, we might have to disable rkt-gc when the CNI plugin is being used on the kubelet to let recycle the pods/networks itself.

aaronlevy:
@pbx0 I just merged the v1.3 changes -- would you be able to rebase this to minimize the changeset?

pbx0:
rebased

multi-node/generic/controller-install.sh
```diff
@@ -33,6 +33,9 @@ export DNS_SERVICE_IP=10.3.0.10
 # Whether to use Calico for Kubernetes network policy.
 export USE_CALICO=false
 
+# Determines the container runtime for kubernetes to use. Accepts 'docker' or 'rkt'.
+export RUNTIME=docker
```
euank: nit, `CONTAINER_RUNTIME` seems clearer.

For consistency, should this be `K8S_CONTAINER_RUNTIME` (to match the network-plugin one)?

pbx0: I'll change it to `CONTAINER_RUNTIME`. The `K8S` prefix seems a bit repetitive since we are clearly (to me at least) configuring kubernetes. The use of that prefix already seems inconsistent in my mind anyway.

multi-node/aws/pkg/config/config.go
```diff
@@ -151,7 +153,7 @@ func (c Cluster) Config() (*Config, error) {
 	config.APIServers = fmt.Sprintf("http://%s:8080", c.ControllerIP)
 	config.SecureAPIServers = fmt.Sprintf("https://%s:443", c.ControllerIP)
 	config.APIServerEndpoint = fmt.Sprintf("https://%s", c.ExternalDNSName)
-	if config.UseCalico {
+	if config.UseCalico || config.ContainerRuntime == "rkt" {
```
aaronlevy: Unless there's a specific reason, we should just start always using CNI

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -36,14 +36,21 @@ coreos:
...
       content: |
+        [Unit]
+        Requires=load-rkt-stage1.service
+        After=load-rkt-stage1.service
         [Service]
+        ExecStartPre=/usr/bin/touch /etc/hosts
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf"
+        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf --volume rktbin,kind=host,source=/usr/bin/rkt --mount volume=rktbin,target=/usr/bin/rkt --volume var-lib-rkt,kind=host,source=/var/lib/rkt --mount volume=var-lib-rkt,target=/var/lib/rkt --volume=stage,kind=host,source=/usr/lib/rkt --mount volume=stage,target=/usr/lib/rkt --volume=mp,kind=host,source=/usr/sbin/modprobe --mount volume=mp,target=/usr/bin/modprobe"
```
aaronlevy: Can we make this multi-line?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -36,14 +36,21 @@ coreos:
       command: start
       enable: true
       content: |
+        [Unit]
+        Requires=load-rkt-stage1.service
+        After=load-rkt-stage1.service
         [Service]
+        ExecStartPre=/usr/bin/touch /etc/hosts
```
aaronlevy: I thought this was being populated now: https://github.com/coreos/baselayout/pull/50

multi-node/generic/controller-install.sh
```diff
@@ -56,7 +59,7 @@ function init_config {
         fi
     done
 
-    if [ $USE_CALICO = "true" ]; then
+    if [ $USE_CALICO = "true" ] || [ $RUNTIME = "rkt" ]; then
```
aaronlevy: Should always use CNI unless there's some reason not to

multi-node/generic/controller-install.sh
```diff
@@ -872,8 +947,21 @@ init_flannel
 systemctl stop update-engine; systemctl mask update-engine
 
 systemctl daemon-reload
+
+# TODO(pb): these options together have not been tested. They shouldn't be mutually exclusive options in the future.
```
aaronlevy: Agreed. We should probably just set "CNI Config" based on calico=true (and default to flannel if not). Then it's just a matter of testing that rkt+calico work nicely together.

single-node/user-data
```diff
@@ -85,16 +88,25 @@ function init_templates {
...
 
 Environment=KUBELET_VERSION=${K8S_VER}
 Environment=KUBELET_ACI=${HYPERKUBE_IMAGE_REPO}
+Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf --volume rktbin,kind=host,source=/usr/bin/rkt --mount volume=rktbin,target=/usr/bin/rkt --volume var-lib-rkt,kind=host,source=/var/lib/rkt --mount volume=var-lib-rkt,target=/var/lib/rkt --volume=stage,kind=host,source=/usr/lib/rkt --mount volume=stage,target=/usr/lib/rkt --volume=mp,kind=host,source=/usr/sbin/modprobe --mount volume=mp,target=/usr/bin/modprobe"
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=http://127.0.0.1:8080 \
   --network-plugin-dir=/etc/kubernetes/cni/net.d \
   --network-plugin=${K8S_NETWORK_PLUGIN} \
+  --v=10 \
```
aaronlevy: need to remember to drop these really high log levels before merge

single-node/user-data
```diff
@@ -85,16 +88,25 @@ function init_templates {
...
+  --container-runtime=${RUNTIME} \
+  --rkt-path=/usr/bin/rkt \
```
aaronlevy: So are the outstanding issues just with the stage1 image? Looks like we load them separately (using whatever is in /opt/bin/rkt), but still use /usr/bin/rkt as runtime path.

single-node/user-data
aaronlevy: We should at least bump this to a newer version (v1.3.0) during testing...

multi-node/aws/pkg/config/config.go
pbx0: Added a commit that moves everything to cni and takes away the complication here.

single-node/user-data
pbx0: dropped

multi-node/generic/controller-install.sh
pbx0: defaults to cni now

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: Appears so, removed.

single-node/user-data
pbx0: The fix in rkt 1.10 is for stage1-fly for use with kube-proxy so your right we might not need rkt 1.10 per say on host.... need to test this still.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: done

pbx0:
I think I addressed the comments. Whats left to do:
[ ] Docs
[ ] https://github.com/coreos/coreos-kubernetes/issues/572
[ ] Figure out kube-proxy or ship it as a systemd unit.

robszumski:
@pbx0 What are you thinking for docs? Are we moving everything over? Or is it an additional guide?

pbx0:
@robszumski I'm not entirely sure yet. Updating the docs for `coreos-kubernetes` is fairly straightforward, I just instruct people to set the `CONTAINER_RUNTIME` flag and use a recent CoreOS version. Additionally, the more generic guide about configuring rkt as a runtime on kubernetes mostly just talks about the flags you set on the kubelet. See: https://github.com/coreos/rkt/pull/2918

I don't know what to do about the manual kubernetes on coreos guides. It seems not great to pepper in a bunch what are essentially if statements in the guide that amount to do X if your configuring docker as the runtime, do Y if your configuring rkt as the runtime.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -36,14 +36,31 @@ coreos:
...
+        After=load-rkt-stage1.service
+{{ end }}
         [Service]
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf"
+        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
+          --mount volume=dns,target=/etc/resolv.conf \
+          --volume rktbin,kind=host,source=/usr/bin/rkt \
+          --mount volume=rktbin,target=/usr/bin/rkt \
```
euank: Sooo, how do you feel about the idea of creating a bash script somewhere in the hyperkube container with the contents:

```sh
#!/bin/sh
exec nsenter -m -u -i -n -p -t 1 -- $0 $@
```

and then symlinking it as `/usr/bin/rkt` or whatever matches the exact path on the host?

This would specifically work around the issue of `rkt gc` by allowing the `var-lib-rkt` bindmount to not exist and the `gc` to be run on the host of things.
I recall suggesting it before, but now it seems like we have a concrete reason to reconsider it.

pbx0: I'm ok with it being a temporary hack. 

Long-term it'd be nice if we handled these ongoing issues regarding a privileged container needing various degrees of access to the host all through rkt. Or, in this case, rkt having a recommended way to do rkt-in-rkt that maybe does this behind the scenes or something?

@aaron any thoughts?

euank: A write-api also solves this issue in the longer term, and this is kinda a solution in the same spirit (in that it's just shoving the problem off as "run rkt on the host, not in my app")

pbx0: Also, just a thought, but, at one point we were looking for something called a `naked-stage1` such that the container wasn't in a chroot but rather just a prefix and could call host binaries or container binaries and things would just work. The main problem was sorting out a solution dependent libaries but also making sure the container was happy not living at root.

Given that old idea and the solution you came up with here, what if fly-stage1 just had the option (with a flag) to add binaries from the host to the containers executable path but have them execute with that nsenter command above? It means you have to specify your escape hatches but thats how fly works anyway.

Anyway just a thought, I'm going to prioritize getting the other blockers on this PR fixed and maybe do a followup with your solution to make rkt gc work.

euank:
ref: https://github.com/kubernetes/kubernetes/pull/29310

We might want to cherry-pick that in.

pbx0:
Yes, we will need that cherry-picked. Made a PR against v1.3.2: https://github.com/coreos/kubernetes/pull/62

Quentin-M:
Additionally to the one above, I also cherry-picked https://github.com/kubernetes/kubernetes/pull/29024, required for my usage.

Quentin-M:
Also this: https://github.com/coreos/kubernetes/pull/64

pbx0:
Added a hack to fix the rkt gc issue. Temporary until the write-api stuff gets done. Last outstanding things:

- [ ] docs need review

- [ ] docker cni networking needs to be fixed or backed out to how it was before

Documentation/kubernetes-on-aws-render.md
```diff
@@ -161,6 +161,19 @@ You can now customize your cluster by editing asset files. Any changes to these
...
+The kube-aws tool now optionally supports using rkt as the kubernetes container runtime. To configure rkt as the container runtime you must run CoreOS on the alpha channel and configure the runtime flag.
+
+Edit the `cluster.yaml` file:
+
+```yaml
+containerRuntime: rkt
+releaseChannel: alpha
+```
+
+Note that while using rkt as the runtime is now supported, it is still a new option as of the Kubernetes v1.3 release and may not yet be fully feature compliant.
```
euank: "and has a few [known issues]\(http://kubernetes.io/docs/getting-started-guides/rkt/notes/\)." maybe? (not blocking on docs though, we can get this in a followup)

Documentation/kubernetes-on-aws-render.md
```diff
@@ -251,4 +264,3 @@ If your files are valid, you are ready to [launch your cluster][aws-step-3].
 [aws-step-3]: kubernetes-on-aws-launch.md
 [k8s-openssl]: openssl.md
 [tls-note]: #certificates-and-keys
```
euank: route53 link missing, is that intended or a rebase issue?

Documentation/kubernetes-on-vagrant-single.md
```diff
@@ -44,6 +44,12 @@ $ git clone https://github.com/coreos/coreos-kubernetes.git
 $ cd coreos-kubernetes/single-node/
 ```
 
+## Choose Container Runtime (optional)
+
+The the runtime will default to docker. If you wish to use rkt simply edit the user-data and change the line beginning with `export CONTAINER_RUNTIME` to:
```
euank: s/The the/The/, s/will default to/defaults to/

Documentation/kubernetes-on-vagrant.md
```diff
@@ -62,6 +62,8 @@ The default cluster configuration is to start a virtual machine for each role &m
 
 By default, Calico network policy is disabled. To enable it, change the line `export USE_CALICO=false` to `export USE_CALICO=true` in both the `../generic/controller-install.sh` and the `../generic/worker-install.sh` scripts.
 
+Also by default the container runtime used it docker. To use rkt as the container runtime, change the line `export CONTAINER_RUNTIME=docker` to `export CONTAINER_RUNTIME=rkt` in both the `../generic/controller-install.    sh` and the `../generic/worker-install.sh` scripts. Additionally, make sure the `update_channel` is set to `alpha` in your `config.rb` file.
```
euank: s/used it/used is/. Spacing is off in "controller-install.    sh".

multi-node/generic/worker-install.sh
```diff
@@ -150,6 +209,8 @@ kind: Pod
 metadata:
   name: kube-proxy
   namespace: kube-system
+  annotations:
```
euank: # TODO, once `privileged` works sufficiently well in rkt, switch stage1 back to `stage1-coreos`

Documentation/kubernetes-on-vagrant-single.md
pbx0: done

Documentation/kubernetes-on-aws-render.md
pbx0: rebase issue, added it back

Documentation/kubernetes-on-aws-render.md
pbx0: done (might as well fix docs while I wait on conformance tests)

single-node/user-data
```diff
@@ -802,11 +870,14 @@ EOF
 [Unit]
 Requires=flanneld.service
 After=flanneld.service
+[Service]
+ExecStart=
+ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// \$DOCKER_OPTS \$DOCKER_CGROUPS \$DOCKER_OPT_MTU
```
euank: Rather than modify docker's `ExecStart` (could cause pain if upstream updates theirs), I'd rather just delete the code above that ends up creating the files "/etc/flannel/options.env /run/flannel/options.env" (and the EnvironmentFile reference for the latter).

That way docker will still have $DOCKER_BIP but it will simply expand to being empty.

euank:
Review pass done. I'd like the flannel change I mentioned above if there's not a good reason not to do it, but looks good apart from that.

Documentation/kubernetes-on-vagrant.md
pbx0: done

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -23,6 +23,9 @@ coreos:
             [Unit]
             Requires=flanneld.service
             After=flanneld.service
+            [Service]
+            ExecStart=
+            ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// $DOCKER_OPTS $DOCKER_CGROUPS $DOCKER_OPT_MTU
```
euank: I think individually overriding environment variables is less fragile than removing them by overriding ExecStart.

In this case, I think that would be:

```
Environment=DOCKER_OPT_IPMASQ=""
Environment=DOCKER_OPT_BIP=""
```

pbx0: For whatever reason, this doesn't work. Not sure how to override the envs correctly here.

multi-node/generic/worker-install.sh
pbx0: Lets file an issue instead. I think I made a similar one for the nsenter hack too.

multi-node/aws/pkg/config/templates/cloud-config-controller
euank: Yeah, turns out files override regular, so writing a new file and using it would work, but that won't be as simple as just overriding execstart.
I'm fine with either approach now.

multi-node/generic/worker-install.sh
pbx0: https://github.com/coreos/coreos-kubernetes/issues/606

Documentation/kubernetes-on-aws-render.md
```diff
@@ -161,6 +161,19 @@ You can now customize your cluster by editing asset files. Any changes to these
 [insecure-registry]: https://coreos.com/os/docs/latest/registry-authentication.html#using-a-registry-without-ssl-configured
 [update]: https://coreos.com/os/docs/latest/cloud-config.html#update
 
+### Kubernetes Container Runtime
+
+The kube-aws tool now optionally supports using rkt as the kubernetes container runtime. To configure rkt as the container runtime you must run CoreOS on the alpha channel and configure the runtime flag.
```
aaronlevy: Can we target a minimum release number - alpha is a moving target.

Documentation/kubernetes-on-vagrant.md
```diff
@@ -62,6 +62,8 @@ The default cluster configuration is to start a virtual machine for each role &m
 
 By default, Calico network policy is disabled. To enable it, change the line `export USE_CALICO=false` to `export USE_CALICO=true` in both the `../generic/controller-install.sh` and the `../generic/worker-install.sh` scripts.
 
+Also by default, the container runtime used is docker. To use rkt as the container runtime, change the line `export CONTAINER_RUNTIME=docker` to `export CONTAINER_RUNTIME=rkt` in both the `../generic/controller-install.sh` and the `../generic/worker-install.sh` scripts. Additionally, make sure the `update_channel` is set to `alpha` in your `config.rb` file.
```
aaronlevy: We should target a minimum version (not just alpha). Also, given that this is vagrant -- I wouldn't be opposed to defaulting to alpha.

multi-node/aws/pkg/config/config.go
```diff
@@ -160,8 +162,10 @@ func (c Cluster) Config() (*Config, error) {
 	config.APIServers = fmt.Sprintf("http://%s:8080", c.ControllerIP)
 	config.SecureAPIServers = fmt.Sprintf("https://%s:443", c.ControllerIP)
 	config.APIServerEndpoint = fmt.Sprintf("https://%s", c.ExternalDNSName)
-	if config.UseCalico {
-		config.K8sNetworkPlugin = "cni"
+	config.K8sNetworkPlugin = "cni"
+
+	if config.ContainerRuntime == "rkt" && config.ReleaseChannel != "alpha" {
```
aaronlevy: Not sure if it's possible for us to introspect an actual release version this early (likely not). If it is difficult, ignore this as this is sufficient. But if we can target a release, that would be ideal.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +76,29 @@ coreos:
...
+      command: start
+      content: |
+        [Unit]
+        Description=Load rkt stage1 images
+        Documentation=http://github.com/coreos/rkt
+        Requires=network-online.target
+        After=network-online.target
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
```
aaronlevy: Why is this necessary?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -157,6 +200,13 @@ write_files:
           "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
       done
 
+  - path: /etc/rkthack
```
aaronlevy: nit: `/opt/bin` or `/usr/local/bin`? etc seems like an odd place. Also maybe `rkt-hostns-wrapper` or something like that? rkthack isn't very descriptive -- and it's not a hack -- it's a "root namespace entry solution" or a "system namespace re-entry procedure"

/cc @philips

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: https://github.com/kubernetes/kubernetes/pull/27707

Doing this is currently necessary if we want rkt to cache the stage1 image making everything faster. I'm not sure why this is necessary instead of  it just caching after first use. 

@yifan-gu are there plans to make this unnecessary to run this unit for stage1 caching to work?

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -13,19 +13,39 @@ coreos:
...
+            ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// $DOCKER_OPTS $DOCKER_CGROUPS $DOCKER_OPT_MTU
 
     - name: kubelet.service
       enable: true
       command: start
       content: |
+{{ if eq .ContainerRuntime "rkt" }}
+        [Unit]
+        Requires=docker.service load-rkt-stage1.service
+        After=docker.service load-rkt-stage1.service
```
aaronlevy: Why do we need the docker dependencies? In the case of rkt -- it shouldn't be used. And otherwise it should be socket activated anyway

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -13,19 +13,39 @@ coreos:
...
+{{ end }}
         [Service]
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf --mount volume=dns,target=/etc/resolv.conf"
+        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
+        --mount volume=dns,target=/etc/resolv.conf \
+        --volume=mp,kind=host,source=/usr/sbin/modprobe \
```
aaronlevy: Why do we need this?

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -13,19 +13,39 @@ coreos:
...
+        --mount volume=mp,target=/usr/bin/modprobe \
+        --volume=rkthack,kind=host,source=/etc/rkthack \
+        --mount volume=rkthack,target=/usr/bin/rkt \
+        --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
+        --mount volume=var-lib-rkt,target=/var/lib/rkt \
+        --volume=stage,kind=host,source=/tmp \
+        --mount volume=stage,target=/tmp"
```
aaronlevy: Do we need the host /tmp?

multi-node/aws/pkg/config/config.go
pbx0: Agreed about targeting a minimum version. Will change on docs. As for determining the version in kube-aws, it looks like we can just use the  `pkg/coreosutil.GetAMIData` function and access the version from that. Then import one of the semver packages for a comparison function. I'll add in a separate commit for this to review in a bit.

multi-node/generic/controller-install.sh
```diff
@@ -56,11 +59,7 @@ function init_config {
         fi
     done
 
-    if [ $USE_CALICO = "true" ]; then
-        export K8S_NETWORK_PLUGIN="cni"
-    else
-        export K8S_NETWORK_PLUGIN=""
-    fi
+    export K8S_NETWORK_PLUGIN="cni"
```
aaronlevy: Drop this and add directly to the kubelet flags.

Documentation/kubernetes-on-vagrant.md
pbx0: I believe I updated both vagrant configs to use alpha, but maybe didn't reflect this in the docs. I'll make sure this is the case and remove this step.

multi-node/generic/worker-install.sh
```diff
@@ -45,11 +48,7 @@ function init_config {
         fi
     done
 
-    if [ $USE_CALICO = "true" ]; then
-        export K8S_NETWORK_PLUGIN="cni"
-    else
-        export K8S_NETWORK_PLUGIN=""
-    fi
+    export K8S_NETWORK_PLUGIN="cni"
```
aaronlevy: drop this / add to kubelet flags

single-node/user-data
```diff
@@ -49,11 +52,7 @@ function init_config {
         fi
     done
 
-    if [ $USE_CALICO = "true" ]; then
-        export K8S_NETWORK_PLUGIN="cni"
-    else
-        export K8S_NETWORK_PLUGIN=""
-    fi
+    export K8S_NETWORK_PLUGIN="cni"
```
aaronlevy: drop

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: I believe so,  there was some case where kubelet and rkt essentially did some sort of communication through temp files. @euank

aaronlevy:
@pbx0 done with that review pass. Mostly some minor comments / questions

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: I think I just left in what was already there, then it got removed in master and survived through rebasing or something. Will remove.

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: I believe this is necessary so kubelet can resolv dns based on host settings correctly.  rkt pods launched with kubernetes and host networking will now automatically do this: https://github.com/kubernetes/kubernetes/pull/29024

This won't affect our kubelet-wrapper container though so we do it manually. I think rkt has some kinda `--dns` option too now, there is has been quite a few recent changes in regards to dns in a pod and resolv.conf but it is not clear if we really need to do anything different here.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: I'll change it to `/opt/bin` and `rkt-hostns-wrapper` seems fine to me too. 

The reason I call it out as a hack is because of how we configure this solution rather then the solution itself. I think escaping the chroot is generally the right thing to do when a fly container wants to call host binaries, but i'd like to see us actually support this as some sort of stage1-fly feature flag or something rather then manually adding a bash script to the host specifically for rkt. 

I thought I had made an issue about removing this in the future, but I can't find it now. The relevant rkt issue is: https://github.com/coreos/rkt/issues/2878

pbx0:
@aaronlevy I believe I answered all the questions and the rest of the suggested fixes look good and I'll push up soon.

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: I misinterpreted, you were asking about modprobe not the existing resolv.conf stuff.

https://github.com/kubernetes/kubernetes/blob/f2ddd60eb9e7e9e29f7a105a9a8fa020042e8e52/pkg/kubelet/network/plugins.go#L159

I believe I ran into this ^^ at some point.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: Actually, how about `rkt-chroot-escape-wrapper`? I think its non-obvious to most people that the purpose of this command is to escape the chroot rather then enter the mount namespace (we are already in the host mount namespace).

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: It's a fair amount of complexity to add something to the cache. So this will never be cached unless we do this?

multi-node/aws/pkg/config/templates/cloud-config-worker
aaronlevy: We shouldn't hit that if we're using CNI across the board. The init() function for CNI:

https://github.com/kubernetes/kubernetes/blob/f2ddd60eb9e7e9e29f7a105a9a8fa020042e8e52/pkg/kubelet/network/cni/cni.go#L100

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: It appears I mis-understood. It will be cached after first use and we don't need this.

Telling the kubelet to use a stage1 from a url over a filepath was enough to fix the caching. This service was never necessary, I just copied it from our upstream configs.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: Went with `host-rkt`

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: OK scratch that we need to pre-fetch for this to work at all due to issues with rkt and the `latest` tag that doesn't exist for our stage1 images. Also, the old flag is apparently removed so we can't go back to specifying on disk stage1 and not caching.

multi-node/aws/pkg/config/templates/cloud-config-worker
euank: pod manifests are communicated via tempfiles right now. hardcoding `/tmp/` is only mildly fragile, but should work on coreos just fine

pbx0:
@aaronlevy I've fixed all the comments I believe. 

Mounting in modprobe is removed and conformance tests are running now.  Also sounds like we need to keep the pre-fetching of the stage1 images in for the time being.

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: > Also, the old flag is apparently removed so we can't go back to specifying on disk stage1 and not caching.

Can you provide a bit more details? What old flag?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -36,14 +39,29 @@ coreos:
       command: start
       enable: true
       content: |
+{{ if eq .ContainerRuntime "rkt" }}
+        [Unit]
+        Requires=load-rkt-stage1.service
```
aaronlevy: Do we want to do this every time kubelet starts? Or should we just opportunistically do it once at boot?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -157,6 +198,13 @@ write_files:
           "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
       done
 
+  - path: /opt/bin/host-rkt
```
aaronlevy: Is there an issue we can reference that describes why this is necessary? If not maybe add some comments - I already don't remember.

pbx0: https://github.com/coreos/rkt/issues/2878 Would adding this to the commit message sufficient?

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: Or possibly adding remainAfterExit so this only needs to be executed once?

Really this is just an optimization, so I won't block on this -- just seems heavy-handed to repeatedly do this.

single-node/Vagrantfile
```diff
@@ -25,7 +25,7 @@ Vagrant.configure("2") do |config|
   config.ssh.insert_key = false
 
   config.vm.box = "coreos-%s" % $update_channel
-  config.vm.box_version = ">= 962.0.0"
+  config.vm.box_version = ">= 1109.1.0"
```
aaronlevy: People could still want to change the channel to stable, and this version would currently be too high. Also, elsewhere in this PR this is set at v1122.0.0

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: I think it might help to have a note that this is a workaround to that issue directly in the template would be helpful. Maybe even briefly describe how it works. Looking at that file, that script is not intuitive (couple months from now we may not even remember ourselves).

aaronlevy:
@colhom can you also sign off on the kube-aws side changes?

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: I'm not sure what flag it used to be, euank mentioned the flag that let you specify an on-disk stage one is no longer around. It might just be the case that rather, it is not longer supported.

I just checked this repo's history and at one point I supply an on-disk stage one using the normal flag we use now, but it appears to no longer work with on-disk stage1 images. It is rejected for not being a valid ACI identifier. Not sure if I ever had that working on an older version of kubernetes or not.

single-node/Vagrantfile
pbx0: will do

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: Sounds good, I'll add some comments around the file contents.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: As is, I believe it will just run once per boot and not on kubelet restart.

pbx0:
I believe all comments are addressed up to this point. Added comments to all the host-rkt scripts. The vagrantfiles now specify version 1122.

@colhom you'll probably want to take a look at the commit called `	all: rkt runtime requires CoreOS alpha` and the vendor commit before that.

pbx0:
rebased on master

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -59,6 +69,17 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+{{ if eq .ContainerRuntime "rkt" }}
+    - name: rkt-api.service
+      command: start
+      enable: true
+      content: |
+        [Service]
```
colhom: Would prefer that this is `RequiredBy` and `Before` `kubelet.service`, and still enabled but not explicitly started.

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -43,6 +53,17 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+{{ if eq .ContainerRuntime "rkt" }}
+    - name: rkt-api.service
+      enable: true
+      command: start
+      content: |
+        [Service]
+        ExecStart=/usr/bin/rkt api-service
```
colhom: ditto dependency of `kubelet.service`

multi-node/generic/worker-install.sh
```diff
@@ -262,8 +306,11 @@ systemctl stop update-engine; systemctl mask update-engine
 systemctl daemon-reload
 systemctl enable flanneld; systemctl start flanneld
 systemctl enable kubelet; systemctl start kubelet
+
+if [ $CONTAINER_RUNTIME = "rkt" ]; then
+        systemctl enable rkt-api; systemctl start rkt-api
```
colhom: I would make `rkt-api.service` a dependency of `kubelet.service` here as well.

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -21,6 +21,11 @@ coreos:
       enable: true
       command: start
       content: |
+{{ if eq .ContainerRuntime "rkt" }}
```
colhom: nit: use `RequiredBy/Before` in the `load-rkt-stage1.service` definition, which is already in an equivalent conditional block. Then you can just get rid of this block.

multi-node/aws/pkg/config/config.go
```diff
@@ -164,6 +166,22 @@ func (c Cluster) Config() (*Config, error) {
 	config.APIServerEndpoint = fmt.Sprintf("https://%s", c.ExternalDNSName)
 	config.K8sNetworkPlugin = "cni"
 
+	// Check if we are running CoreOS 1122.0.0 or greater when using rkt as
+	// runtime. Proceed regardless if running alpha. TODO(pb) delete when rkt
+	// works well with stable.
+	if config.ContainerRuntime == "rkt" && config.ReleaseChannel != "alpha" {
```
colhom: Can we just remove the second clause of the condition? The check is super cheap already.

multi-node/vagrant/Vagrantfile
```diff
@@ -8,7 +8,7 @@ require 'yaml'
 
 Vagrant.require_version ">= 1.6.0"
 
-$update_channel = "stable"
+$update_channel = "alpha"
```
colhom: it's kind of odd that we're changing this unilaterally for all Vagrant clusters on behalf of rkt support, but honestly the vagrant stuff is strictly for dev use so alpha is probably preferable anyways

multi-node/aws/pkg/config/config.go
pbx0: I was just guarding against the case in which if we failed to parse or download the version metadata for some reason then this could cause people to fail to start a rkt cluster when clearly, alpha, should never have a version that is too old.

multi-node/vagrant/Vagrantfile
pbx0: Yeah, part of the motivation is that alpha makes sense for Vagrant clusters anyway and it avoids having to write silly version logic for people using rkt.

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: Good idea, will do.

multi-node/aws/pkg/config/templates/cloud-config-worker
pbx0: Will do.

multi-node/generic/worker-install.sh
pbx0: Yup, I'll put the `RequiredBy`/`Before` stuff in both `rkt-api.service` and `load-rkt-stage1.service` for all files and take out the `kubelet.service` Require` and `After` stuff

pbx0:
@colhom addressed your comments regarding expressing `load-rkt-stage1` dependents using `Before` and RequiredBy`

multi-node/aws/pkg/config/config.go
```diff
@@ -160,8 +164,22 @@ func (c Cluster) Config() (*Config, error) {
 	config.APIServers = fmt.Sprintf("http://%s:8080", c.ControllerIP)
 	config.SecureAPIServers = fmt.Sprintf("https://%s:443", c.ControllerIP)
 	config.APIServerEndpoint = fmt.Sprintf("https://%s", c.ExternalDNSName)
-	if config.UseCalico {
-		config.K8sNetworkPlugin = "cni"
+	config.K8sNetworkPlugin = "cni"
```
colhom: Why is this not conditioned on `config.UseCalico==true` anymore?

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: @pbx0 the lexical ordering on the names of the dropin files determines it. First the unit definition is processed, followed by drop-in files in lexical order by file name. 

With `ExecStart` or `Environment` overrides, it's the last (most recently encountered) definition that wins when the unit is evaluated.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,31 @@ coreos:
...
+        ExecStart=/usr/bin/rkt api-service
+        Restart=always
+        RestartSec=10
+
+    - name: load-rkt-stage1.service
+      command: start
```
colhom: In order for `RequiredBy` to work, this unit has to be `enabled:true`. Also get rid of `command: start`, the whole point of `RequiredBy` is that you don't explicitly start dependencies- in the same way you don't explicitly install apt dependencies. 

There's a really important distinction to not munge between "this unit is active as a dependency of these other units" VS "this unit is active because the operator said so (`command: start`)"

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,31 @@ coreos:
...
+      content: |
+        [Unit]
+        Description=Load rkt stage1 images
+        Documentation=http://github.com/coreos/rkt
+        Requires=network-online.target
+        After=network-online.target
+        RequiredBy=kubelet.service rkt-api.service
+        Before=kubelet.service rkt-api.service
```
colhom: For the `RequiredBy/Before` lines, just list rkt-api.service here, and move the `kubelet.service` dependency injection up to `rkt-api.service`.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,31 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+{{ if eq .ContainerRuntime "rkt" }}
+    - name: rkt-api.service
+      command: start
```
colhom: get rid of `command: start` and list `RequiredBy/Before: kubelet.service`

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -40,6 +53,31 @@ coreos:
...
+      content: |
+        [Unit]
+        Description=Load rkt stage1 images
+        Documentation=http://github.com/coreos/rkt
+        Requires=network-online.target
+        After=network-online.target
+        RequiredBy=kubelet.service rkt-api.service
+        Before=kubelet.service rkt-api.service
+        [Service]
+        Type=oneshot
```
colhom: Probably want `RemainAfterExit: yes` here as well

multi-node/aws/pkg/config/config.go
pbx0: See commit: https://github.com/coreos/coreos-kubernetes/pull/551/commits/bc6814694a0ced6b807cad56a06da99993038c37

We can setup both networks through CNI now, so everything uses CNI now.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: ack

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -40,6 +53,31 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+{{ if eq .ContainerRuntime "rkt" }}
+    - name: rkt-api.service
+      enable: true
+      command: start
```
colhom: nix `command: start`

multi-node/aws/pkg/config/config.go
colhom: SWEET!

pbx0:
@colhom Should have addressed last remaining comments.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,37 @@ coreos:
...
+      enable: true
+      content: |
+        [Service]
+        ExecStart=/usr/bin/rkt api-service
+        Restart=always
+        RestartSec=10
+		[Install]
+        RequiredBy=kubelet.service
+        Before=kubelet.service
+        WantedBy=multi-user.target
```
colhom: nix `WantedBy` line. This means "this program should be running if we made it to multi-user mode" which is not what we want to say.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,37 @@ coreos:
...
 
+{{ if eq .ContainerRuntime "rkt" }}
+    - name: rkt-api.service
+      enable: true
+      content: |
+        [Service]
+        ExecStart=/usr/bin/rkt api-service
+        Restart=always
+        RestartSec=10
+		[Install]
```
colhom: fix indent

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,37 @@ coreos:
...
+        Requires=network-online.target
+        After=network-online.target
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
+        [Install]
+        RequiredBy=rkt-api.service
+        Before=rkt-api.service
+        WantedBy=multi-user.target
```
colhom: nix WantedBy

multi-node/generic/controller-install.sh
```diff
@@ -114,6 +122,65 @@ WantedBy=multi-user.target
...
+
+[Service]
+RemainAfterExit=yes
+Type=oneshot
+ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
+
+[Install]
+RequiredBy=rkt-api.service
+Before=rkt-api.service
+WantedBy=multi-user.target
```
colhom: nix

multi-node/generic/controller-install.sh
```diff
@@ -114,6 +122,65 @@ WantedBy=multi-user.target
...
+        cat << EOF > $TEMPLATE
+[Service]
+ExecStart=/usr/bin/rkt api-service
+Restart=always
+RestartSec=10
+
+[Install]
+RequiredBy=kubelet.service
+Before=kubelet.service
+WantedBy=multi-user.target
```
colhom: nix

multi-node/generic/worker-install.sh
```diff
@@ -83,11 +91,69 @@ WantedBy=multi-user.target
...
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
+
+[Install]
+RequiredBy=rkt-api.service
+Before=rkt-api.service
+WantedBy=multi-user.target
```
colhom: nix

multi-node/generic/worker-install.sh
```diff
@@ -83,11 +91,69 @@ WantedBy=multi-user.target
...
+        cat << EOF > $TEMPLATE
+[Service]
+ExecStart=/usr/bin/rkt api-service
+Restart=always
+RestartSec=10
+
+[Install]
+RequiredBy=kubelet.service
+Before=kubelet.service
+WantedBy=multi-user.target
```
colhom: nix

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -40,6 +53,37 @@ coreos:
...
+      enable: true
+      content: |
+        [Service]
+        ExecStart=/usr/bin/rkt api-service
+        Restart=always
+        RestartSec=10
+		[Install]
+        RequiredBy=kubelet.service
+        Before=kubelet.service
+        WantedBy=multi-user.target
```
colhom: nix

multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -40,6 +53,37 @@ coreos:
...
+        Requires=network-online.target
+        After=network-online.target
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
+        [Install]
+        RequiredBy=rkt-api.service
+        Before=rkt-api.service
+        WantedBy=multi-user.target
```
colhom: nix

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: So we should remove this from kubelet.service too then, right?

single-node/user-data
```diff
@@ -103,8 +110,65 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
...
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
 
 [Install]
+RequiredBy=rkt-api.service
+Before=rkt-api.service
+WantedBy=multi-user.target
```
colhom: nix

single-node/user-data
```diff
@@ -103,8 +110,65 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
...
+        cat << EOF > $TEMPLATE
+[Service]
+ExecStart=/usr/bin/rkt api-service
+Restart=always
+RestartSec=10
+
+[Install]
+RequiredBy=kubelet.service
+Before=kubelet.service
 WantedBy=multi-user.target
```
colhom: nix

colhom:
@pbx0 looks good minus the `WantedBy` stuff. Anywhere you have a service that is already `RequiredBy` dependencies, you don't want to say `WantedBy: multi-user.target`.

pbx0:
@colhom Thanks for the massive help fixing my systemd units. But, since kubelet.service has dependent units, we also want to git rid of the multi-user.target thing there as well, right?

multi-node/generic/controller-install.sh
```diff
@@ -114,6 +122,63 @@ WantedBy=multi-user.target
...
+Requires=network-online.target
+After=network-online.target
+
+[Service]
+RemainAfterExit=yes
+Type=oneshot
+ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
+
+[Install]
+RequiredBy=rkt-api.service
```
dghubble: Why not have rkt-api use a `Requires` and `After` under unit? Since these requirements are the rkt-api's expectations.

multi-node/generic/controller-install.sh
```diff
@@ -114,6 +122,63 @@ WantedBy=multi-user.target
...
+        echo "TEMPLATE: $TEMPLATE"
+        mkdir -p $(dirname $TEMPLATE)
+        cat << EOF > $TEMPLATE
+[Service]
+ExecStart=/usr/bin/rkt api-service
+Restart=always
+RestartSec=10
+
+[Install]
+RequiredBy=kubelet.service
```
dghubble: Requires and After `kubelet.service` under unit?

dghubble: er, nvm. It would be a change in the Kubelet unit.

multi-node/generic/controller-install.sh
pbx0: If we do that, then we have to start the rkt-api and start the kubelet. The way we have it now, we just start the kubelet and systemd will start the rkt-api for us.

I think this is right, but, systemd units have been confusing me today. @colhom?

multi-node/generic/controller-install.sh
dghubble: Alright. I see you'd rather it be here because this is a bash script that supports both docker and rkt and you'd rather condition the child units, than the `[Unit]` section of the `kubelet.service`.

pbx0: Right, thats how I had it before, but, the configuration was simplified this way (same kubelet.service for docker and rkt), but also, we only have to start the kubelet unit this way and the api and stage1 thing will be run.

pbx0:
OK, I believe all comments are addressed up to this point and we've settled on the unit stuff.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -56,6 +69,35 @@ coreos:
...
+    - name: rkt-api.service
+      enable: true
+      content: |
+        [Service]
+        ExecStart=/usr/bin/rkt api-service
+        Restart=always
+        RestartSec=10
+        [Install]
+        RequiredBy=kubelet.service
+        Before=kubelet.service
```
colhom: Oops sorry @pbx0, `Before` and `After` both go in the `[Unit]` section, not `[Install]`

colhom:
@pbx0 missed something in my review, @dghubble pointed it out to me. `Before=` directive goes in the `[Unit]` section. Anywhere you have it in `[Install]`, just move it up.

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: Man, I am terrible at following directions regarding systemd units. Fixing now.

colhom:
After that, LGTM from me.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -157,6 +200,21 @@ write_files:
...
       done
 
+  - path: /opt/bin/host-rkt
+    permissions: 0755
+    owner: root:root
+    content: |
+      #!/bin/sh
+      # This is bind mounted into the kubelet rootfs and all rkt shell-outs go
+      # through this rkt wrapper. It essentially enters the host mount namespace
+      # (which is it already in) only for the purpose of breaking out of the chroot
```
derekparker: nit: s/is it/it is/

pbx0: done

multi-node/generic/controller-install.sh
dghubble: Yeah, you can disregard now.

multi-node/aws/pkg/config/templates/cloud-config-controller
dghubble: [Relevant section](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=) for Euan and Colin's points.

> Settings from these EnvironmentFiles override settings made with Environment=. If the same variable is set twice from EnvironmentFiles, the files will be read in the order they are specified and the later setting will override the earlier setting.

pbx0:
OK, now I think we've got the unit file stuff sorted out.

colhom:
Thanks @pbx0 for shepherding this through, great job rktnetes team.

PR #559, opened 2016-06-29
kube-aws: add optional root volume configs

### LINKED ISSUE #554: kube-aws: ebs volumes using old-generation standard magnetic disks
The generated [Cloud Formation template](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/stack-template.json#L258) do not specify which type of disk must be created and attached to each master/worker instance. So, according to the [documentation](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html), the Cloud Formation will create volumes of type `standard`.

One solution would be to set the [volume type](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html#cfn-ec2-blockdev-template-volumetype) property to `"gp2"` (General Purpose SSD), or allow the user to specify the volume type (and possibly the iops) for the controller/worker EBS volumes in `cluster.yaml`.

danielfm:
Uses General Purpose SSD `'gp2'` disks for worker and controller root volumes by default, but allows the user to choose a different volume type, such as the previously used `'standard'`, or `'io1'` for the provisioned IOPS SSD.

If the user set `'io1'` as the volume type, the code also ensures that the IOPS are set to a valid number.

Fixes #554

### DIFF
--- a/multi-node/aws/pkg/config/config_test.go
+++ b/multi-node/aws/pkg/config/config_test.go
@@ -259,6 +259,50 @@ releaseChannel: non-existant #this release channel will never exist
 
 }
 
+func TestAvailabilityZones(t *testing.T) {
+	testCases := []struct {
+		conf string
+		azs  []string
+	}{
+		{
+			conf: singleAzConfigYaml,
+			azs:  []string{"us-west-1c"},
+		},
+		{
+			conf: minimalConfigYaml + `
+# You can specify multiple subnets to be created in order to achieve H/A
+vpcCIDR: 10.4.3.0/16
+controllerIP: 10.4.3.50
+subnets:
+  - availabilityZone: ap-northeast-1a
+    instanceCIDR: 10.4.3.0/24
+  - availabilityZone: ap-northeast-1c
+    instanceCIDR: 10.4.4.0/24
+`,
+			azs: []string{"ap-northeast-1a", "ap-northeast-1c"},
+		},
+	}
+
+	for _, conf := range testCases {
+		confBody := conf.conf
+		c, err := ClusterFromBytes([]byte(confBody))
+		if err != nil {
+			t.Errorf("failed to parse config %s: %v", confBody, err)
+			continue
+		}
+
+		actualAzs := c.AvailabilityZones()
+		if !reflect.DeepEqual(actualAzs, conf.azs) {
+			t.Errorf(
+				"availability zones %s do not match actual list %s in config: %s",
+				conf.azs,
+				actualAzs,
+				confBody,
+			)
+		}
+	}
+}
+
 func TestMultipleSubnets(t *testing.T) {
 
 	validConfigs := []struct {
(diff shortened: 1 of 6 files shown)


multi-node/aws/pkg/config/config_test.go
```diff
@@ -417,3 +417,185 @@ subnets:
...
+			conf: `
+workerRootVolumeType: io1
+workerRootVolumeIOPS: 2000
+`,
+			volumeType: "io1",
+			iops:       2000,
+		},
+	}
+
+	invalidConfigs := []string{
```
colhom: each invalid config string should contain within it a short comment denoting why that config is invalid.

multi-node/aws/pkg/config/templates/stack-template.json
```diff
@@ -256,7 +256,11 @@
           {
             "DeviceName": "/dev/xvda",
             "Ebs": {
-              "VolumeSize": "{{.ControllerRootVolumeSize}}"
+              "VolumeSize": "{{.ControllerRootVolumeSize}}",
+              {{if gt .ControllerRootVolumeIOPS 0}}
```
colhom: the integrity of this templating operation can be covered via unit tests in the `cluster` pkg.

multi-node/aws/pkg/config/config_test.go
danielfm: @colhom Done.

multi-node/aws/pkg/config/templates/stack-template.json
danielfm: @colhom 

> I would also like to see a unit test in the cluster pkg which ensures that the cloudformation resource definitions for the controller instance and worker launch configuration match what is specified in the config.
>
> The way that stack tags are currently tested at the cluster layer is a good example for what you should do here.

I'm not sure how the example you gave me maps to this particular change because the stack tags are sent in its own field in `CreateStackInput` tag (thus much easier to test, as you pointed out), whereas my changes are sent in `TemplateBody` field with the rest of the stack code.

> the integrity of this templating operation can be covered via unit tests in the cluster pkg.

The template already had lots of instructions such as the `{{if ...}}` block you mentioned but from what I saw in the code, there's currently no test to check the result of the rendered stack for these other `if`s.

If I understood you correctly, what you want is a test to validate the rendered stack template in both scenarios (controller/worker iops == 0 and controller/worker iops > 0).

Is that enough?

danielfm: @colhom I just came up with [this](https://github.com/coreos/coreos-kubernetes/pull/559/commits/e320532aef62014c162b9102919d6c3564ba6d9b), so now there's a test to check that volume type and iops options are getting rendered correctly in the final stack template.

I'm not a Golang expert, so there might be a better way to write this code. Let me know if you have a better idea.

danielfm:
Rebased with latest changes in `master` branch.

colhom:
@danielfm much appreciate the effort to test the rendered stack template for the correct volume config. The place to do that is not by re-implementing rendering procedure in `config_test.go`, but instead to test the existing rendering pipeline in `cluster_test.go`.

Do you want to back out  `ee608c7` and I'll send a PR w/ the cluster test stuff to your branch so we can merge this one, or you can take another stab if you like.

colhom:
@danielfm much appreciate the effort to test the rendered stack template for the correct volume config. The place to do that is not by re-implementing rendering procedure in `config_test.go`, but instead to test the existing rendering pipeline in `cluster_test.go`.

Do you want to back out  ee608c7 and I'll send a PR w/ the cluster test stuff to your branch so we can merge this one, or you can take another stab if you like.

danielfm:
@colhom As I mentioned before, I tried to find a way to implement the tests you proposed in `cluster_test.go` (even taking the stack tags tests as a starting point), but I could not see how. Could you give any more pointers on how to implement these tests myself, or it would be easier for you to send the patch yourself?

I know you guys are pretty busy with adding support for Kubernetes 1.3 to kube-aws, so I don't want waste your time with minor issues.

colhom:
@danielfm to be clear- contributing useful code is never a "waste of our time". It's very much appreciated- we're just busy so sometimes it takes us awhile to get around to reviews :(

If you want to give the `cluster_test.go` a shot, the StackTags example is pretty much what you're doing verbatim. Here's [the PR](https://github.com/coreos/coreos-kubernetes/pull/421/commits/e5474c076626a58968b83626aa9684febeea7538) which introduced the cluster tests for stack tags. Remember that all the cluster tests do is submit yaml strings to the config parser and receive dummy API calls as that stack is "created".

If things still aren't clear, back out ee608c7 and then ping me and I can write the tests in a follow-up PR that we merge in quick succession.

danielfm:
@colhom 28683901de1c5172af730bb4ed66adda78560e25. Is that it?

colhom:
@danielfm in the right direction! You still need to assert that the yaml configuration strings do indeed result in the correct cloudformation stack definition.

Two remaining steps here:

* Check out [this struct definition](https://github.com/coreos/coreos-  kubernetes/pull/559#issuecomment-231879356) for stack tag test cases. Notice how a "test case" is the struct

  ```go
  struct {
  		expectedTags []*cloudformation.Tag
		clusterYaml  string
	}{
  ```
  Your test cases are just a list of strings- you're missing the `expectedTags` equivalent that goes along with each cluster yaml string.  For you this would be something like `expectedVolumeConfigs`


* Create another dummy cloudformation service to test volume config. Call this one `testVolumeCloudformationService`. The [existing dummy service](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/cluster/cluster_test.go#L398) used to to test StackTags is a good example. (This should probably be renamed `dummyCloudFormationService --> testStackTagCloudformationService` to avoid confusion)


 In the `CreateStack` method of that `testVolumeCloudformationService` you create, that is where you check that the stack that is being created does indeed have volume configuration that is specified in `cluster.yaml`.

danielfm:
@colhom that was helpful. Thanks!

> In the CreateStack method of that testVolumeCloudformationService you create, that is where you check that the stack that is being created does indeed have volume configuration that is specified in cluster.yaml.

I just see one problem with this approach; the `createStack` function in `cluster.go` receives the already rendered `stackBody` from the caller and just passes it along to the mocked `CreateStack` method, so there's really no way, at least at this point, to verify that the volumes are being created as specified in the yaml file.

danielfm:
@colhom I think I managed to write the tests you asked for.

When you have some time, take a look at 36357d79e0a88228caa8cd2a59f7524b98d19d65 and tell me what you think.

danielfm:
Rebased with latest changes in `master` branch.

danielfm:
Squashed those intermediate commits to clean things up a little bit.

colhom:
@danielfm this looks great! LGTM and thanks.

PR #563, opened 2016-07-05
cloud-config-worker: kubelet.service no longer depends on docker.service

### LINKED ISSUE #545: Suggestion: make kubelet.service NOT depend on docker.service
Currently the systemd units have the following dependencies:

kubelet.service (Requires, After) docker.service (Requires, After) flanneld.service

Ref: https://github.com/coreos/coreos-kubernetes/blob/96e9d1baa781d7f1eb9e362f0c15f9a82c72b5d4/multi-node/aws/pkg/config/templates/cloud-config-worker#L22

There is a nasty failure mode where:

1. flanneld crashes
1. systemd restarts flanneld.service
1. as part of the restart process, systemd stops docker.service and kubelet.service
1. flanneld fails to come up for some reason
1. docker.service and kubelet.service marked as failed for reason 'dependency'
1. systemd restarts flanneld.service again
1. flanneld successfully comes up on second try
1. systemd does not try to start docker.service and kubelet.service

To repro, see my post on systemd-devel: https://lists.freedesktop.org/archives/systemd-devel/2016-June/036897.html

If we don't make kubelet.service depend on docker.service, kubelet wouldn't be stopped in the first place on flanneld.service restart. If kubelet itself crashes, it will always be restarted by systemd independently. If docker.service isn't running when kubelet starts, it will be started via socket activation.

colhom:
The controller [`kubelet.service`]( https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-controller#L35) definition already does without an explicit dependency on docker.

Fixes #545

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-worker
+++ b/multi-node/aws/pkg/config/templates/cloud-config-worker
@@ -18,10 +18,6 @@ coreos:
       enable: true
       command: start
       content: |
-        [Unit]
-        Requires=docker.service
-        After=docker.service
-
         [Service]
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}


pbx0:
lgtm

robszumski:
Need changes to the step by step docs to reflect these changes

colhom:
@robszumski can you link to the docs you're talking about?

robszumski:
Hm interesting, both of these never had it set. I wonder how we got so varied in how we did this...

https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-workers.md#create-the-kubelet-unit
https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-master.md#create-the-kubelet-unit

colhom:
@robszumski when we did the big kube-aws rewrite a few months ago, I was a bit sloppy and re-worked some of the systemd units without updating the step-by-step docs. I bet the docker service dependency was added (probably by me) during that big rewrite.

robszumski:
@colhom gotcha, all that matters is that we're back in sync. 

I checked the Vagrant/generic scripts, they match. Sounds like this is good to go now.

PR #573, opened 2016-07-13
Change default update channel to stable

aaronlevy:


colhom:
LGTM

pbx0:
Rktnetes will need to be on alpha for the time being so either we instruct people to configure alpha in rktnetes docs or attempt to change to alpha whenever runtime is set to rkt.

PR #575, opened 2016-07-15
Improved kubernetes client install instructions for Mac OS

geertschuring:
I noticed an easier way to install the kubernetes cli commands for MacOS and updated the documentation page for running it with Vagrant.

### DIFF
--- a/Documentation/kubernetes-on-vagrant.md
+++ b/Documentation/kubernetes-on-vagrant.md
@@ -21,17 +21,19 @@ The linux `kubectl` binary can be fetched with a command like:
 $ curl -O https://storage.googleapis.com/kubernetes-release/release/v1.3.0-beta.2/bin/linux/amd64/kubectl
 ```
 
-On an OS X workstation, replace `linux` in the URL above with `darwin`:
-
-```sh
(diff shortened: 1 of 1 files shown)


robszumski:
My worry is that this ties us to brew and the versions that it offers. Because Kubernetes is moving so quickly, I worry that brew may lag behind, and users will run into version skew/compatibility issues due to this.

It may reflect badly on this guide and/or CoreOS, because of an external dependency. IMO, using brew doesn't enhance the UX too much, and exposes these potential downsides. Would love to hear other opinions.

bfallik:
@robszumski Sure, here's my $0.02. For my team our docs suggest installing kubectl via brew and then link to the CoreOS docs with a note to ignore the part about manually installing kubectl. Part of our reasoning is that brew is more familiar and trusted than downloading binaries directly.

Also, while the direct link might be more up-to-date initially, over time I think it's much more likely that developers will forget to manually download a newer version, while our developers do typically run `brew update`.

aaronlevy:
I agree that we can't rely on a user having / using brew. curl is a pretty safe assumption as a common tool. 

Also, we would have to rely on the formula being updated regularly (for example, the v1.3.5 release hasn't been added yet: https://github.com/Homebrew/homebrew-core/blob/master/Formula/kubernetes-cli.rb )

PR #584, opened 2016-07-22
all: update hyperkube version to v1.3.2_coreos.0

colhom:
\cc @aaronlevy @pbx0 

Conformance tests look good, after some flakes.


pbx0:
lgtm

PR #592, opened 2016-07-29
fixes cluster not starting post kube-aws up

johnt337:
v0.8.0 would not start until I added the requires.

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -36,6 +36,10 @@ coreos:
       command: start
       enable: true
       content: |
+        [Unit]
+        Requires=docker.service
+        After=docker.service
+
         [Service]
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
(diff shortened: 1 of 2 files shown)


iwarp:
Thanks just tested this and cluster is starting perfectly

cgag:
Ugh.  We removed this in an attempt to fix this issue, but it looks like it didn't help. https://github.com/coreos/coreos-kubernetes/issues/545.  

cc @colhom.

robszumski:
Can confirm that this issue hit my existing cluster. Upon reboot/OS update, the kubelet's depdency chain failed and didn't restart it.

colhom:
The issue has to do with retry limits of one or more of the various __dependencies__ of `kubelet.service`. The way to fix this is go through all service dependencies and make sure each is allowed to retry for a very long time / ad infinitum.

johnt337:
@colhom thanks, i'll give that a go when I have a few cycles. Should I close this pull request in that case?

colhom:
@johnt337 I'll close it out here. When do you think you'll have that PR up?

johnt337:
@colhom if it's still open I'll try and take a look this weekend.

robszumski:
There is a lot of background on (what I believe is) the issue at https://github.com/coreos/coreos-kubernetes/issues/545

johnt337:
Sorry so just getting around to looking into this again.
I see that by default /usr/lib64/systemd/system/flanneld.service is set to `type=notify` and it contains a `--set-env=NOTIFY_SOCKET=/run/systemd/notify \` in it's ExecStart.

Should it be a matter of `Wants=docker.service` to maintain start order during boot time?

PR #597, opened 2016-08-02
docs: fix minor formatting error

robszumski:
Text is expanding out of the page due to single backtick instead of triple.

image


pbx0:
LGTM

PR #555, opened 2016-06-24
v1.3 updates

aaronlevy:
This branch will continue to be updated until there is an official 1.3 release - but want to put it up for other's to start any initial validation / testing

PR #603, opened 2016-08-04
Uses kube-dns v19 instead of v11

zllovesuki:
Fetch from https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns

### DIFF
--- a/Documentation/deploy-addons.md
+++ b/Documentation/deploy-addons.md
@@ -137,7 +137,7 @@ Next, start the DNS add-on:
 $ kubectl create -f dns-addon.yml
 ```
 
-And check for `kube-dns-v11-*` pod up and running:
+And check for `kube-dns-v19-*` pod up and running:
 
 ```sh
 $ kubectl get pods --namespace=kube-system | grep kube-dns-v19
--- a/Documentation/deploy-addons.md
+++ b/Documentation/deploy-addons.md
@@ -39,27 +39,27 @@ spec:
 apiVersion: v1
 kind: ReplicationController
 metadata:
-  name: kube-dns-v19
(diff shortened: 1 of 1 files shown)


aaronlevy:
Complete oversight on my part. Missed updating the documentation in https://github.com/coreos/coreos-kubernetes/pull/555 (where we bumped to v15)

I'd like to keep the addon tracking the current release (rather than master), but looks like that needs a version bump too.

Based on this: https://github.com/kubernetes/kubernetes/blob/v1.3.4/cluster/saltbase/salt/kube-dns/skydns-rc.yaml.base

Looks like version is now v17.1

@zllovesuki would you be willing to make a PR which updates the docs / and installation tools to use this addon version?

zllovesuki:
@aaronlevy For some reasons the emails were not coming through... huh

I was looking at https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/skydns-rc.yaml.in, and that was v19. There's a discrepancy here or am I mistaken>

pbx0:
Looks like @euank has made a similar PR https://github.com/coreos/coreos-kubernetes/pull/607

So how about first person to chime in gets to finish the PR and update the installation tools or defer it to the other person ;) ?

euank:
I made the remaining changes in my PR and picked v17.1 instead of v19 since that was the one in the v1.3.5 release. Hopefully everything looks sane over in that branch. Apologies for doing this after you already had some work, @zllovesuki, I didn't see this PR when I made mine. Thanks for your interest and help :)

pbx0:
#607 is merged. Will update to v19 kubedns likely when v1.4.0 kubernetes is out.

Issue #503, opened 2016-05-24
Add build-release tool


aaronlevy:
We should have a build tool (make/bash/no strong feelings) which consistently builds our release tarballs (which are ultimately linked to the github release page.

/cc @colhom

aaronlevy:
I'm going to keep this open for now, as I think we still need to add some consistency to the build process. e.g. build in a container with specific go version. As it stands the release would be somewhat tied to whatever system it ended up being built on. /cc @colhom

aaronlevy:
An example we could use in this repo (build in rkt container): https://github.com/coreos/bootkube/blob/master/build/build-release.sh

PR #604, opened 2016-08-05
aws: containerized build of release binaries

colhom:
\cc @aaronlevy


colhom:
fixes #503 , minus GPG signing which we can't really automate anyways right now.

aaronlevy:
lgtm

PR #607, opened 2016-08-09
Update to the modern kube-dns; a few other minor tweaks

euank:
Happy to split into two PRs if we want to discuss any of this more, though I think it shouldn't be too bad to just discuss it all here.

cc @joshix @aaronlevy @pbx0

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -455,23 +455,23 @@ write_files:
             "labels": {
               "k8s-app": "kube-dns",
               "kubernetes.io/cluster-service": "true",
-              "version": "v15"
+              "version": "v17.1"
             },
-            "name": "kube-dns-v15",
+            "name": "kube-dns-v17.1",
             "namespace": "kube-system"
           },
           "spec": {
             "replicas": 1,
             "selector": {
               "k8s-app": "kube-dns",
-              "version": "v15"
+              "version": "v17.1"
             },
             "template": {
               "metadata": {
                 "labels": {
                   "k8s-app": "kube-dns",
                   "kubernetes.io/cluster-service": "true",
-                  "version": "v15"
+                  "version": "v17.1"
                 }
               },
               "spec": {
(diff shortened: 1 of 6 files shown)


pbx0:
OK, nice! But two things. First we already have an on-going PR in this vein here: https://github.com/coreos/coreos-kubernetes/pull/603

Its about as far as this PR but, kube-dns 17.1 seems to be the version that matches the v1.3.4 branch we are currently on. So maybe you could coordinate there on who will finish this PR?

Secondly, this is super annoying but we also need the installation tools updated which requires updating:

```
single-node/user-data
multi-node/generic/controller-install.sh
multi-node/generic/worker-install.sh
multi-node/aws/pkg/config/templates/cloud-config-controller
multi-node/aws/pkg/config/templates/cloud-config-worker
```

The first three also actually require you convert the yaml to bash. We normally use:
`python -c "import json,yaml; print json.dumps(yaml.load(open('manifest.yaml')), indent=2, sort_keys=True)"`

euank:
I picked this back up, so I think you owe me some validation @pbx0 :)

I didn't see anything to change in `multi-node/aws/pkg/config/templates/cloud-config-worker` and grepping for `skydns` and `kubedns-amd64:1.3` shows nothing left to change.

pbx0:
Indeed! I have run this through conformance and it looks good. 

LGTM

single-node/user-data
```diff
@@ -448,33 +448,33 @@ EOF
...
-          "version": "v15"
+          "version": "v17.1"
         }
       },
       "spec": {
         "containers": [
           {
             "args": [
-              "--domain=cluster.local.",
+              "--domain=cluster.local",
```
pbx0: I believe preserving this dot is important.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -455,33 +455,33 @@ write_files:
...
-                  "version": "v15"
+                  "version": "v17.1"
                 }
               },
               "spec": {
                 "containers": [
                   {
                     "args": [
-                      "--domain=cluster.local.",
+                      "--domain=cluster.local",
```
pbx0: Same here. I think we need the dot.

euank:
Good catch @pbx0, thanks! Updated to include the `.`.

pbx0:
LGTM

Issue #230, opened 2015-12-30
resize cluster without destroying it first


adamfeuer:
We would like to use `kube-aws` to resize the cluster without first destroying it - for instance, to add more worker nodes by updating the CloudFormation template.

We tried rendering the CloudFormation template and using the `awscli` tool to update the stack, but that didn't work because the SSL cert contents are not rendered into the template when using `kube-aws render`. 

We want to be able to add and remove worker nodes dynamically, and this feature would make it a lot easier.

I'm willing to contribute PR for this if you think this would be helpful.

colhom:
@adamfeuer we are currently working on reproducible cluster assets (#218), which would make dropping new components into an existing CF template much easier.

adamfeuer:
@colhom Well, I'd prefer not to alter the CF template directly - it defeats the purpose of having a higher-level tool like `kube-aws`. But if `kube-aws` would do the altering, then it would help me. Is there some way we can help this effort...?

colhom:
The general flow that I'm envisioning would be:

````
kube-aws render --output ./assets.tgz

#create cluster
kube-aws up ./assets.tgz

#destroy cluster
kube-aws destroy ./assets.tgz

#scale cluster
kube-aws scale ./assets.tgz -worker-count N -controller-count N

```

adamfeuer:
@colhom That flow would work for me. Ideally the assets could be either a tgz or a directory, in case I need to use [BlackBox](https://github.com/StackExchange/blackbox) to secure the files.

xied75:
@adamfeuer If all you want is increase or decrease the number of worker nodes, I think you missed the design point of kube-aws all together. kube-aws uses an aws cloudformation template which in turn uses aws auto scaling group, and the initial template does not even mention the worker nodes at all, it is the job of asg that create those nodes according to the asg LaunchConfiguration, which owns how the aws instance should be created with correct user-data etc.

So after you have a running cluster, you can do this to increase the nodes (let's say init size=2):
```
aws autoscaling update-auto-scaling-group --auto-scaling-group-name kube1-AutoScaleWorker-41W5TOYL3U6A --desired-capacity 4 --min-size 4 --max-size 4
```

You can also set the min and max to a larger range, so you can also call (let's say min=2 max=8):

```
aws autoscaling set-desired-capacity --auto-scaling-group-name kube1-AutoScaleWorker-41W5TOYL3U6A --desired-capacity 6
```

[...]

PR #608, opened 2016-08-09
Cluster upgrades

### LINKED ISSUE #104: Secure access to etcd using TLS
Multi-node setups (e.g. AWS) should [secure access to etcd using TLS](https://github.com/coreos/etcd/blob/master/Documentation/security.md#example-2-client-to-server-authentication-with-https-client-certificates).

colhom:
Complete and working upgrade path for kube-aws clusters, minus the discrete etcd cluster instances.

As part of this, we now have external CA support for TLS asset generation, along with support for allowing user to generate all TLS assets.

Fixes #104 #161 
Depends #544 #596 
Follow up with #465  

Unfortunately does not support upgrading clusters that have already launched. --edit-- by already launched, i mean created by kube-aws code prior to this functionality merging.

@mumoshu I'd like to get your work on node draining on shutdown integrated as well.

\cc @plange @whereisaaron @robszumski @sym3tri @bfallik 

Ref #340 #230 #161

### DIFF
--- a/Documentation/kube-aws-cluster-updates.md
+++ b/Documentation/kube-aws-cluster-updates.md
@@ -1,0 +1,40 @@
+# kube-aws cluster updates
+
+## Types of cluster update
+There are two distinct categories of cluster update.
+
+* **Parameter-level update**: Only changes to `cluster.yaml` and/or TLS assets in `credentials/` folder are reflected. To enact this type of update. Modifications to CloudFormation or cloud-config userdata templates will not be reflected. In this case, you do not have to re-render:
+
+```sh
+kube-aws up --update
+```
+
+* **Full update**: Any change (besides changes made to the etcd cluster- more on that later) will be enacted, including structural changes to CloudFormation and cloudinit templates. This is the type of upgrade that must be run on installing a new version of kube-aws, or more generally when cloudinit or CloudFormation templates are modified:
+
+```sh
+kube-aws render
+git diff # view changes to rendered assets
+kube-aws up --update
+```
+
+## Certificate rotation
+
+The parameter-level update mechanism can be used to rotate in new TLS credentials:
+
+```sh
+kube-aws render --generate-credentials
+kube-aws up --update
+```
+
+## the etcd caveat
+
+There is no solution for hosting an etcd cluster in a way that is easily updateable in this fashion- so updates are automatically masked for the etcd instances. This means that, after the cluster is created, nothing about the etcd ec2 instances is allowed to be updated.
+
+Fortunately, CoreOS update engine will take care of keeping the members of the etcd cluster up-to-date, but you as the operator will not be able to modify them after creation via the update mechanism.
+
+In the (near) future, etcd will be hosted on Kubernetes and this problem will no longer be relevant. Rather than concocting overly complex bandaide, we've decided to "punt" on this issue of the time being.
+
+
+
+
+
--- a/Documentation/kube-aws-cluster-updates.md
+++ b/Documentation/kube-aws-cluster-updates.md
@@ -6,15 +6,16 @@ There are two distinct categories of cluster update.
 * **Parameter-level update**: Only changes to `cluster.yaml` and/or TLS assets in `credentials/` folder are reflected. To enact this type of update. Modifications to CloudFormation or cloud-config userdata templates will not be reflected. In this case, you do not have to re-render:
 
 ```sh
-kube-aws up --update
+kube-aws update
 ```
 
 * **Full update**: Any change (besides changes made to the etcd cluster- more on that later) will be enacted, including structural changes to CloudFormation and cloudinit templates. This is the type of upgrade that must be run on installing a new version of kube-aws, or more generally when cloudinit or CloudFormation templates are modified:
 
 ```sh
-kube-aws render
+kube-aws render stack
+kube-aws render credentials
 git diff # view changes to rendered assets
-kube-aws up --update
+kube-aws update
 ```
 
 ## Certificate rotation
@@ -22,8 +23,8 @@ kube-aws up --update
 The parameter-level update mechanism can be used to rotate in new TLS credentials:
 
 ```sh
-kube-aws render --generate-credentials
-kube-aws up --update
+kube-aws render credentials
+kube-aws update
 ```
 
 ## the etcd caveat
--- a/Documentation/kubernetes-on-aws-render.md
+++ b/Documentation/kubernetes-on-aws-render.md
@@ -67,11 +67,23 @@ There will now be a `cluster.yaml` file in the asset directory. This is the main
 
 ### Render contents of the asset directory
 
-Next, generate a set of cluster assets in the current directory, based on the settings in your `cluster.yaml` file.
+* In the simplest case, you can have kube-aws generate both your TLS identities and certificate authority for you.
 
-```sh
-$ kube-aws render
-```
+  ```sh
+  $ kube-aws render --generate-credentials --generate-ca
+  ```
+
+  This is not recommended for production, but is fine for development or testing purposes.
+
+* It is recommended that you supply your own immediate certificate signing authority and let kube-aws take care of generating the cluster TLS credentials.
+
+  ```sh
+  $ kube-aws render --generate-credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
+  ```
+
+  For more information on operating your own CA, check out this [awesome guide](https://jamielinux.com/docs/openssl-certificate-authority/).
+
+* In certain cases, such as users with advanced pre-existing PKI infrastructure, the operator may wish to pre-generate all cluster TLS assets. In this case, you can run `kube-aws render` and copy in your TLS assets into the `credentials/` folder before running `kube-aws up`.
 
 Here's what the directory structure looks like:
 
@@ -88,6 +100,10 @@ $ tree
 │   ├── ca.pem
 │   ├── worker-key.pem
 │   └── worker.pem
+│   ├── etcd-key.pem
+│   └── etcd.pem
+│   ├── etcd-client-key.pem
+│   └── etcd-client.pem
 ├── kubeconfig
 ├── stack-template.json
 └── userdata
--- a/Documentation/kubernetes-on-aws-render.md
+++ b/Documentation/kubernetes-on-aws-render.md
@@ -70,7 +70,7 @@ There will now be a `cluster.yaml` file in the asset directory. This is the main
 * In the simplest case, you can have kube-aws generate both your TLS identities and certificate authority for you.
 
   ```sh
-  $ kube-aws render --generate-credentials --generate-ca
+  $ kube-aws render credentials --generate-ca
   ```
 
   This is not recommended for production, but is fine for development or testing purposes.
@@ -78,12 +78,12 @@ There will now be a `cluster.yaml` file in the asset directory. This is the main
 * It is recommended that you supply your own immediate certificate signing authority and let kube-aws take care of generating the cluster TLS credentials.
 
   ```sh
-  $ kube-aws render --generate-credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
+  $ kube-aws render credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
   ```
 
   For more information on operating your own CA, check out this [awesome guide](https://jamielinux.com/docs/openssl-certificate-authority/).
 
-* In certain cases, such as users with advanced pre-existing PKI infrastructure, the operator may wish to pre-generate all cluster TLS assets. In this case, you can run `kube-aws render` and copy in your TLS assets into the `credentials/` folder before running `kube-aws up`.
+* In certain cases, such as users with advanced pre-existing PKI infrastructure, the operator may wish to pre-generate all cluster TLS assets. In this case, you can run `kube-aws render stack` and copy in your TLS assets into the `credentials/` folder before running `kube-aws up`.
 
 Here's what the directory structure looks like:
 
@@ -280,7 +280,8 @@ This includes the certificate authority, signed server certificates for the Kube
 After you have completed your customizations, re-render your assets with the new settings:
 
 ```sh
-$ kube-aws render
+$ kube-aws render credentials
+$ kube-aws render stack
 ```
 
 The `validate` command check the validity of your changes to the cloud-config userdata files and the CloudFormation stack description.
--- a/multi-node/aws/README.md
+++ b/multi-node/aws/README.md
@@ -92,7 +92,7 @@ There will now be a cluster.yaml file in the asset directory.
 * In the simplest case, you can have kube-aws generate both your TLS identities and certificate authority for you.
 
   ```sh
-  $ kube-aws render --generate-credentials --generate-ca
+  $ kube-aws render credentials --generate-ca
   ```
 
   This is not recommended for production.
@@ -100,7 +100,7 @@ There will now be a cluster.yaml file in the asset directory.
 * It is recommended that, for production, you supply your own immediate certificate signing authority.
 
   ```sh
-  $ kube-aws render --generate-credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
+  $ kube-aws render credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
   ```
 
   For more information on operating your own CA, check out this [awesome guide](https://jamielinux.com/docs/openssl-certificate-authority/).
--- a/multi-node/aws/README.md
+++ b/multi-node/aws/README.md
@@ -89,9 +89,31 @@ There will now be a cluster.yaml file in the asset directory.
 
 ## Render contents of the asset directory
 
-```sh
-$ kube-aws render
-```
+* In the simplest case, you can have kube-aws generate both your TLS identities and certificate authority for you.
+
+  ```sh
+  $ kube-aws render --generate-credentials --generate-ca
+  ```
+
+  This is not recommended for production.
+
+* It is recommended that, for production, you supply your own immediate certificate signing authority.
+
+  ```sh
+  $ kube-aws render --generate-credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
+  ```
+
+  For more information on operating your own CA, check out this [awesome guide](https://jamielinux.com/docs/openssl-certificate-authority/).
+
+* In certain cases, such as users with advanced pre-existing PKI infrastructure, you may wish to pre-generate all cluster TLS assets. In this case, make sure the file tree below exists in your cluster assets directory before running `kube-aws up`.
+
+  ```sh
+  ls -R credentials/
+  credentials/:
+  admin-key.pem  apiserver-key.pem  ca.pem               etcd-client.pem  etcd.pem        worker.pem
+  admin.pem      apiserver.pem      etcd-client-key.pem  etcd-key.pem     worker-key.pem
+  ```
+
 
 This generates the default set of cluster assets in your asset directory. These assets are templates and credentials that are used to create, update and interact with your Kubernetes cluster.
 
@@ -161,6 +183,10 @@ It can take some time after `kube-aws up` completes before the cluster is availa
 ```sh
 $ kube-aws up --export
 ```
+
+## Update an existing kube-aws cluster
+
+Read the [cluster update](../../Documentation/kube-aws-cluster-updates.md) documentation.
 
 ## Development
 
--- a/multi-node/aws/cmd/kube-aws/command_render.go
+++ b/multi-node/aws/cmd/kube-aws/command_render.go
@@ -7,7 +7,13 @@ import (
 	"os"
 	"text/template"
 
+	"crypto/rsa"
+	"crypto/x509"
+
+	"path"
+
 	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/config"
+	"github.com/coreos/coreos-kubernetes/multi-node/aws/pkg/tlsutil"
 	"github.com/spf13/cobra"
 )
 
@@ -19,10 +25,20 @@ var (
 		RunE:         runCmdRender,
 		SilenceUsage: true,
 	}
+	renderOpts = struct {
+		generateCredentials bool
+		generateCA          bool
+		caKeyPath           string
+		caCertPath          string
+	}{}
 )
 
 func init() {
 	cmdRoot.AddCommand(cmdRender)
+	cmdRender.Flags().BoolVar(&renderOpts.generateCredentials, "generate-credentials", false, "generate new cluster TLS assets")
+	cmdRender.Flags().BoolVar(&renderOpts.generateCA, "generate-ca", false, "if generating credentials, generate root CA key and cert. NOT RECOMMENDED FOR PRODUCTION USE- use '-ca-key-path' and '-ca-cert-path' options to provide your own certificate authority assets")
+	cmdRender.Flags().StringVar(&renderOpts.caKeyPath, "ca-key-path", "./credentials/ca-key.pem", "path to pem-encoded CA RSA key")
+	cmdRender.Flags().StringVar(&renderOpts.caCertPath, "ca-cert-path", "./credentials/ca.pem", "path to pem-encoded CA x509 certificate")
 }
 
 func runCmdRender(cmd *cobra.Command, args []string) error {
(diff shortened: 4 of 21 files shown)


pieterlange:
This is awesome! I will have to make some time to test this (along with the HA stuff).

bfallik:
Looking forward to testing this!

colhom:
@bfallik once #465 is rebased on this PR, you'll have pods drained off nodes before they shutdown and are destroyed.

Cluster upgrades are entirely functional here, but keep in mind that the nodes will be shutdown ungracefully, and consequently requests will be routed to the pod ips for some amount of time after the containers have disappeared.

robszumski:
At a high level, I'd like to think more about these commands and flags.

### Current (this PR)

The current method in this PR:

```sh
$ kube-aws render
$ git diff # view changes to rendered assets
$ kube-aws up --update
 ```

```sh
$ kube-aws render --generate-credentials
$ kube-aws up --update
```

This retains the two primary commands that we are used to, but makes them much more complicated. The `up` really doesn't do the same thing as before, where the user was taught that it brings up a complete stack. Now it just modifies the stack _or_ creates a new stack, based on this flag.

## Proposal

I propose changing these names. Here are the same scenarios:

```sh
$ kube-aws render stack
$ git diff # view changes to rendered assets
$ kube-aws update stack
 ```

```sh
$ kube-aws render credentials
$ kube-aws update credentials
```

Note the use of the same subcommand for each. Makes it easier to teach you the terms and pieces that are involved.

This PR does a great job of separating out the `render` vs `update` part, this retains that and makes it even more explicit.

### Backwards Compatible

For backwards compatibility, we can alias (but not document) the render command from the last release:

```sh
$ kube-aws render         # v0.8.1
$ kube-aws render stack   # master
```

Documentation/kube-aws-cluster-updates.md
```diff
@@ -0,0 +1,40 @@
+# kube-aws cluster updates
```
robszumski: To fit the naming scheme, can we name this doc `kubernetes-on-aws-updates.md`?

Documentation/kube-aws-cluster-updates.md
```diff
@@ -0,0 +1,40 @@
...
+## Certificate rotation
+
+The parameter-level update mechanism can be used to rotate in new TLS credentials:
+
+```sh
+kube-aws render --generate-credentials
+kube-aws up --update
+```
+
+## the etcd caveat
```
robszumski: I vote removing this section entirely, and creating a new intro to this doc that basically says:

 - this updates Kubernetes software releases
 - this updates credentials used
 - manual changes to your user-data is applied but not tested in any way
 - does not cover updating the etcd cluster yet

colhom:
@robszumski great suggestion! i'll think it over more thoroughly while implementing it, but sgtm and I'll move forward with what you have outlined. I was unsure of what to do with the command tree... thanks for figuring it out.

Documentation/kube-aws-cluster-updates.md
```diff
@@ -0,0 +1,40 @@
+# kube-aws cluster updates
+
+## Types of cluster update
+There are two distinct categories of cluster update.
+
+* **Parameter-level update**: Only changes to `cluster.yaml` and/or TLS assets in `credentials/` folder are reflected. To enact this type of update. Modifications to CloudFormation or cloud-config userdata templates will not be reflected. In this case, you do not have to re-render:
```
cgag: "To enact this type of update."?

Documentation/kube-aws-cluster-updates.md
```diff
@@ -0,0 +1,40 @@
...
+kube-aws up --update
+```
+
+## Certificate rotation
+
+The parameter-level update mechanism can be used to rotate in new TLS credentials:
+
+```sh
+kube-aws render --generate-credentials
+kube-aws up --update
```
cgag: The docs above made it sound like a parameter-level update is distinguished by not requiring a render, but we're rendering here?

Maybe generate-credentials should be its own command?  Maybe there's no need to distinguish between the methods but just note the special cases where it is required?  I believe re-rendering is always safe, if so, maybe we should just always recommend rendering?

I just feel like this notion of two methods of updating with the same command is confusing.

Documentation/kube-aws-cluster-updates.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+## the etcd caveat
+
+There is no solution for hosting an etcd cluster in a way that is easily updateable in this fashion- so updates are automatically masked for the etcd instances. This means that, after the cluster is created, nothing about the etcd ec2 instances is allowed to be updated.
```
cgag: I could just be ignorant but "masked" to me feels like systemd jargon, I'd rather just say disabled unless there's a technical reason not to.

Documentation/kube-aws-cluster-updates.md
```diff
@@ -0,0 +1,40 @@
...
+
+Fortunately, CoreOS update engine will take care of keeping the members of the etcd cluster up-to-date, but you as the operator will not be able to modify them after creation via the update mechanism.
+
+In the (near) future, etcd will be hosted on Kubernetes and this problem will no longer be relevant. Rather than concocting overly complex bandaide, we've decided to "punt" on this issue of the time being.
```
cgag: band-aid (®)

multi-node/aws/README.md
```diff
@@ -89,9 +89,31 @@ There will now be a cluster.yaml file in the asset directory.
...
+
+* In certain cases, such as users with advanced pre-existing PKI infrastructure, you may wish to pre-generate all cluster TLS assets. In this case, make sure the file tree below exists in your cluster assets directory before running `kube-aws up`.
+
+  ```sh
+  ls -R credentials/
+  credentials/:
+  admin-key.pem  apiserver-key.pem  ca.pem               etcd-client.pem  etcd.pem        worker.pem
+  admin.pem      apiserver.pem      etcd-client-key.pem  etcd-key.pem     worker-key.pem
+  ```
+
```
cgag: Is this deliberately different from `Documentation/kubernetes-on-aws-render.md`?  That one uses tree, and has slightly different language around not being production ready and being ok for testing.

multi-node/aws/README.md
```diff
@@ -162,6 +184,10 @@ It can take some time after `kube-aws up` completes before the cluster is availa
 $ kube-aws up --export
 ```
```
cgag: not related to this commit but can we just make a kube-aws export command?

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -32,16 +48,45 @@ func runCmdRender(cmd *cobra.Command, args []string) error {
...
+				return fmt.Errorf("failed generating cluster CA: %v", err)
+			}
+			fmt.Printf("-> Generating new TLS CA\n")
+		} else {
+			fmt.Printf("-> Parsing existing TLS CA\n")
+			if caKeyBytes, err := ioutil.ReadFile(renderOpts.caKeyPath); err != nil {
+				return fmt.Errorf("failed reading ca key file %s : %v", renderOpts.caKeyPath, err)
+			} else {
+				if caKey, err = tlsutil.DecodePrivateKeyPEM(caKeyBytes); err != nil {
+					return fmt.Errorf("failed parsing ca key: %v", err)
```
cgag: Why is this in an else block?

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -32,16 +48,45 @@ func runCmdRender(cmd *cobra.Command, args []string) error {
...
+				}
+			}
+			if caCertBytes, err := ioutil.ReadFile(renderOpts.caCertPath); err != nil {
+				return fmt.Errorf("failed reading ca cert file %s : %v", renderOpts.caCertPath, err)
+			} else {
+				if caCert, err = tlsutil.DecodeCertificatePEM(caCertBytes); err != nil {
```
cgag: also seems like an unnecessary `else`

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -32,16 +48,45 @@ func runCmdRender(cmd *cobra.Command, args []string) error {
...
+		fmt.Printf("Generating TLS credentials...\n")
+		var caKey *rsa.PrivateKey
+		var caCert *x509.Certificate
+		if renderOpts.generateCA {
+			var err error
+			caKey, caCert, err = config.NewTLSCA()
+			if err != nil {
+				return fmt.Errorf("failed generating cluster CA: %v", err)
+			}
+			fmt.Printf("-> Generating new TLS CA\n")
```
cgag: seems like this should be at the top of this block

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -32,16 +48,45 @@ func runCmdRender(cmd *cobra.Command, args []string) error {
...
+		}
+		fmt.Printf("-> Generating new TLS assets\n")
+		assets, err := cluster.NewTLSAssets(caKey, caCert)
+		if err != nil {
+			return fmt.Errorf("Error generating default assets: %v", err)
+		}
+		if err := os.MkdirAll("credentials", 0700); err != nil {
+			return err
+		}
+		if err := assets.WriteToDir("./credentials", renderOpts.generateCA); err != nil {
```
cgag: Is the ./ necessary?   It'd be nice to either add it to the MkdirAll call above or remove it here.

multi-node/aws/cmd/kube-aws/command_render.go
```diff
@@ -32,16 +48,45 @@ func runCmdRender(cmd *cobra.Command, args []string) error {
...
-		return fmt.Errorf("Error generating default assets: %v", err)
-	}
-	if err := os.Mkdir("credentials", 0700); err != nil {
-		return err
-	}
-	if err := assets.WriteToDir("./credentials"); err != nil {
-		return fmt.Errorf("Error create assets: %v", err)
+	if renderOpts.generateCredentials {
+		fmt.Printf("Generating TLS credentials...\n")
+		var caKey *rsa.PrivateKey
```
cgag: I don't think these should be pointers, or that NewTLSAssets should accept pointers.  It doesn't look like there's any reason for these to be nullable, and it looks like this problem goes all the way down to tlsutil, where eventually x509.CreateCertificate needs a pointer.  We should just take an address at that point rather than propagating the pointer type all the way up here.

multi-node/aws/cmd/kube-aws/command_up.go
```diff
@@ -26,7 +26,7 @@ var (
 func init() {
 	cmdRoot.AddCommand(cmdUp)
 	cmdUp.Flags().BoolVar(&upOpts.export, "export", false, "Don't create cluster, instead export cloudformation stack file")
-	//	cmdUp.Flags().BoolVar(&upOpts.update, "update", false, "update existing cluster with new cloudformation stack")
+	cmdUp.Flags().BoolVar(&upOpts.update, "update", false, "update existing cluster with new cloudformation stack")
```
cgag: The other two flags start with capital letters.  I believe the rest of the flags in the project are all lowercase.  Looks like none use punctuation.

We should decide on what flags should look like and make them consistent (ideally across all our projects).  I'd vote for capitalizing without punctuation by way of cargo-culting cargo/rustc.  Thoughts @robszumski ?

multi-node/aws/pkg/cluster/cluster.go
```diff
@@ -272,21 +359,45 @@ func (c *Cluster) Update(stackBody string) (string, error) {
...
+			},
+			PageSize: aws.Int64(2),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("error describing load balancer %s: %v", elbName, err)
+		}
+		if len(resp.LoadBalancerDescriptions) == 0 {
+			return nil, fmt.Errorf("could not find a load balancer with name %s", elbName)
+		}
+		if len(resp.LoadBalancerDescriptions) > 1 {
```
cgag: Is this actually possible?

multi-node/aws/pkg/config/config.go
```diff
@@ -47,9 +47,13 @@ func newDefaultCluster() *Cluster {
...
 		WorkerRootVolumeIOPS:     0,
 		WorkerRootVolumeSize:     30,
+		EtcdCount:                1,
+		EtcdInstanceType:         "m3.medium",
+		EtcdRootVolumeSize:       30,
+		EtcdDataVolumeSize:       30,
 		CreateRecordSet:          false,
 		RecordSetTTL:             300,
-		Subnets:                  []Subnet{},
+		Subnets:                  []*Subnet{},
```
cgag: Why the switch to pointers?

Documentation/kube-aws-cluster-updates.md
colhom: yeah, should probably add that.

Documentation/kube-aws-cluster-updates.md
colhom: Yeah, agreed. Rob suggests that we split this out into `kube-aws update credentials` and `kube-aws render  credentials` which would have the nice side effect of solving this problem.

multi-node/aws/pkg/config/config.go
```diff
@@ -156,9 +166,13 @@ var supportedReleaseChannels = map[string]bool{
...
 	config := Config{Cluster: c}
-	config.ETCDEndpoints = fmt.Sprintf("http://%s:2379", c.ControllerIP)
-	config.APIServers = fmt.Sprintf("http://%s:8080", c.ControllerIP)
-	config.SecureAPIServers = fmt.Sprintf("https://%s:443", c.ControllerIP)
+
+	config.MinWorkerCount = config.WorkerCount - 1
+	config.MaxWorkerCount = config.WorkerCount + 1
+
+	config.MinControllerCount = config.ControllerCount - 1
+	config.MaxControllerCount = config.ControllerCount + 1
```
cgag: Could put all these unconditional pieces of config into the Config literal above.

multi-node/aws/README.md
colhom: We're kind of in this weird place where the two docs have duplicated a lot of the same content.

multi-node/aws/README.md
colhom: file an issue? i think i agree.

multi-node/aws/pkg/config/config.go
```diff
@@ -181,20 +195,93 @@ func (c Cluster) Config() (*Config, error) {
 		config.VPCRef = fmt.Sprintf("%q", config.VPCID)
 	}
 
+	config.EtcdInstances = make([]etcdInstance, config.EtcdCount)
+	var etcdEndpoints, etcdInitialCluster bytes.Buffer
+	for etcdIndex := 0; etcdIndex < config.EtcdCount; etcdIndex++ {
+
+		//Round-robbin etcd instances across all available subnets
```
cgag: robin

multi-node/aws/pkg/config/config.go
```diff
@@ -181,20 +195,93 @@ func (c Cluster) Config() (*Config, error) {
...
+		//This is complicated, must also factor in DHCP addresses
+		//for ASG components
+
+		//Punt on this- we're going to have an answer for dynamic etcd clusters at some point. Then we can either throw
+		//the instances in an ASG and use DHCP like all other instances, or simply self-host on cluster
+
+		config.EtcdInstances[etcdIndex] = instance
+
+		//TODO: ipv6 support
+		if len(instance.IPAddress) != 4 {
```
cgag: I'd prefer to validate this before creating the etcdInstance with it, it reads stranely to me to have this after adding this to `config.EtcdInstances`.

multi-node/aws/pkg/config/config.go
```diff
@@ -181,20 +195,93 @@ func (c Cluster) Config() (*Config, error) {
...
+			return nil, fmt.Errorf("Non ipv4 address for etcd node: %v", instance.IPAddress)
+		}
+
```
cgag: Can we avoid all this by just using the actual private ip instead of this DNS name derived from the private ip?

multi-node/aws/pkg/config/config.go
```diff
@@ -181,20 +195,93 @@ func (c Cluster) Config() (*Config, error) {
...
+			instance.IPAddress[1],
+			instance.IPAddress[2],
+			instance.IPAddress[3],
+			dnsSuffix,
+		)
+
+		fmt.Fprintf(&etcdEndpoints, "https://%s:2379", hostname)
+		fmt.Fprintf(&etcdInitialCluster, "%s=https://%s:2380", hostname, hostname)
+		if etcdIndex < config.EtcdCount-1 {
+			fmt.Fprintf(&etcdEndpoints, ",")
```
cgag: I think it'd be clearer to just make `etcdEndpoints` a []string and then strings.Join(etcdEndpoints, ",")

multi-node/aws/cmd/kube-aws/command_up.go
robszumski: Note that @cgag is talking about flag _descriptions_, not the flags themselves. 

Going lowercase sounds good to me and seems to be what a lot of other CLIs do.

Documentation/kubernetes-on-aws-render.md
```diff
@@ -67,11 +67,23 @@ There will now be a `cluster.yaml` file in the asset directory. This is the main
...
+  ```sh
+  $ kube-aws render --generate-credentials --generate-ca
+  ```
+
+  This is not recommended for production, but is fine for development or testing purposes.
+
+* It is recommended that you supply your own immediate certificate signing authority and let kube-aws take care of generating the cluster TLS credentials.
+
+  ```sh
+  $ kube-aws render --generate-credentials --ca-cert-path=/path/to/ca-cert.pem --ca-key-path=/path/to/ca-key.pem
```
sym3tri: yay!

sym3tri: Also might be nice to add a required flag here `--certificate-expires` so users _must_ enter a value and will be made aware that the certs will expire in x days so there are no surprises later.

Documentation/kube-aws-cluster-updates.md
sym3tri: maybe some text below this to explain at a high-level what this actually does

multi-node/aws/pkg/config/config.go
colhom: yeah that would be more concise.

Documentation/kubernetes-on-aws-render.md
chancez: Agree with @sym3tri , but maybe make it a config file value instead?

colhom:
closes #522

multi-node/aws/pkg/cluster/cluster.go
colhom: If it happens, the user should know. APIs make mistakes.

multi-node/aws/pkg/config/config.go
colhom: IP SANS is a hack, CN / Subject Alternate Name is preferable. The fact that the DNS name is derived from IP is incidental.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -91,17 +93,18 @@ coreos:
       content: |
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
-        Before=kubelet.service
-        After=docker.service
-        Requires=docker.service
+        Before=kubelet.service flanneld.service
```
cgag: Do we need this when flanneld after/requires decrypt already?  Should we add an After=decrypt to the kubelet? 

Looking around I see a mixture of use of RequiredBy vs Required.  I think they're equivalent and having units specify their dependencies makes more sense than a units specify who depends on them.  Am I wrong?  If not I'd like to standardize on Requires/After.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -176,7 +179,7 @@ write_files:
 
       for encKey in $(find /etc/kubernetes/ssl/*.pem.enc);do
         tmpPath="/tmp/$(basename $encKey).tmp"
-        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext | base64 --decode > $tmpPath
+        docker run --net host --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext | base64 --decode > $tmpPath
```
cgag: Why did this need to be added?

multi-node/aws/pkg/config/templates/cluster.yaml
```diff
@@ -72,6 +72,31 @@ kmsKeyArn: "{{.KMSKeyARN}}"
...
 # workerSpotPrice: "0.05"
 
+## Etcd Cluster config
+## WARNING: Any changes to etcd parameters after the cluster is first created will not be applied
+## during a cluster upgrade, due to concerns over data loss.
+## This situation is being rectified with work towards automated management of etcd clusters
+
+# Number of etcd nodes
+# (Set to an odd number >= 3 for HA control plane)
+# etcdCount: 1
```
cgag: I'd expect to be modifying controllerCount for HA control plane.  I actually don't see controllerCount in  cluster.yaml at all actually.

colhom: Two things:

- There's a discrete etcd cluster and controllers in an ASG, so `etcdCount` and `controllerCount` are entirely decoupled now. TLDR; you can modify controller count, etcdCount is set-on-create then no update

- I forgot to doc it in `cluster.yaml`, but it's exposed by the `Cluster` object. I'll fix that up.

wirtsi: Will this also enable changing the cloud-config-* files and redeploying the changes through a new compiled CloudFormation template? If so, then this is awesome :)

colhom: @wirtsi yep!

pieterlange:
Enabling either calico or using your own VPC in `cluster.yaml` breaks template generation with a very useless error. I'll come back with better details soon. :)

colhom:
@pieterlange impossible! This software is bug-free!

colhom:
@pieterlange any news on the calico problem you encountered with this PR?

colhom:
@pieterlange I've cherry-picked in your commit

colhom:
@robszumski check out b4d05dc

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: AWS metadata service is a special IP address on the host network.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Dependency injection is a distinct pattern, the inverse of traditional `Requires` dependency listing. Desiring to standardize on one or the other is like requiring a road-system only have right-hand turns for simplicity sake (cough new jersey).

A textbook use-case for dependency injection (`RequiredBy`) is pluggable dependencies (aka providers). The kubelet needs a "network provider", that can be flannel, weave, AWS sdk, etc. Defining and enabling the `flanneld.service` will transparently inject `flanneld.service` as a dependency of the kubelet, which is much preferable to mutating `kubelet.service` every time you want to change network providers.

robszumski:
> @robszumski check out b4d05dc

Nice, lookin' good!

multi-node/aws/pkg/config/templates/cloud-config-controller
cgag: Fair enough, I didn't consider that use case, though it's not like we have to outright ban anything.  Just preferring one unless there's a good reason to prefer the other would be fine.  UPS makes something like 90% right turns. 

Assuming I'm sold on this being better than mutating the kubelet dependency, I think we could get the same dependency injection effect with more clarity by using targets.  Kubelet should require something like network-provider.target and then having flanneld's RequiredBy set to the target.  This seems to be the pattern I see poking around /etc/systemd/system on my local arch install.

I just really dislike looking at a service and it not being obvious if it has any dependencies. 

Regardless of what whether or not we're following this injection pattern, it doesn't look like we're being consistent either way since flannel explicitly Requires decrypt, while kubelet doesn't.

iwarp:
How far away is this from being merged? Im keen to start using this.

colhom:
The last two commits are this PR. Prior two are for #596 and #544

colhom:
@iwarp we're working on getting this code reviewed! Sorry for the delay

If you're really keen to start using it, it should all be functional if you pull from `colhom:cluster-upgrades`.

colhom:
Note to self- I also need to add `UpdatePolicy` stanzas to to `Subnets` and `VPC` prohibiting updates to them . To update a subnet or vpc with cloudformation, the whole deployment is essentially replicated in a different availability zone. Kube-aws in general will not be able to support this in the near future.

pieterlange:
I've been using the `colhom:cluster-upgrades` branch (plus some minor patches for an external etcd cluster) for a little over a week now and it works great. Updating the stack works as expected!

Minor note: the update-policy for the worker autoscaler might need a little increase from the default 2 minutes depending on app startup time. The kubernetes master also has a brief window where it's unavailable but everything recovers just fine

sasso:
Any updates on this? Would love to start using it 🚀

colhom:
An update for all interested parties:

We'll be merging this functionality (along with some of @mumoshu 's work regarding node draining) in an `experimental` branch in the near future. Work is encouraged in that direction, though the officially released code in master will not receiving this functionality. Going forward, the goal is to orchestrate these critical behaviors via the Kubernetes control plane, rather than via CloudFormation.

mumoshu:
@colhom Thanks for the update!
Does it mean that you and your colleagues won't be focusing on things in `experimental` branch anymore?
(Btw, in the longterm, I agree with the goal you've mentioned)

camilb:
@colhom 
Have a working version based on #608, #629 and the latest changes from the master. 
At the moment I'm running the tests using :

- 3 Controllers in Multi-AZ with LoadBalancer

- 3 external ETCD nodes configured with SSL  in Multi-AZ

- 3  Workers  in Multi-AZ

- CNI

- Hyperkube v1.4.0_coreos.2

Tested using the latest stable and alpha OS releases.

For my current setup this works pretty good. Next I will try to put ETCD in a Auto Scaling Group with S3 daily backups.

If there is someone interested in it, I have a working branch: 
https://github.com/camilb/coreos-kubernetes/tree/1.4.0-ha

iwarp:
Hmmm interesting change of direction. What's the guidance for a highly available cluster that i should be using right now then? I was planning that this PR was going to be complete before going live on a new project 

Do i need to create multiple k8s clusters and load balance across which is closer to the k8s federation approach.

How have others approached this?

apenney:
Echoing the previous response.  I put off deploying kubernetes until this PR was finished but now I find myself unsure how to proceed.  I might just look at kops at this point until there's a clearer vision for coreos-kubernetes.  We looked at enterprise support for coreos but this project was a blocker for us being able to proceed with that (in case it helps justify anyone spending time on laying out a clear roadmap).

sasso:
I was pushing the date to deploy coreos kube to production and waiting for this PR to land master. We would like to have a procedure to do future updates and this seemed as a good solution. Any plans on providing some guidance on how updates will be done with future new releases?

PR #611, opened 2016-08-11
vagrant/generic: use the Ruby OpenSSL library to generate certificates

### LINKED ISSUE #157: [Windows] single-node vagrant up fails
after cloning the repo and running the `vagrant up` command 

```    
PS C:\www\coreos-kubernetes\single-node> vagrant up    
'.' is not recognized as an internal or external command,    
operable program or batch file.    
failed generating SSL CA artifacts
```

a directory list after the batch file runs looks like  

```
d-----       2015-11-19   5:27 PM                -p   
d-----       2015-11-19   5:07 PM                .vagrant   
d-----       2015-11-19   5:07 PM                ssl   
-a----       2015-11-19   5:07 PM             14 .gitignore   
-a----       2015-11-19   5:07 PM            297 README.md   
-a----       2015-11-19   5:07 PM          15833 user-data   
-a----       2015-11-19   5:07 PM           2189 Vagrantfile   
```

ah45:
The Vagrant single and multi node setups automatically generate SSL
certificates via shell scripts and use of the `/usr/bin/openssl`
command. This fails on non *nix platforms, i.e. Windows.

Vagrant bundles Ruby and includes the Ruby OpenSSL library which can
be used to generate the certificates in a safe, cross platform, manner.

Fixes #157


aaronlevy:
@ah45 sorry for the very late reply.

I don't think we can merge this for a couple reasons:

We don't (in house) have much ruby expertise -- so maintaining this could become problematic.

And, we're also trying to move away from externally generated TLS assets - and instead make use of internal k8s pki tooling (see: https://github.com/kubernetes/features/issues/43 for feature overview).

The above should be available (in its initial implementation) as of v1.4 - which should ship towards the end of September.

The end state would be that a user needs to provide a CA keypair -- which we can reasonably document across different environments.

PR #623, opened 2016-08-17
Propagate aws kms decrypt's exit code in decrypt-tls-assets

bjornswift:
decrypt-tls-assets runs `aws kms` in a docker container, and then
pipes the output to `base64`. This does not return a non-zero exit
code when decryption fails, and so the unit is not retried.

This commit adds an explicit test, forcing a non-zero exit code,
causing bash -e to barf.

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -234,7 +234,8 @@ write_files:
 
       for encKey in $(find /etc/kubernetes/ssl/*.pem.enc);do
         tmpPath="/tmp/$(basename $encKey).tmp"
-        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext | base64 --decode > $tmpPath
+        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext > $tmpPath.b64
+        base64 --decode < $tmpPath.b64 > $tmpPath
         mv  $tmpPath /etc/kubernetes/ssl/$(basename $encKey .enc)
       done
 
--- a/multi-node/aws/pkg/config/templates/cloud-config-worker
+++ b/multi-node/aws/pkg/config/templates/cloud-config-worker
@@ -166,7 +166,8 @@ write_files:
 
       for encKey in $(find /etc/kubernetes/ssl/*.pem.enc);do
         tmpPath="/tmp/$(basename $encKey).tmp"
-        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext | base64 --decode > $tmpPath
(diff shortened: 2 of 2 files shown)


bjornswift:
Here is an example of a failed run that returned an exit code of 0 (without this patch):

```
$ systemctl status decrypt-tls-assets
● decrypt-tls-assets.service - decrypt kubelet tls assets using amazon kms
   Loaded: loaded (/etc/systemd/system/decrypt-tls-assets.service; enabled; vendor preset: disabled)
   Active: active (exited) since Tue 2016-08-16 22:13:39 UTC; 2h 7min ago
  Process: 1155 ExecStart=/opt/bin/decrypt-tls-assets (code=exited, status=0/SUCCESS)
 Main PID: 1155 (code=exited, status=0/SUCCESS)
    Tasks: 0
   Memory: 0B
      CPU: 0
   CGroup: /system.slice/decrypt-tls-assets.service

Aug 16 22:13:26 machine-ip decrypt-tls-assets[1155]: 0c1d23313396: Pull complete
Aug 16 22:13:26 machine-ip decrypt-tls-assets[1155]: 0c1d23313396: Pull complete
Aug 16 22:13:26 machine-ip decrypt-tls-assets[1155]: a3ed95caeb02: Pull complete
Aug 16 22:13:26 machine-ip decrypt-tls-assets[1155]: a3ed95caeb02: Pull complete
Aug 16 22:13:33 machine-ip decrypt-tls-assets[1155]: 526415505a9c: Pull complete
Aug 16 22:13:33 machine-ip decrypt-tls-assets[1155]: 526415505a9c: Pull complete
Aug 16 22:13:33 machine-ip decrypt-tls-assets[1155]: Digest: sha256:2da3dc06d6894992ec07257040af24f79dd443b239e0b13d94035fb9ec483228
Aug 16 22:13:33 machine-ip decrypt-tls-assets[1155]: Status: Downloaded newer image for quay.io/coreos/awscli:latest
Aug 16 22:13:35 machine-ip decrypt-tls-assets[1155]: A client error (UnrecognizedClientException) occurred when calling the Decrypt operation: The security token included in the request is invalid.
Aug 16 22:13:39 machine-ip systemd[1]: Started decrypt kubelet tls assets using amazon kms.
```

bjornswift:
Adding @colhom as someone who's worked in this area.

colhom:
@bjornswift back from vacation, sorry for the delay. great catch, thanks for the eyeballs!

Could we perhaps avoid using `;` and `PIPESTATUS` to fix this- I feel it's kind of an esoteric solution for a simple problem.

I propose instead:

```sh 
docker run aws decrypt > $tmpPath.b64
base64 --decode <  $tmpPath.b64 > $tmpPath
```

colhom:
@bjornswift lgtm! squash the commits and we'll merge up.

cgag:
Setting pipefail would solve this as well. 

You could change
`#!/bin/bash -e`
to 
```
#!/bin/bash
set -euo pipefail
```

and I believe it would be equivalent to adding that pipestatus check to all pipelines in the script (http://redsymbol.net/articles/unofficial-bash-strict-mode/).  These happen to be the only pipelines in the code but I like the potential future-proofing and the explicit piping rather than introducing tmp paths.  I don't feel that strongly though so if you want to stick to this form.

bjornswift:
@colhom: Thanks, squashed and tested.

@cgag: Yeah, I had originally done the `PIPESTATUS` dance to avoid having to create (and leave behind or clean) more temp files. `pipefail` would also be interesting. I don't have a strong opinion here, but would love for some version of this to make it upstream (we've patched this locally).

colhom:
I've created #646 to future-proof use against this particular class of bugs as well. Thanks @bjornswift!

PR #631, opened 2016-08-22
all: bump to v1.3.5+coreos.1 release

pbx0:


aaronlevy:
There are a couple places the bump-version script does not pick up. One such place is the kubectl docs: https://github.com/pbx0/coreos-kubernetes/blob/235b6a8ef602f3974710522306ce5b2c1167240e/Documentation/configure-kubectl.md#download-the-kubectl-executable

Can you do a quick grep of repo to pick those up? We also could update the bump-version script to have a base_version=v1.3.5 build_version=coreos.1 so we can pick up both version types.

pbx0:
Ack, Ill fix up what is left out. I think i figured out how to turn v1.3.5_coreos.1 to "v1.3.5" and "coreos.1" in bash and then we can just substitute each one separately.

contrib/bump-version.sh
```diff
@@ -4,24 +4,35 @@
...
 TARGET_VERSION=${1}
 
+CURRENT_INDEX=`expr index $CURRENT_VERSION _`
+TARGET_INDEX=`expr index $TARGET_VERSION _`
+
+CURRENT_VERSION_BASE=${CURRENT_VERSION:0:$CURRENT_INDEX -1}
+TARGET_VERSION_BASE=${TARGET_VERSION:0:$TARGET_INDEX -1}
+
+CURRENT_VERSION_BUILD=${CURRENT_VERSION:$CURRENT_INDEX}
+TARGET_VERSION_BUILD=${TARGET_VERSION:$TARGET_INDEX}
```
aaronlevy: I appreciate the effort here, but at this point I think lets just leave it as is and maybe an echo to the end along the lines of:

NOTE: grep for major.minor version, this script may not catch all changes: git grep "v1\.3\."

I say this because there are examples of extraneous version changes in this PR alone. The script wasn't meant to be perfect, just reduce the annoyance.

Documentation/kubernetes-upgrade.md
```diff
@@ -54,9 +54,9 @@ The master node components (kube-controller-manager,kube-scheduler, kube-apiserv
 
 For example, to upgrade the kube-apiserver version you could update the pod image tag in `/etc/kubernetes/manifests/kube-apiserver.yaml`:
 
-From: `image: quay.io/coreos/hyperkube:v1.0.6_coreos.0`
+From: `image: quay.io/coreos/hyperkube:v1.0.6_coreos.1`
 
-To: `image: quay.io/coreos/hyperkube:v1.0.7_coreos.0`
+To: `image: quay.io/coreos/hyperkube:v1.0.7_coreos.1`
```
aaronlevy: This is what I mean in the other comment about extraneous version changes

contrib/bump-version.sh
aaronlevy: If you really want to add something like this, maybe just 2 passes:

1 pass for full version string: `vx.y.z_coreos.v`
1 pass for upstream version: `vx.y.z`

pbx0:
@aaronlevy PTAL. I went with your second suggestion, but added one third pass. I believe the script is correct now (At least as far as these heuristics ever go).

1 pass for full image string: `vx.y.z_coreos.v`
1 pass for full semver string: `vx.y.z+coreos.v`(catches updating the conformance script)
1 pass for upstream version: `vx.y.z`

I know I kinda went down the rabbit hole on this one, but in general I feel that if you can push something that is 90% percent automated to 100% percent it usually a win. That said, I know i'm just chasing a pretty minor problem that doesn't even necessarily have a truly correct way of doing things. So, if this is still broken, I'm fine throwing out the script changes at this point too or generally deleting any substitutions that will ever do things wrong.

aaronlevy:
lgtm

PR #638, opened 2016-08-24
documentation: document using rbd volume plugins

pbx0:
@aaronlevy @robszumski This documentation will be valid once https://github.com/coreos/kubernetes/pull/71 makes it into one of our hyperkube releases. 

Do not merge until ceph-common is included in our latest hyperkube image.


Documentation/kubelet-wrapper.md
```diff
@@ -74,6 +74,17 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   ...other flags...
 ```
 
+### Allow pods to use rbd volumes
+
+Pods using the [rbd volume plugin][rbd-example] to consume data from ceph must ensure that the kubelet has access to modprobe. Add the following options to the `RKT_OPTS` env before launching the kubelet via kubelet-wrapper:
+
+```ini
```
robszumski: Can you show this in the same format of the others?

```
[Service]
Environment="RKT_OPTS=stuff"
Environment=KUBELET_VERSION=v1.3.4_coreos.0
...
```

robszumski:
Opened https://github.com/coreos/coreos-kubernetes/pull/639 to surface this information more clearly

Documentation/kubelet-wrapper.md
pbx0: I waffled on this because I didn't want people to copy and paste this over existing RKT_OPTS. But, I think it makes sense to be consistent with the other examples and hint at where you might find a RKT_OPS env being set in the first place. Updated to reflect this.

robszumski:
LGTM

aaronlevy:
This is waiting on there being a hyperkube image with ceph-common, correct?

pbx0:
@aaronlevy correct

Documentation/kubelet-wrapper.md
```diff
@@ -74,6 +74,20 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
...
+[Service]
+Environment="RKT_OPTS=--volume modprobe,kind=host,source=/usr/sbin/modprobe \
+--mount volume=modprobe,target=/usr/sbin/modprobe \
+--volume lib-modules,kind=host,source=/lib/modules \
+--mount volume=lib-modules,target=/lib/modules \
+Environment=KUBELET_VERSION=v1.3.4_coreos.0
```
aaronlevy: Can we add a note about minimum image version that contains the ceph packages (which will be v1.3.6)

pbx0: Added a sentence to this effect.

aaronlevy: We should probably make this line be at least the minimum version too

pbx0: added commit to bump to v1.3.6+coreos.0

aaronlevy: Looks like this still didn't get bumped

aaronlevy:
Looks like just that one last line to change, but otherwise lgtm

pbx0:
Fixed.

PR #639, opened 2016-08-24
docs: mention optional config for all platforms (rdb, iscsi, etc)

robszumski:
Directly linking to these and listing them out will make more folks get the install they are looking for. This also uses terms that folks are more apt to be looking for (`rbd`, `iscsi`).

Should land near https://github.com/coreos/coreos-kubernetes/pull/638


pbx0:
LGTM -- Lets merge it when #638 is ready to go.

PR #642, opened 2016-08-25
all: make docker cni configuration flexible

pbx0:
Overriding the 'ExecStart' to remove some environment variables will
break in CoreOS 1151.0.0 because Docker v1.12 will balk if given a
'daemon' argument to dockerd. We need to override the environment
variables set from an EnvironmentFile directive which requires another
EnvironmentFile directive. This way things will work with Docker v1.11
and Docker v1.12

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-worker
+++ b/multi-node/aws/pkg/config/templates/cloud-config-worker
@@ -14,8 +14,7 @@ coreos:
             Requires=flanneld.service
             After=flanneld.service
             [Service]
-            ExecStart=
-            ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// $DOCKER_OPTS $DOCKER_CGROUPS $DOCKER_OPT_MTU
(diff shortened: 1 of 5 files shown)


multi-node/aws/pkg/config/templates/cloud-config-worker
```diff
@@ -131,6 +130,11 @@ coreos:
         RequiredBy=kubelet.service
 
 write_files:
+ - path: /etc/kubernetes/cni/docker_opts_cni.env
+    content: |
```
euank: indentation glitch here I think

euank:
Looks like a reasonable way to fix it if it works. Just double checked that it is [explicitly documented](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=) that the order of files should be respected, so LGTM after spacing is happy and travis is happy

pbx0:
Fixed the spacing issue. Should be good.

PR #643, opened 2016-08-26
all: bump rktnetes min version

pbx0:
docker2aci bug in v1.11 rkt is problematic, bump minimum OS version now that alpha is rkt v1.13


dghubble:
:+1:

aaronlevy:
lgtm

Issue #621, opened 2016-08-16
enable support for non-expiring certificates


sym3tri:
So users don't get locked out of their clusters.

Make expiry duration an explicit choice during provisioning, and/or default to not expiring.

iameli:
Related to https://github.com/coreos/coreos-kubernetes/issues/561, which is the ticket to allow for updating TLS assets in CloudFormation on AWS.

philips:
Lets use  3153600000 seconds. This is about 68 years and the max you can do. https://groups.google.com/forum/#!topic/mailing.openssl.users/3kK_f0ywCZQ

philips:
Once you get back can you take this over @colhom ?

brianredbeard:
I just have to express that this is a bad idea.  I know that users having a shitty experience is one thing, but #YOLO-OPS isn't the answer.

mike-saparov:
Can we not use 68 years, users will never pass any security certification with such settings... We are encouraging a bad practice IMHO. Can we instead consider very aggressive notification about certificate expiration?

aaronlevy:
How about we just let people choose as an option?

sym3tri:
> How about we just let people choose as an option?

My original suggestion was to _force_ a choice. 
No defaults == no surprises.
Choice == choose-your-own-security.

pbx0:
It sounds like our original reason for expiring the certs was to force people to come up with a better solution if they use this in production. Its not that unlimited certs are particularly less secure then our 90 day certs because its YOLO either way if people can't update or revoke certs. Its just that 90 days forces people to take control over their TLS infra or be broken.

It seems like we should do something in the interim between now and having auto-tls boostraping and auto-rotation (sounds like the v1.5 timeframe for it all to work. It is just such an awful user experience otherwise to have the cluster hard-fail at 90 days.

Talking with @aaronlevy and @gtank It seems like a reasonable solution for now is to:
- leave the default expiration but add a warning on when they start the cluster
- add an `--insecure-yolo-unlimited-tls` option for those that would do this anyway.

@sym3tri We could also force the choice and not provide a default expiration, but if most users are going to choose 10 years anyway, I'd rather them be forced to pick the yolo option. I don't feel very strongly about this last point though. Any thoughts?

sym3tri:
> sounds like the v1.5 timeframe for it all to work

I thought everything was merged & ready to go in 1.4?

[...]

PR #645, opened 2016-08-30
configure tls cert expiration

pbx0:
Implementation of my last comment in this thread: https://github.com/coreos/coreos-kubernetes/issues/621

Happy to change this as that thread forms more of a consensus. I don't feel too strongly about my stance on this interim state.

cc @aaronlevy

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -40,6 +40,8 @@ func newDefaultCluster() *Cluster {
 		DNSServiceIP:             "10.3.0.10",
 		K8sVer:                   "v1.3.5_coreos.1",
 		HyperkubeImageRepo:       "quay.io/coreos/hyperkube",
+		TLSCADurationDays:        365 * 10,
+		TLSCertDurationDays:      365,
 		ContainerRuntime:         "docker",
 		ControllerInstanceType:   "m3.medium",
 		ControllerRootVolumeType: "gp2",
@@ -136,6 +138,8 @@ type Cluster struct {
 	KMSKeyARN                string            `yaml:"kmsKeyArn,omitempty"`
 	CreateRecordSet          bool              `yaml:"createRecordSet,omitempty"`
 	RecordSetTTL             int               `yaml:"recordSetTTL,omitempty"`
+	TLSCADurationDays        int               `yaml:"tlsCADurationDays,omitempty"`
+	TLSCertDurationDays      int               `yaml:"tlsCertDurationDays,omitempty"`
 	HostedZone               string            `yaml:"hostedZone,omitempty"`
 	HostedZoneID             string            `yaml:"hostedZoneId,omitempty"`
 	StackTags                map[string]string `yaml:"stackTags,omitempty"`
--- a/multi-node/aws/pkg/config/tls_config.go
+++ b/multi-node/aws/pkg/config/tls_config.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net"
 	"path/filepath"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/kms"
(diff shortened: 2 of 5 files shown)


aaronlevy:
I'm starting to think we should just have:

```
--tls-ca-expiration=X
--tls-cert-expiration=y
```

Upstream kube-up uses easy-rsa which looks like it defaults to 10years for both (https://github.com/brendandburns/easy-rsa/blob/e11310d2e8dfdf01910185abc944e78053ae2bfe/easyrsa3/easyrsa#L1006-L1007)

I'm perfectly fine just defaulting to that. Or if we want we can still default to 90days/1year.

Then we just log how long we've generated a cert for in the output. e.g. "NOTE: Generated TLS assets will expire in X days". 

I don't think adding scary-named fixed duration flags is necessarily. Configurable just makes this more flexible to end user, and we can accomplish this in clearly named flags.

@philips @mike-saparov

cgag:
cc @gtank

aaronlevy:
Unless there are any objections, lets move forward with the option to self-select the CA/cert expiration and default to 10years (same as upstream).

If possible, I'd probably prefer that this is consumed from the config file over cli flags. This way we can have comments describing the configurable items - and can reduce flag bloat.

pbx0:
PTAL: We now default to 10 years, warn regardless, and allow the duration to be configured in days

aaronlevy:
We should allow for for differing expirations between CA and certs.

pbx0:
PTAL: There are two yaml variables for the CA expiration vs cert expiration

aaronlevy:
@pbx0 can you add the values to the config.yaml with short comment description. Also, might not hurt to update the config tests to check that we're not allowing invalid values.

/cc @colhom given it's kube-aws

aaronlevy:
Also looks like the config unit tests are failing

multi-node/aws/pkg/config/config.go
```diff
@@ -136,6 +136,8 @@ type Cluster struct {
 	KMSKeyARN                string            `yaml:"kmsKeyArn,omitempty"`
 	CreateRecordSet          bool              `yaml:"createRecordSet,omitempty"`
 	RecordSetTTL             int               `yaml:"recordSetTTL,omitempty"`
+	TLSCADurationDays        int               `yaml:"tlsCAValidDurationDays,omitempty"`
```
colhom: nit: convention is that struct field names are exactly equivalent to yaml tags if you normalize capitalization. nothing explicitly relies on this convention, but i don't see a reason to break it.

multi-node/aws/pkg/config/tls_config.go
```diff
@@ -39,7 +40,7 @@ type CompactTLSAssets struct {
 	AdminKey      string
 }
 
-func (c *Cluster) NewTLSAssets() (*RawTLSAssets, error) {
+func (c *Cluster) NewTLSAssets(ca, cert time.Duration) (*RawTLSAssets, error) {
```
colhom: `caDuration` and `certDuration`.

multi-node/aws/pkg/config/config.go
colhom: @pbx0 you should set default values [here](https://github.com/pbx0/coreos-kubernetes/blob/3ff4b7e2bbc80dfdeaa50e4bfc16ec7e0a2fadb2/multi-node/aws/pkg/config/config.go#L32) for the durations. That will fix the unit tests.

colhom:
@aaronlevy unit tests are probably failing because the default value for the config struct fields aren't set, so I imagine it's defaulting to 0.

multi-node/aws/pkg/config/config.go
pbx0: Oh yeah, missed updating the yaml tag when I was playing with names. 

Right now I validate the values in `command_render.go` but we can add test validation as well. Also `x509.go` additionally will use default values if it receives durations <= 0.

I'll change `x509.go` do the runtime validation but not set defaults. Add a test validation step. And set the defaults the normal way instead.

multi-node/aws/pkg/config/tls_config.go
pbx0: done

pbx0:
PTAL: Unit tests were failing because the signature of NewTLSAssets() had changed. That is no longer the case as I missed some obvious simplifications before. I think I cleaned up things a bit. 

The unit tests in `pkg/config/tls_config_test.go` should already cover this code. Runtime validation occurs in `multi-node/aws/pkg/tlsutil/x509.go`

colhom:
LGTM

Issue #322, opened 2016-03-17
kubelet-wrapper needs to mount /var/log


aknuds1:
The kubelet-wrapper needs to include /var/log among its mount points, since kubelet puts container logs in /var/log/containers. I've confirmed this by modifying kubelet-wrapper myself.

colhom:
@aknuds1 check out #320 

We are going to ship the logging components as cluster add-ons in the near future, at which point we will have the rkt kubelet host-mount it's container log files. Until then, just add an Environment entry for `RKT_OPTS` in kubelet.service, which defines the addtional mount flags for `/var/log/container`

aknuds1:
OK, cool. Keep me posted. You should've mounted /var/log/containers from the get-go though, as it breaks kubelet.

aaronlevy:
@aknuds1 curious what breaks in the kubelet?

aknuds1:
@aaronlevy As I state in the issue description, Kubelet isn't able to place logs in /var/log/containers, meaning that in turn fluentd can't collect container logs. So basically, the Kubernetes logging stack (Fluentd/Elasticsearch/Kibana) doesn't work as a result.

aaronlevy:
Right. Just wanted to make sure there wasn't some other breakage I was missing.

FWIW, the kubelet is still collecting the logs, but just inside the chroot of the kubelet container (not the host /var/log/containers). So the kubelet should function normally, but you're correct that external logging applications would not have access to this data.

As @chom mentioned, you can add this mount using `Environment=RKT_OPTS=` in the kubelet unit file.

The original intention was to keep the kubelet-wrapper script as a simple starting point, but we may need to add more to the defaults. I'll leave this issue open to track this feature request.

MurgaNikolay:
Hi guys.
How we can mount `/var/log/containers` to rkt via RKT_OPTS?
I try to add `Environment=RKT_OPTS=--volume var-log-container,kind=host,source=/var/log/containers --mount volume=log,target=/var/log/containers` to the kubelet unit but i have error in the log:

    [/etc/systemd/system/kubelet.service:5] Invalid environment assignment, ignoring: RKT_OPTS=--volume var-log-container,kind=host,source=/var/log/containers    --mount volume=log,target=/var/log/containers
    ....
[...]

PR #650, opened 2016-09-01
Expose hosts' /var/log to the kubelet by default

danielfm:
The kubelet creates symlinks that capture the pod name, namespace, container name & Docker container ID to the Docker logs for pods in the `/var/log/containers` directory on the host.

By mapping the volume for this directory to the kubelet container by default, it becomes easier to use cluster level logging in Kubernetes, i.e., [fluentd-elasticsearch](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/fluentd-elasticsearch), if the user choose to (see my comment in #320).

This PR updates both single-node and multi-node configurations.

(I should probably update some documentation as well, but first I want to know how you guys feel about this.)

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -43,12 +43,15 @@ coreos:
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
         Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
         --mount volume=dns,target=/etc/resolv.conf \
-        --volume=rkt,kind=host,source=/opt/bin/host-rkt \
+        --volume rkt,kind=host,source=/opt/bin/host-rkt \
         --mount volume=rkt,target=/usr/bin/rkt \
         --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
         --mount volume=var-lib-rkt,target=/var/lib/rkt \
-        --volume=stage,kind=host,source=/tmp \
-        --mount volume=stage,target=/tmp"
+        --volume stage,kind=host,source=/tmp \
+        --mount volume=stage,target=/tmp \
+        --volume var-log,kind=host,source=/var/log \
+        --mount volume=var-log,target=/var/log"
+        ExecStartPre=/usr/bin/mkdir -p /var/log/containers
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers=http://localhost:8080 \
         --network-plugin-dir=/etc/kubernetes/cni/net.d \
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -104,7 +104,6 @@ Note that the kubelet running on a master node may log repeated attempts to post
   - [allow pods to mount RDB][rdb] or [iSCSI volumes][iscsi]
   - [allowing access to insecure container registries][insecure-registry]
   - [use host DNS configuration instead of a public DNS server][host-dns]
-  - [enable the cluster logging add-on][cluster-logging]
   - [changing your CoreOS auto-update settings][update]
 
 **/etc/systemd/system/kubelet.service**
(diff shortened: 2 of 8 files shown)


robszumski:
I'm in favor of this, as it fixes https://github.com/coreos/coreos-kubernetes/issues/322#issuecomment-244144227. What do you think @colhom?

You are correct, the docs will need to be updated in all of the manual example manifests, etc.

AlmogBaku:
LGTM

danielfm:
@robszumski I figured there was no documentation about exposing the `/var/log` directory to the kubelet, but [apparently](https://github.com/coreos/coreos-kubernetes/blob/3d7ce8a88059eac7966c516aa6db7f196379afbd/Documentation/kubelet-wrapper.md#use-the-cluster-logging-add-on) I was wrong.

Should I change anything there?

chancez:
I am strongly in favor of this change.

robszumski:
@danielfm yes, it is documented there, but if we are going to do this by default we need to remove the example from there and add it to each of the example units in all of the guides.

I think thats these, in addition to what you've already changed:

https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-master.md#create-the-kubelet-unit
https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-workers.md#create-the-kubelet-unit

robszumski:
Nice, I think you got all of em.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -43,12 +43,15 @@ coreos:
...
         Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
         --mount volume=dns,target=/etc/resolv.conf \
-        --volume=rkt,kind=host,source=/opt/bin/host-rkt \
+        --volume rkt,kind=host,source=/opt/bin/host-rkt \
         --mount volume=rkt,target=/usr/bin/rkt \
         --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
         --mount volume=var-lib-rkt,target=/var/lib/rkt \
-        --volume=stage,kind=host,source=/tmp \
-        --mount volume=stage,target=/tmp"
+        --volume stage,kind=host,source=/tmp \
```
colhom: What is the motivation for host mounting `/tmp`?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -43,12 +43,15 @@ coreos:
...
+        --mount volume=stage,target=/tmp \
+        --volume var-log,kind=host,source=/var/log \
+        --mount volume=var-log,target=/var/log"
```
colhom: Source and target should both be `/var/log/containers`. I don't think we want to merge the entire `/var/log` directory between host and kubelet container- that could get messy for a variety of reasons. 

Plus, as far as I can see, there's no reason to? Why does the kubelet need to see all the host's logs, or visa versa?

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Our setup actually mounts in an entirely separate directory to store cursor positions for our aggregrator (currently fluentd) across pod lifecycle. I suspect maybe that's what this is for?

chancez: The default setup uses a file in /var/log for the cursor, which is why it's /var/log instead of just /var/log/container, so it's not for that.

multi-node/aws/pkg/config/templates/cloud-config-controller
chancez: See comment above. Cursor files go in /var/log by default. I don't think this is anything we need to worry about. The kubelet runs just fine outside of a container where it doesn't have a separate /var/log to itself, so mounting it in shouldn't be a big deal.

multi-node/aws/pkg/config/templates/cloud-config-controller
danielfm: @colhom 

> What is the motivation for host mounting `/tmp`?
>
> Our setup actually mounts in an entirely separate directory to store cursor positions for our aggregrator (currently fluentd) across pod lifecycle. I suspect maybe that's what this is for?

The `/mnt` mounting already existed there; I only removed the `=` there after `--volume` (to keep it consistent with the other flags).

danielfm:
Is there anything else I need to do to get this merged?

multi-node/aws/pkg/config/templates/cloud-config-controller
pbx0: `/tmp` is mounted in for rktnetes. It will eventually go away but for now it is needed for some communication between host rkt and containerized kubelet.

aaronlevy:
Overall this lgtm. @chancez or @colhom can you confirm these changes work for you? I've not ever setup any k8s log aggregation

danielfm:
I've been running my clusters with this mod for at least three months, but of course, the more people testing this the better!

chancez:
@aaronlevy these emulate exactly what i've had in a patch file for multiple months, if I find sometime I can validate exactly this PR on vagrant etc (no time for much else), but all the changes look fine to me.

I personally prefer the `--flag=value` form with the `=`, but thats not really a big deal and I'd rather get this in.

colhom:
Thanks @danielfm.

PR #652, opened 2016-09-01
Calico: Mount /etc/resolv.conf so DNS works

### LINKED ISSUE #651: Calico doesn't resolve hostnames
Reported by a user on Twitter:

> @coreoslinux k8s setup guide https://coreos.com/kubernetes/docs/latest/deploy-master.html uses rkt for calico node - doesn’t resolve anything other than local host :/

https://twitter.com/alexouzounis/status/771392935204192256

> @coreoslinux after spending hours trying to realise that it’s the resolver, I got fed up and used @docker. Worked straight away

https://twitter.com/alexouzounis/status/771393213722685440

> @coreoslinux @docker issue is the etcd end points works only for local host. If you want to use rkt, one would need to mnt /etc/resolv.conf

https://twitter.com/alexouzounis/status/771393720239423489

tomdee:
Fixes #651

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -350,6 +350,8 @@ Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
 ExecStart=/usr/bin/rkt run --inherit-env --stage1-from-dir=stage1-fly.aci \
 --volume=modules,kind=host,source=/lib/modules,readOnly=false \
 --mount=volume=modules,target=/lib/modules \
+--volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true \
+--mount=volume=dns,target=/etc/resolv.conf \
 --trust-keys-from-https quay.io/calico/node:v0.19.0
 
 KillMode=mixed
--- a/Documentation/deploy-workers.md
+++ b/Documentation/deploy-workers.md
@@ -257,6 +257,8 @@ Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
 ExecStart=/usr/bin/rkt run --inherit-env --stage1-from-dir=stage1-fly.aci \
 --volume=modules,kind=host,source=/lib/modules,readOnly=false \
 --mount=volume=modules,target=/lib/modules \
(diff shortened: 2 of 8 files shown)


robszumski:
I'd like to see this use the `RKT_OPTS` variable similar to https://github.com/coreos/coreos-kubernetes/pull/650 ([example](https://github.com/coreos/coreos-kubernetes/pull/650/files#diff-602bf92bbbb33b5e28e6b8ede6821f98R117)) so that we can make the wrapper look very similar to if you were just using the same flags on the actual kubelet. It makes it easier to understand whats going on, vs mixing in rkt flags with k8s flags.

robszumski:
ping @tomdee does that sound ok?

aaronlevy:
@robszumski I might be misunderstanding, but the RKT_OPTS is for passing options to the kubelet wrapper - these flag changes are for executing the calico container. Are you suggesting we may a calico wrapper script instead?

Documentation/deploy-master.md
```diff
@@ -350,6 +350,8 @@ Environment=ETCD_ENDPOINTS=${ETCD_ENDPOINTS}
 ExecStart=/usr/bin/rkt run --inherit-env --stage1-from-dir=stage1-fly.aci \
 --volume=modules,kind=host,source=/lib/modules,readOnly=false \
 --mount=volume=modules,target=/lib/modules \
+--volume=dns,kind=host,source=/etc/resolv.conf,readOnly=false \
```
aaronlevy: Why do you need write access to host resolv.conf?

robszumski:
> @robszumski I might be misunderstanding, but the RKT_OPTS is for passing options to the kubelet wrapper - these flag changes are for executing the calico container. Are you suggesting we may a calico wrapper script instead?

Oh right. Disregard me.

tomdee:
OK, changed to readonly=true, PTAL.

tomdee:
I should give it another test - probably won't be today though.

tomdee:
OK, this is now tested and is good to merge. @aaronlevy @robszumski

PR #653, opened 2016-09-02
aws: use hyperkube kubectl to install kube-system

### LINKED ISSUE #447: install-kube-system is not robust to some failures

colhom:
convert all manifests to yaml as well

fixes #447 

\cc @cgag @aaronlevy @daveey

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -398,7 +398,7 @@ write_files:
             timeoutSeconds: 15
 
 {{ if .UseCalico }}
-  - path: /srv/kubernetes/addons/calico-policy-controller.yaml
+  - path: /etc/kubernetes/addons/calico-policy-controller.yaml
     content: |
       apiVersion: v1
       kind: Pod
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -43,12 +43,15 @@ coreos:
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
         Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
         --mount volume=dns,target=/etc/resolv.conf \
-        --volume=rkt,kind=host,source=/opt/bin/host-rkt \
+        --volume rkt,kind=host,source=/opt/bin/host-rkt \
         --mount volume=rkt,target=/usr/bin/rkt \
         --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
         --mount volume=var-lib-rkt,target=/var/lib/rkt \
-        --volume=stage,kind=host,source=/tmp \
-        --mount volume=stage,target=/tmp"
+        --volume stage,kind=host,source=/tmp \
+        --mount volume=stage,target=/tmp \
+        --volume var-log,kind=host,source=/var/log \
+        --mount volume=var-log,target=/var/log"
+        ExecStartPre=/usr/bin/mkdir -p /var/log/containers
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers=http://localhost:8080 \
         --network-plugin-dir=/etc/kubernetes/cni/net.d \
@@ -134,8 +137,6 @@ coreos:
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
         Before=kubelet.service
-        After=docker.service
-        Requires=docker.service
 
         [Service]
         Type=oneshot
@@ -211,10 +212,24 @@ write_files:
       #!/bin/bash -e
 
       for encKey in $(find /etc/kubernetes/ssl/*.pem.enc);do
-        tmpPath="/tmp/$(basename $encKey).tmp"
-        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext > $tmpPath.b64
-        base64 --decode < $tmpPath.b64 > $tmpPath
-        mv  $tmpPath /etc/kubernetes/ssl/$(basename $encKey .enc)
+        sudo rkt run \
+        --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
+        --mount=volume=ssl,target=/etc/kubernetes/ssl \
+        --uuid-file-save=/var/run/coreos/decrypt-tls-assets.uuid \
+        --dns=8.8.8.8 --dns=8.8.4.4 \
+        --net=host \
+        --trust-keys-from-https \
+        quay.io/coreos/awscli --exec=/bin/bash -- \
+          -c \
+          "/usr/bin/aws \
+            --region {{.Region}} kms decrypt \
+            --ciphertext-blob fileb://$encKey \
+            --output text \
+            --query Plaintext \
+            > $encKey.b64"
+
+        base64 --decode < $encKey.b64 > ${encKey%.enc}
+        sudo rkt rm --uuid-file=/var/run/coreos/decrypt-tls-assets.uuid
       done
 
   - path: /etc/kubernetes/manifests/kube-proxy.yaml
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -43,15 +43,12 @@ coreos:
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
         Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
         --mount volume=dns,target=/etc/resolv.conf \
-        --volume rkt,kind=host,source=/opt/bin/host-rkt \
+        --volume=rkt,kind=host,source=/opt/bin/host-rkt \
         --mount volume=rkt,target=/usr/bin/rkt \
         --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
         --mount volume=var-lib-rkt,target=/var/lib/rkt \
-        --volume stage,kind=host,source=/tmp \
-        --mount volume=stage,target=/tmp \
-        --volume var-log,kind=host,source=/var/log \
-        --mount volume=var-log,target=/var/log"
-        ExecStartPre=/usr/bin/mkdir -p /var/log/containers
+        --volume=stage,kind=host,source=/tmp \
+        --mount volume=stage,target=/tmp"
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers=http://localhost:8080 \
         --network-plugin-dir=/etc/kubernetes/cni/net.d \
@@ -137,6 +134,8 @@ coreos:
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
         Before=kubelet.service
+        After=docker.service
+        Requires=docker.service
 
         [Service]
         Type=oneshot
@@ -160,45 +159,30 @@ coreos:
         ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
         ExecStart=/opt/bin/install-kube-system
 
-{{ if .UseCalico }}
-    - name: install-calico-system.service
-      command: start
-      content: |
-        [Unit]
-        Requires=kubelet.service docker.service
-        After=kubelet.service docker.service
-
-        [Service]
-        Type=simple
-        StartLimitInterval=0
-        Restart=on-failure
-        ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
-        ExecStart=/opt/bin/install-calico-system
-{{ end }}
-
 write_files:
   - path: /opt/bin/install-kube-system
     permissions: 0700
     owner: root:root
     content: |
-      #!/bin/bash -e
-      /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
-      -d @"/srv/kubernetes/manifests/kube-dns-rc.json" \
-      "http://127.0.0.1:8080/api/v1/namespaces/kube-system/replicationcontrollers"
+      #!/bin/bash
+      set -euo pipefail
 
-      /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
-      -d @"/srv/kubernetes/manifests/kube-dashboard-rc.json" \
-      "http://127.0.0.1:8080/api/v1/namespaces/kube-system/replicationcontrollers"
+      uuid_file_save="/var/run/coreos/rkt-kubectl-${RANDOM}.uuid"
+      function cleanup {
+        sudo rkt rm --uuid-file="$uuid_file_save"
+        sudo rm "$uuid_file_save"
+      }
 
-      /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
-      -d @"/srv/kubernetes/manifests/heapster-de.json" \
-      "http://127.0.0.1:8080/apis/extensions/v1beta1/namespaces/kube-system/deployments"
+      trap cleanup EXIT
 
-      for manifest in {kube-dns,heapster,kube-dashboard}-svc.json;do
-          /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
-          -d @"/srv/kubernetes/manifests/$manifest" \
-          "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
-      done
+      sudo rkt run \
+        --net=host \
+        --uuid-file-save="$uuid_file_save" \
+        --trust-keys-from-https \
+        --volume manifests,kind=host,source=/srv/kubernetes \
+        --mount volume=manifests,target=/srv/kubernetes \
+        {{.HyperkubeImageRepo}}:{{.K8sVer}} \
+        --exec /hyperkube -- kubectl create -f /srv/kubernetes/addons
 
   - path: /etc/kubernetes/cni/docker_opts_cni.env
     content: |
@@ -220,17 +204,6 @@ write_files:
       # https://github.com/coreos/rkt/issues/2878
       exec nsenter -m -u -i -n -p -t 1 -- /usr/bin/rkt "$@"
 
-{{ if .UseCalico }}
-  - path: /opt/bin/install-calico-system
-    permissions: 0700
-    owner: root:root
-    content: |
-      #!/bin/bash -e
-      /usr/bin/curl -H "Content-Type: application/json" -XPOST -d @"/srv/kubernetes/manifests/calico-system.json" "http://127.0.0.1:8080/api/v1/namespaces"
-
-      /usr/bin/cp /srv/kubernetes/manifests/calico-policy-controller.yaml /etc/kubernetes/manifests
-{{ end }}
-
   - path: /opt/bin/decrypt-tls-assets
     owner: root:root
     permissions: 0700
@@ -238,24 +211,10 @@ write_files:
       #!/bin/bash -e
 
       for encKey in $(find /etc/kubernetes/ssl/*.pem.enc);do
-        sudo rkt run \
-        --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
-        --mount=volume=ssl,target=/etc/kubernetes/ssl \
-        --uuid-file-save=/var/run/coreos/decrypt-tls-assets.uuid \
-        --dns=8.8.8.8 --dns=8.8.4.4 \
-        --net=host \
-        --trust-keys-from-https \
-        quay.io/coreos/awscli --exec=/bin/bash -- \
-          -c \
-          "/usr/bin/aws \
-            --region {{.Region}} kms decrypt \
-            --ciphertext-blob fileb://$encKey \
-            --output text \
-            --query Plaintext \
-            > $encKey.b64"
-
-        base64 --decode < $encKey.b64 > ${encKey%.enc}
-        sudo rkt rm --uuid-file=/var/run/coreos/decrypt-tls-assets.uuid
+        tmpPath="/tmp/$(basename $encKey).tmp"
+        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext > $tmpPath.b64
+        base64 --decode < $tmpPath.b64 > $tmpPath
+        mv  $tmpPath /etc/kubernetes/ssl/$(basename $encKey .enc)
       done
 
   - path: /etc/kubernetes/manifests/kube-proxy.yaml
@@ -424,7 +383,7 @@ write_files:
             timeoutSeconds: 15
 
 {{ if .UseCalico }}
-  - path: /srv/kubernetes/manifests/calico-policy-controller.yaml
+  - path: /srv/kubernetes/addons/calico-policy-controller.yaml
     content: |
       apiVersion: v1
       kind: Pod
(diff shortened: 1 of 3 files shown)


multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -159,45 +159,28 @@ coreos:
         ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
         ExecStart=/opt/bin/install-kube-system
 
-{{ if .UseCalico }}
```
colhom: We don't need a separate service for this. If the calico manifests are present in `/srv/kubernetes/manifests`, they get installed as part of `install-kube-system.service` now

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -159,45 +159,24 @@ coreos:
...
-        [Service]
-        Type=simple
-        StartLimitInterval=0
-        Restart=on-failure
-        ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
-        ExecStart=/opt/bin/install-calico-system
-{{ end }}
-
 write_files:
+  - path: /opt/bin/kubectl-common
```
rochacon: Should we keep this separate from the `/opt/bin/install-kube-system`?

If yes, why not rename this to `/opt/bin/kubectl` (and give execution permissions)? That way we may use it as an "kubectl wrapper" everywhere. E.g `/opt/bin/kubectl get nodes`, `/opt/bin/kubectl apply -f ...`

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -437,399 +405,270 @@ write_files:
...
+              name: dnsmasq
+              ports:
+              - containerPort: 53
+                name: dns
+                protocol: UDP
+              - containerPort: 53
+                name: dns-tcp
+                protocol: TCP
+            - args:
+              - -cmd=nslookup kubernetes.default.svc.cluster.local 127.0.0.1 >/dev/null
```
pbx0: Just double checking, this will be treated as one line right?

pbx0:
LGTM

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: The json -> yaml conversion was done via a tool, so I'm relying on it to make that decision correctly. This won't merge till it passes e2e tests, so that's the other end of this equation.

cgag:
Sweet.  LGTM.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -159,45 +159,25 @@ coreos:
...
 write_files:
+  - path: /opt/bin/kubectl
+    permissions: 0755
+    content: |
+      # source this file to set up a kubectl alias
+      sudo rkt run \
+        --net=host \
+        --volume bk,kind=host,source=/srv/kubernetes \
+        --mount volume=bk,target=/srv/kubernetes \
+        --insecure-options=image docker://{{.HyperkubeImageRepo}}:{{.K8sVer}} \
```
aaronlevy: If we're using quay, we don't need insecure-options or docker://

I guess choice is allow all arbitrary docker images, and and remove trust - or standardize on quay images instead.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -159,45 +159,25 @@ coreos:
...
-      -d @"/srv/kubernetes/manifests/heapster-de.json" \
-      "http://127.0.0.1:8080/apis/extensions/v1beta1/namespaces/kube-system/deployments"
-
-      for manifest in {kube-dns,heapster,kube-dashboard}-svc.json;do
-          /usr/bin/curl  -H "Content-Type: application/json" -XPOST \
-          -d @"/srv/kubernetes/manifests/$manifest" \
-          "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
-      done
+
+      /opt/bin/kubectl apply -f /srv/kubernetes/manifests
```
aaronlevy: I think we should be using `create` not `apply` otherwise an admin can never update their addons (this will be run on every boot). `create`, I believe, will only create if nothing exists - if addon is already present it is a no-op.

aaronlevy: Also, can we move these to `/srv/kubernetes/addons` that is somewhat of an upstream convention - and other places (like bootkube checkpointer) manifests get dropped in /srv/kubernetes/manifests as an inactive storage spot, which we do not want to try and create.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -159,45 +159,25 @@ coreos:
...
-        [Service]
-        Type=simple
-        StartLimitInterval=0
-        Restart=on-failure
-        ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
-        ExecStart=/opt/bin/install-calico-system
-{{ end }}
-
 write_files:
+  - path: /opt/bin/kubectl
```
aaronlevy: Every time we run this it is going to create a new container - I'm not sure we want to expose this in a generic way on the hosts (Ideally we just want to use it once on first boot). In normal operations an admin shouldn't need to execute kubectl on the host itself.

If we do want to provide this as a generic tool, we probably need to add more into this wrapper like self-cleanup after execution. See flannel-wrapper / flannel.service wrappers @steveeJ has been working on: https://github.com/coreos/coreos-overlay/pull/2094

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -159,45 +159,25 @@ coreos:
...
+    permissions: 0755
+    content: |
+      # source this file to set up a kubectl alias
+      sudo rkt run \
+        --net=host \
+        --volume bk,kind=host,source=/srv/kubernetes \
+        --mount volume=bk,target=/srv/kubernetes \
+        --insecure-options=image docker://{{.HyperkubeImageRepo}}:{{.K8sVer}} \
+        --exec /hyperkube -- kubectl $@
```
aaronlevy: Should likely quote `"$@"`

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: I'm also a bit on the fence about just doing an entire directory vs being explicit about certain addons.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: i vote standardize on quay (or generically any repo that supports acis), anyone who want to use other stuff can remove trust on their own.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: I don't think so, seeing as we want $@ to expand as an argument array to append to 

`./hyperkube kubectl arg1 arg2 .. argn`

rather than concatenating:

`/hyperkube kubectl "arg1 arg2 .. argn` if we quoted `$@`

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Executing `kubectl` seamlessly on the host is very,very handy when you need it- it'd be counter-productive to hide it. I'll add a trap for now that cleans up the container.

multi-node/aws/pkg/config/templates/cloud-config-controller
cgag: I don't think that's how the quoting works.  It does stuff like prevent glob expansion: 

test.sh: 
```
#!/bin/bash
echo kubectl "$@"
echo "---"
echo kubectl $@
```
```
~ » ./test.sh a b c "*.go"
kubectl a b c *.go
---
kubectl a b c scratch.go
```
Also, should this have a shell pragma?

colhom: @cgag 
> it does stuff like prevent glob expansion:

Yes, this is one example. Argument quoting controls the level at which **any subexpresion**, not just glob expansions, are expanded. Though that is entirely orthogonal to the argument parsing issue.

> I don't think that's how the quoting works.

Unfortunately that **is** how quoting works w.r.t argument parsing in bash, and your echo example does not address this behavior. Try this:

```sh
[chom@hompad-work ~]$ cat test.sh
#!/bin/bash

echo "arg count is: $#"

for arg in "$@";do
  echo "arg: \"${arg}\""
done

[chom@hompad-work ~]$ ./test.sh "larry moe curly"
arg count is: 1
arg: "larry moe curly"
[chom@hompad-work ~]$ ./test.sh larry moe curly
arg count is: 3
arg: "larry"
arg: "moe"
arg: "curly"
```

See what I mean?

colhom: @cgag @aaronlevy just tested- you guys are both right about how it behaves in this case. My bad on that.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Agreed, `apply` would be very disruptive. 
> I'm also a bit on the fence about just doing an entire directory vs being explicit about certain addons.

I really tend to like minimizing layers of indirection between configuration filesystem state and application state. Treating a configuration file's presence in a directory as positive indication it should be loaded is a very nice paradigm used by many common *nix applications (nginx, sshd, udev, systemd, the kubelet itself w.r.t static manifests). 

I feel introducing another state variable buried in a script somewhere to control whether the file is "live" or not obfuscates the state of the system and is not necessary given that the "load any `* .conf` files in this directory" approach is widely used already. (Of course, for this it's `.yaml`)

colhom:
@aaronlevy PTAL

multi-node/aws/pkg/config/templates/cloud-config-controller
aaronlevy: I agree that it could be useful, but I don't think things like arbitrarily choosing to mount `/srv/kubernetes` is generally useful. I think providing a generically useful kubectl wrapper is something that can be tackled in a separate PR - but I'd rather not have something called `kubectl` in PATH that is very specific in functionality.

As it stands, this is being used exactly one time by this script. I'd probably rather we just execute this in-line or drop the file in /tmp on each boot.

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -174,10 +176,9 @@ write_files:
 
       sudo rkt run \
         --net=host \
-        --volume bk,kind=host,source=/srv/kubernetes \
-        --mount volume=bk,target=/srv/kubernetes \
         --uuid-file-save="$uuid_file_save" \
         --trust-keys-from-https \
+        ${RKT_OPTS:-} \
```
colhom: figured it works the same was `kubectl-wrapper` in this regard. Both are wrapping `rkt run hyperkube`

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -163,7 +163,9 @@ write_files:
   - path: /opt/bin/kubectl
```
aaronlevy: Can we please move this out of PATH (`/tmp`?) or just use this functionality directly when creating the addons.

What this will lead to:

```
$ ls manifest.yaml
manifest.yaml
$ kubectl create -f manifest.yaml
error: open manifest.yaml: no such file or directory
```

I agree this functionality will be nice to have -- but it needs more thought and is not strictly required for this PR. Let's have this be a follow up.

colhom: `/tmp` can get garbage collected if disk space is low. I've got a commit up which makes it non-executable so it won't show up on the PATH.

colhom:
@aaronlevy the `rkt run kubectl` logic is now baked into the `install-kube-system` script- no separate file for kubectl logic. I've got #669 up ( \cc @rochacon ) to track effort towards providing a kubectl executable on the PATH.

aaronlevy:
Can you resolve conflicts and make same changes to the single-node and multi-node/generic scripts?

colhom:
Yep will do.

colhom:
@aaronlevy PTAL 

* prior stuff squashed/rebased
* the single/multi node vagrant stuff is in f31f624, both have been tested.

conformance for AWS running at https://jenkins.coreos.systems/view/Kubernetes/job/kubernetes-conformance-aws/115/console

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -43,15 +43,12 @@ coreos:
...
         --mount volume=dns,target=/etc/resolv.conf \
-        --volume rkt,kind=host,source=/opt/bin/host-rkt \
+        --volume=rkt,kind=host,source=/opt/bin/host-rkt \
         --mount volume=rkt,target=/usr/bin/rkt \
         --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
         --mount volume=var-lib-rkt,target=/var/lib/rkt \
-        --volume stage,kind=host,source=/tmp \
-        --mount volume=stage,target=/tmp \
-        --volume var-log,kind=host,source=/var/log \
-        --mount volume=var-log,target=/var/log"
```
aaronlevy: Why dropping the /var/log mounts?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -137,6 +134,8 @@ coreos:
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
         Before=kubelet.service
+        After=docker.service
+        Requires=docker.service
```
aaronlevy: Why is this a requirement of this PR?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -220,42 +204,17 @@ write_files:
...
 
-{{ if .UseCalico }}
-  - path: /opt/bin/install-calico-system
-    permissions: 0700
-    owner: root:root
-    content: |
-      #!/bin/bash -e
-      /usr/bin/curl -H "Content-Type: application/json" -XPOST -d @"/srv/kubernetes/manifests/calico-system.json" "http://127.0.0.1:8080/api/v1/namespaces"
-
-      /usr/bin/cp /srv/kubernetes/manifests/calico-policy-controller.yaml /etc/kubernetes/manifests
```
aaronlevy: policy-controller is a static pod - I don't think this can be removed (it's not created like an addon)

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -424,7 +383,7 @@ write_files:
             timeoutSeconds: 15
 
 {{ if .UseCalico }}
-  - path: /srv/kubernetes/manifests/calico-policy-controller.yaml
+  - path: /srv/kubernetes/addons/calico-policy-controller.yaml
```
aaronlevy: policy-controller is a static pod - it's not created as an addon

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -220,42 +204,17 @@ write_files:
...
-            --output text \
-            --query Plaintext \
-            > $encKey.b64"
-
-        base64 --decode < $encKey.b64 > ${encKey%.enc}
-        sudo rkt rm --uuid-file=/var/run/coreos/decrypt-tls-assets.uuid
+        tmpPath="/tmp/$(basename $encKey).tmp"
+        docker run --rm -v /etc/kubernetes/ssl:/etc/kubernetes/ssl --rm quay.io/coreos/awscli aws --region {{.Region}} kms decrypt --ciphertext-blob fileb://$encKey --output text --query Plaintext > $tmpPath.b64
+        base64 --decode < $tmpPath.b64 > $tmpPath
+        mv  $tmpPath /etc/kubernetes/ssl/$(basename $encKey .enc)
```
aaronlevy: Why is this change necessary for this PR?

multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -160,45 +159,30 @@ coreos:
...
+      trap cleanup EXIT
+
+      sudo rkt run \
+        --net=host \
+        --uuid-file-save="$uuid_file_save" \
+        --trust-keys-from-https \
+        --volume manifests,kind=host,source=/srv/kubernetes \
+        --mount volume=manifests,target=/srv/kubernetes \
+        {{.HyperkubeImageRepo}}:{{.K8sVer}} \
+        --exec /hyperkube -- kubectl create -f /srv/kubernetes/addons
```
aaronlevy: I'm wondering how exit codes will be handled here. For example, does `kubectl create` exit non-zero if it's trying to create assets that already exists (e.g. pretty much every reboot). And if it does exit non-zero, will `rkt run` exit non-zero causing this bash script to fail?

multi-node/generic/controller-install.sh
```diff
@@ -946,16 +818,21 @@ function start_addons {
...
-    echo "K8S: Heapster addon"
-    curl --silent -H "Content-Type: application/json" -XPOST -d"$(cat /srv/kubernetes/manifests/heapster-de.json)" "http://127.0.0.1:8080/apis/extensions/v1beta1/namespaces/kube-system/deployments" > /dev/null
-    curl --silent -H "Content-Type: application/json" -XPOST -d"$(cat /srv/kubernetes/manifests/heapster-svc.json)" "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services" > /dev/null
-    echo "K8S: Dashboard addon"
-    curl --silent -H "Content-Type: application/json" -XPOST -d"$(cat /srv/kubernetes/manifests/kube-dashboard-rc.json)" "http://127.0.0.1:8080/api/v1/namespaces/kube-system/replicationcontrollers" > /dev/null
-    curl --silent -H "Content-Type: application/json" -XPOST -d"$(cat /srv/kubernetes/manifests/kube-dashboard-svc.json)" "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services" > /dev/null
+
+    for f in /srv/kubernetes/addons/{kube-{dns-{rc,svc},dashboard-{rc,svc}},heapster-{de,svc}}.yaml;do
+      uuid_file_save="/var/run/coreos/rkt-kubectl-${RANDOM}.uuid"
```
aaronlevy: Would be cleaner to move this into a function then just call function for each path. And just an "easier to maintain" nit would be using explicit paths:

```
create-addon /srv/kubernetes/addons/kube-dns-rc.yaml
create-addon /srv/kubernetes/addons/kube-dns-svc.yaml
...
```

single-node/user-data
```diff
@@ -83,7 +83,7 @@ function init_templates {
         mkdir -p $(dirname $TEMPLATE)
         cat << EOF > $TEMPLATE
 [Service]
-ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
```
aaronlevy: We still run static manifests which need this directory

single-node/user-data
```diff
@@ -208,7 +208,7 @@ WantedBy=multi-user.target
 EOF
     fi
 
-    local TEMPLATE=/etc/kubernetes/manifests/kube-proxy.yaml
+    local TEMPLATE=/etc/kubernetes/addons/kube-proxy.yaml
```
aaronlevy: I'm assuming this wasn't tested?

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: git error, sorry.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: git error, sorry.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Yeah, that is definitely now a concern that slipped my mind. `kubectl` will return exit status 1, which will cause the service to retry until systemd labels it failed. Which will not explicitly prevent anything else from happening, though it is very undesirable.

I'll figure out something to make this work.

single-node/user-data
colhom: It was tested:
```sh
[chom@hompad-work single-node]$ kubectl --kubeconfig=kubeconfig --namespace=kube-system get pods
NAME                                  READY     STATUS    RESTARTS   AGE
heapster-v1.1.0-3647315203-rpynr      2/2       Running   0          2m
kube-apiserver-172.17.4.99            1/1       Running   0          4m
kube-controller-manager-172.17.4.99   1/1       Running   0          4m
kube-dns-v17.1-izlmn                  3/3       Running   0          3m
kube-proxy-172.17.4.99                1/1       Running   0          4m
kube-scheduler-172.17.4.99            1/1       Running   0          3m
kubernetes-dashboard-v1.1.1-072gk     1/1       Running   0          3m
```

Why are you assuming it wasn't tested?

single-node/user-data
colhom: I moved the static manifests to `/etc/kubernetes/addons`, and the kube-system stuff is in `/srv/kubernetes/addons`. I'll defer to you on correctness of that decision.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: git error, sorry.

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: will change.

multi-node/generic/controller-install.sh
colhom: will do

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom: Yeah, I will move the policy-controller to `/etc/kubernetes/addons`, which is where kubelet get's its static pods from. If you don't like that change, I can revert it to `/etc/kubernetes/manifests` with a simple sed after this review.

single-node/user-data
aaronlevy: I missed that the kubelet `--config` directory was also changed - and didn't see how this would work. I'll comment inline with that change.

single-node/user-data
aaronlevy: We should not move the static pod location (and they're not what you would consider addons).

`/etc/kubernetes/manifests` = static pods
`/srv/kubrnetes/addons` = addons
`/srv/kubernetes/manifests` = this has historically been used as a "at rest" location for inactive manifests.

aaronlevy: For that matter /etc/kubernets/addons would be fine too. No personal preference there.

colhom: So I should leave the static stuff in `/etc/kubernetes/addons` then? It seems like the word `manifests` is falling out of favor.

single-node/user-data
colhom: Oh ok. Should have mentioned that in the commit message.

single-node/user-data
aaronlevy: It is a long standing convention - there is no need to change it now.

Can we just go with this:

`/etc/kubernetes/manifests` = static pods
`/etc/kubernetes/addons` = addons which will be created with `kubectl create -f`
`/srv/kubernetes/manifests` is reserved, but fine if unused.
```

**Comment by pbx0** [comment]
When I was updating the addons for v1.4 I took another attempt at curling the addons in yaml format and things worked. Maybe we rebase this work on that and just use curl?

**Comment by aaronlevy** [comment]
I think that would be a simpler / preferable change. Is there another reason we would want to use kubectl here instead of curl here @colhom?

**Comment by colhom** [comment]
@aaronlevy @pbx0 for all the same reasons that end users generally prefer `kubectl` over `curl` for deploying/managing their applications... 

I'll just state what they are for me I guess.

* changes to HTTP API, content-type, encoding/char-set nuiances, new required headers, etc are all taken care of
* TLS stuff (I have API server TLS rolling our in our next internal update)
* client-side validation
* much greater automation potential- having kubectl available (vs curl) is really convenient for building out automations that update `kube-system` for you... though I'm currently leaning towards using something like [kpm](https://github.com/coreos/kpm) to manage kube-system across cluster upgrades.

**Comment by aaronlevy** [comment]
> for all the same reasons that end users generally prefer kubectl over curl for deploying/managing their applications...

This isn't a user interacting with the api, it's the cloud-config. I'd like to separate these concerns because we shouldn't discuss this functionality from a user perspective. A user also shouldn't generally need to use kubectl from on-host - and providing functionality to easily do this via hyperkube container should be discussed in a separate PR which implements a kubectl-wrapper.

> changes to HTTP API, content-type, encoding/char-set nuiances, new required headers, etc are all taken care of

That I can remember, there has been exactly once this has changed in well over a year of using `curl`, and that was adding the content-type header. Using `curl` has historically been very stable.
 
> TLS stuff (I have API server TLS rolling our in our next internal update)

I would need more specifics here. We are hitting the api-server on the insecure port. What TLS stuff is used here? How does this help with TLS rolling?

> client-side validation

As far as I'm aware, there is no client side validation in `kubectl`. Can you provide an example?

> much greater automation potential- having kubectl available (vs curl) is really convenient for building out automations that update kube-system for you... though I'm currently leaning towards using something like kpm to manage kube-system across cluster upgrades.

Again, this PR is about creating assets, generally only on first boot of the cluster. I agree that running kubectl via a container (possibly via a kubectl-wrapper script) can be useful, but should be discussed / implemented in a PR for that purpose.

---

The functionality we are targeting in this PR is quite focused: create initial cluster assets, likely only a single time on the first master boot. Using `curl` is simple, fast, and stable -- we should have strong reasons to switch for this specific functionality.

### Simplicity:

This PR as it stands currently adds a lot of complexity compared to a simple curl POST.

The goal for switching was primarily because we could not directly create YAML assets, which led to pain points when updating these manifests from their upstream copies. Namely we manually convert the templates each time, have to do it in a specific way, and was often error-prone.

However, it seems that now we can set content-type header to yaml, and this is not longer an issue. So we should get what we want with minimal changes - and maintaining simplicity.

### Speed

Running kubectl from a container is going to be much slower. Some simple tests:

```
core@c1 ~ $ time sudo rkt run --uuid-file-save=/var/run/coreos/kubectl-foo.uuid --trust-keys-from-https --volume manifests,kind=host,source=/srv/kubernetes --mount volume=manifests,target=/srv/kubernetes quay.io/coreos/hyperkube:v1.3.7_coreos.0 --net=host --exec /hyperkube -- kubectl create -f /srv/kubernetes/manifests/pod.yaml
image: using image from local store for image name coreos.com/rkt/stage1-coreos:1.7.0
image: using image from local store for image name quay.io/coreos/hyperkube:v1.3.7_coreos.0
[ 4472.065643] hyperkube[5]: pod "busybox" created

real    0m3.891s
user    0m2.766s
sys     0m0.380s
```

Compared to:

```
core@c1 ~ $ time curl -H "Content-Type: application/yaml" -XPOST -d"$(cat /srv/kubernetes/manifests/pod.yaml)" "http://localhost:8080/api/v1/namespaces/default/pods"
real    0m0.015s
user    0m0.002s
sys     0m0.003s
```

So we are roughly adding ~4seconds per component to every boot (also, when asset is already created, it takes over 4 seconds to return "already exists" -- so subsequent boots are even slower).

A quick check of manifests looks like this, as it is proposed, would add an additional 25-30 seconds to the boot time.

### Stability

We have had almost no issues (that I can remember) using `curl` for over a year. The exceptions were adding the content-header during an update.

### What we are trying to solve:

The only stated issue in https://github.com/coreos/coreos-kubernetes/issues/447 was that we were not differentiating "already exits" from "actual creation errors". Per my comment https://github.com/coreos/coreos-kubernetes/issues/447#issuecomment-217272506, it was not clear to me how using `kubectl` would actually help this issue. Using `curl` we would likely need to parse HTTP status codes - using `kubectl` we would likely need to parse message output. That comment is old - so maybe something has changed -- but I'm not seeing that this change resolves the stated issue. Maybe we just let the curl output be logged instead of silencing it and redirecting output -- that seems to be just as helpful as using kubectl.

Beyond that, we wanted to use the yaml manifests, because it made updating the addons much easier and less error prone for us - at the possible expense of speed / complexity. But now that we can just use `curl` -- I'm not seeing a strong reason to change.

**Comment by colhom** [comment]
@aaronlevy disagree strongly with every point.

> Simplicity: This PR as it stands currently adds a lot of complexity compared to a simple curl POST.

This is not true. Currently we're hiding behind the convenient fact that curl command returns 0 whether or not the API returns 200 for a particular call, with no differentiation. If we wanted the **equivalent behavior to what we have today with curl** out of this setup, install-kube-system would in fact be **much simpler**:

```sh
#!/bin/bash
kubectl="rkt run hyperkube...."
if ! $kubectl get -f /srv/kubernetes/addons;then
     $kubectl create -f /srv/kubernetes/addons
fi 
```
versus

```sh
      #!/bin/bash -e
      /usr/bin/curl  -H "Content-Type: application/yaml" -XPOST \
     --cafile=/etc/kubernetes/ssl/ca.pem \
     --certfile=/etc/kubernetes/ssl/apiserver.pem \
     --keyfile=/etc/kubernetes/ssl/apiserver-key.pem \
      -d @"/srv/kubernetes/manifests/kube-dns-rc.yaml" \
      "http://127.0.0.1:8080/api/v1/namespaces/kube-system/replicationcontrollers"

      /usr/bin/curl  -H "Content-Type: application/yaml" -XPOST \
     --cafile=/etc/kubernetes/ssl/ca.pem \
     --certfile=/etc/kubernetes/ssl/apiserver.pem \
     --keyfile=/etc/kubernetes/ssl/apiserver-key.pem \
      -d @"/srv/kubernetes/manifests/kube-dashboard-rc.yaml" \
      "http://127.0.0.1:8080/api/v1/namespaces/kube-system/replicationcontrollers"

      /usr/bin/curl  -H "Content-Type: application/yaml" -XPOST \
     --cafile=/etc/kubernetes/ssl/ca.pem \
     --certfile=/etc/kubernetes/ssl/apiserver.pem \
     --keyfile=/etc/kubernetes/ssl/apiserver-key.pem \
      -d @"/srv/kubernetes/manifests/heapster-de.yaml" \
      "http://127.0.0.1:8080/apis/extensions/v1beta1/namespaces/kube-system/deployments"

      for manifest in {kube-dns,heapster,kube-dashboard}-svc.yaml;do
       --cafile=/etc/kubernetes/ssl/ca.pem \
       --certfile=/etc/kubernetes/ssl/apiserver.pem \
       --keyfile=/etc/kubernetes/ssl/apiserver-key.pem \
          /usr/bin/curl  -H "Content-Type: application/yaml" -XPOST \
          -d @"/srv/kubernetes/manifests/$manifest" \
          "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
      done
```
Things you have **uncessarily** tell curl manually here:
* which rest endpoint **each different** manifest you have needs to be posted to
  * which namespace each manifest is in
  * which API version each different manifest you have is
  * what resource type each different manifest is
* where your various TLS assets live (I know we don't have any yet in `master`, make this point even more valid)
* what HTTP method you're using
* content-type headers

This is all entirely redundant information that is already in your manifests and/or known to kubectl- you are just wasting your time duplicating the information by using curl. With `kubectl`, none of this is a concern. Throw the file in the folder.. you're done. I know everyone likes to type a lot and feel like we're getting stuff done...

> Speed: Running kubectl from a container is going to be much slower...as it is proposed, would add an additional 25-30 seconds to the boot time.


 It will make initially **provisioning the cluster** take an additional 25-30 seconds before kube-system is available. Seeing as this whole process already takes 10-15 minutes, I do not see this a big deal.

On subsequent boots kube-system is already functional as soon as the kubelet starts all the containers, and does not block on `install-kube-system` completing. 

If we're really concerned about actual boot time (as defined by the kubelet joining the cluster), maybe we should just curl the kubelet binary and run that directly instead of waiting for kubelet-wrapper to pull the image, do the aci conversion and start the container?

> Stability: We have had almost no issues (that I can remember) using curl for over a year. 

But earlier you said:

> The goal for switching was primarily because we could not directly create YAML assets, which led to pain points when updating these manifests from their upstream copies. Namely we manually convert the templates each time, have to do it in a specific way, and was often error-prone.

The reason that we did this for over a year was because it took over a year **because** we didn't have kubectl available for `install-kube-system`. I'd say this is indicative that using kubectl is the right idea, given we were manually converting yaml --> json for many months to work around this.

> As far as I'm aware, there is no client side validation in kubectl. Can you provide an example?

`--validate[=true]: If true, use a schema to validate the input before sending it`

[source](http://kubernetes.io/docs/user-guide/kubectl/kubectl_create/) It's like a delinter for your k8s manifests:

aaronlevy:
I've stated my preference. I'll leave the decision up to you.

PR #666, opened 2016-09-14
kube-aws: use rkt for decrypt-tls-assets

### LINKED ISSUE #545: Suggestion: make kubelet.service NOT depend on docker.service

colhom:
fixes #545 

My investigation has found (at least on my reproductions) that `kubelet.service` failing for dependency reasons is caused by `decrypt-tls-assets.service` explicitly listing docker a dependency and failing for that reason.

I took this opportunity to switch us over to `rkt` for decrypting the TLS assets as well.

\cc @aaronlevy @robszumski @pbx0 @cgag

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -137,8 +137,6 @@ coreos:
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
         Before=kubelet.service
-        After=docker.service
-        Requires=docker.service
 
         [Service]
         Type=oneshot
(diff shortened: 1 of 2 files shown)


multi-node/aws/pkg/config/templates/cloud-config-controller
```diff
@@ -137,8 +137,6 @@ coreos:
         [Unit]
         Description=decrypt kubelet tls assets using amazon kms
         Before=kubelet.service
```
robszumski: why leave this? isn't it a noop unless you also have requires?

pbx0:
\m/ lgtm besides addressing robs comment

multi-node/aws/pkg/config/templates/cloud-config-controller
colhom:
```sh
[Install]
RequiredBy=kubelet.service
```
Dependency injection.

colhom: @robszumski also keep in mind that ordering constraints in systemd are entirely decoupled from dependency relationships. 

Even in the absence of any explicit `Required/RequiredBy` (or equivalent) relationship between the two services, If both services were being brought up as dependencies of something else, the `Before/After` relationships between them would still be applied even though the two services do not depend on each other in any way.

colhom:
I see the confusion between dependencies and ordering dependencies pop up a lot in discussion of systemd units here. @pbx0 @robszumski I highly recommend reading [this section on Before/After](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Before=) of the systemd docs carefully. The crux is:

>  Note that this setting is independent of and orthogonal to the requirement dependencies as configured by Requires=.

robszumski:
Understood, SGTM

colhom:
We have a successful e2e run on this PR, so I'm going to merge it and fix #670

PR #677, opened 2016-09-17
Update getting-started.md

codergr:
Wasn't able to find where I can edit directly the official etcd clustering guide (https://coreos.com/etcd/docs/latest/clustering.html).  For my setup i had to replace etcd with etcd2 to make it work otherwise was complaining of unknown variables.


robszumski:
I'm thinking that the best solution for this is to link to the guide for etcd in containers: https://coreos.com/etcd/docs/latest/docker_guide.html. I will put together a PR to swap out that link. Sound good?

PR #679, opened 2016-09-19
aws: top-level services are runtime: true

### LINKED ISSUE #676: command:start and enable:true are a dangerous pair
@crawford told me this a nasty surprise cloud-config offers it's users: 

`command:start` and `enabled:true` for the same service can lead to it being started prematurely and failing. We have this affliction in our codebase:

https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-controller#L38-L39

https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-worker#L20-L21

\cc @robszumski @cgag , another candidate for the `failed kubelet.service` due to dependencies symptom.

Will fix

colhom:
Fixes #676 

This passes conformance tests: https://jenkins.coreos.systems/view/Kubernetes/job/kubernetes-conformance-aws/131/consoleFull

\cc @crawford @pbx0 

`command: start` and `runtime: true` for all top-level services. 

The only things that are enabled are brought up as dependencies by systemd on behalf of other units (and have no command associated with them).

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -15,6 +15,7 @@ coreos:
   units:
     - name: etcd2.service
       command: start
+      runtime: true
 
     - name: docker.service
       drop-ins:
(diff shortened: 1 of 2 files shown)


pbx0:
Looks good to me, I'll let crawford approve it though since he is familiar with the problem.

crawford:
LGTM

PR #680, opened 2016-09-19
V1.4 updates

pbx0:
Still running some conformance tests and looking for more actionable changes in the release notes. I will update the default hyperkube image once a non pre-release version is out.

Notably, I noticed we can now curl the static assets in yaml and not json.


aaronlevy:
@pbx0 can you also update the addon manifests in the Documentation?

aaronlevy:
lgtm

robszumski:
Are we planning on doing a release to cover these or wait for it to shake out a few days?

This is the only way we get the docs updated cleanly.

pbx0:
@robszumski I imagine we will wait a few days before cutting a coreos-kubernetes release. I guess we'll try to coordinate the doc sync for when we cut the release?

I plan to merge this soon -- just re-running a few tests for a sanity check.

aaronlevy:
FWIW I would lean towards cutting a release as soon as possible. Not sure what waiting really gets us (previous releases are still there). /cc @colhom

PR #681, opened 2016-09-21
docs: mount resolv.conf from host

robszumski:
I discovered that the Vagrant boxes have already been mounting resolv.conf into the kubelet-wrapper:

https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/generic/controller-install.sh#L95-L96

This PR brings the rest of the guides to parity.

fixes https://github.com/coreos/bugs/issues/1228 and https://github.com/coreos/coreos-kubernetes/issues/506

PR #684, opened 2016-09-22
docs: remove erroneous equal sign to fix mount

robszumski:
This fails silently otherwise.

Note that https://github.com/coreos/coreos-kubernetes/pull/681 was correct, but this location wasn't touched in that change.


aaronlevy:
@robszumski mind doing a rebase?

robszumski:
Done in https://github.com/coreos/coreos-kubernetes/pull/762

PR #687, opened 2016-09-25
Allow setting K8s Version as env

johscheuer:
This PR allows users to specify their K8s version without modifing the scripts they just have to set the env variable like `export K8S_VER=v1.3.7_coreos.0` and the scripts will take this version. This allows users to use older/newer versions of K8s without any modification of the scripts. 

All possible K8s versions can be found in https://quay.io/repository/coreos/hyperkube?tag=latest&tab=tags (even if some of these versions are not tested?) 

As an alternativ we could use the [`config.rb`](https://github.com/johscheuer/coreos-kubernetes/blob/master/Documentation/kubernetes-on-vagrant.md#start-the-machines) and add here a field `K8S_VER`.

If you like to add this feature I would add some docs for the [Vagrant setup](https://github.com/johscheuer/coreos-kubernetes/blob/master/Documentation/kubernetes-on-vagrant.md)

### DIFF
--- a/multi-node/generic/controller-install.sh
+++ b/multi-node/generic/controller-install.sh
@@ -5,7 +5,7 @@ set -e
 export ETCD_ENDPOINTS=
 
 # Specify the version (vX.Y.Z) of Kubernetes assets to deploy
-export K8S_VER=v1.3.6_coreos.0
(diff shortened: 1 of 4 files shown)


zbwright:

PR #695, opened 2016-09-28
docs: add missing word in external dns copy

spoonben:
"This hostname be used to provision" to "This hostname *will* be used to provision"


robszumski:
LGTM

PR #697, opened 2016-09-29
aws: workaround for systemd #1312

### LINKED ISSUE #675: kube-aws: decrypt-tls-assets.service failed in a controller node
With the latest master(066d888caa85c00112d891f3c69a8416a59aaee4), it failed and kubelet.service didn't come up.

`systemctl status decrypt-tls-assets` showed:

```
● decrypt-tls-assets.service - decrypt kubelet tls assets using amazon kms
   Loaded: loaded (/etc/systemd/system/decrypt-tls-assets.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Fri 2016-09-16 13:02:33 UTC; 3min 34s ago
  Process: 884 ExecStart=/opt/bin/decrypt-tls-assets (code=exited, status=1/FAILURE)
 Main PID: 884 (code=exited, status=1/FAILURE)
```

`journalctl -u decrypt-tls-assets` showed:

```
Sep 16 13:02:03 ip-10-0-0-50.ap-northeast-1.compute.internal systemd[1]: Starting decrypt kubelet tls assets using amazon kms...
Sep 16 13:02:04 ip-10-0-0-50.ap-northeast-1.compute.internal sudo[890]:     root : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/bin/rkt run --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=fal
Sep 16 13:02:04 ip-10-0-0-50.ap-northeast-1.compute.internal sudo[890]: pam_unix(sudo:session): session opened for user root by (uid=0)
Sep 16 13:02:04 ip-10-0-0-50.ap-northeast-1.compute.internal decrypt-tls-assets[884]: image: using image from file /usr/lib64/rkt/stage1-images/stage1-coreos.aci
Sep 16 13:02:18 ip-10-0-0-50.ap-northeast-1.compute.internal decrypt-tls-assets[884]: image: searching for app image quay.io/coreos/awscli
Sep 16 13:02:33 ip-10-0-0-50.ap-northeast-1.compute.internal decrypt-tls-assets[884]: run: discovery failed
Sep 16 13:02:33 ip-10-0-0-50.ap-northeast-1.compute.internal systemd[1]: decrypt-tls-assets.service: Main process exited, code=exited, status=1/FAILURE
Sep 16 13:02:33 ip-10-0-0-50.ap-northeast-1.compute.internal systemd[1]: Failed to start decrypt kubelet tls assets using amazon kms.
Sep 16 13:02:33 ip-10-0-0-50.ap-northeast-1.compute.internal systemd[1]: decrypt-tls-assets.service: Unit entered failed state.
Sep 16 13:02:33 ip-10-0-0-50.ap-northeast-1.compute.internal systemd[1]: decrypt-tls-assets.service: Failed with result 'exit-code'.
```

A work-around for me was to run

```
sudo systemctl restart decrypt-tls-assets
sudo systemctl restart kubelet
```

and then waiting several minutes until `curl localhost:8080` succeeds.
I'm not really sure but the fix for #666 caused another issue?

### LINKED ISSUE #682: why is docker.service not automatically restarting
https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/aws/pkg/config/templates/cloud-config-worker#L9

here it seems like docker service doesn't have a restart policy. I am sure I am missing something, and I apoligize if this isn't the best place to discuss this, but I did experience some problem in my cluster when, for any reason, docker died, and I had to automatically restart it.

mumoshu:
On top of https://github.com/coreos/coreos-kubernetes/issues/682#issuecomment-248692751, replaces a dependency from kubelet to the oneshot service `decrypt-tls-assets` with an ExecStartPre in the kubelet service because systemd doesn't seem to restart failed `decrypt-tls-assets` services that way.

Also, I've changed where to add `ExecStartPre`, instead of `[Unit]` sections as done in https://gist.github.com/spacepluk/a14f10cfed3756c0f1f079e73cdc6c9a#gistcomment-1885799, to `[Service]` sections as done in this PR.

Mainly purposed to fix #675 which I've encountered earlier, but would fix #682 too.

Although my cluster brought up via kube-aws with this change is working without any problem,
I have not yet a chance to reproduce #682 to test this.
Is anyone aware of a handy way to reproduce it?

cc @cgag @spacepluk

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -22,10 +22,10 @@ coreos:
         - name: 40-flannel.conf
           content: |
             [Unit]
-            Requires=flanneld.service
-            After=flanneld.service
+            Wants=flanneld.service
             [Service]
             EnvironmentFile=/etc/kubernetes/cni/docker_opts_cni.env
+            ExecStartPre=systemctl is-active flanneld.service
 
     - name: flanneld.service
       drop-ins:
@@ -39,6 +39,8 @@ coreos:
       command: start
       runtime: true
       content: |
+        [Unit]
+        Wants=docker.service
         [Service]
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
@@ -52,6 +54,8 @@ coreos:
         --mount volume=stage,target=/tmp \
         --volume var-log,kind=host,source=/var/log \
         --mount volume=var-log,target=/var/log"
+        ExecStartPre=systemctl is-active docker.service
+        ExecStartPre=/opt/bin/decrypt-tls-assets
         ExecStartPre=/usr/bin/mkdir -p /var/log/containers
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers=http://localhost:8080 \
@@ -91,12 +95,12 @@ coreos:
         [Unit]
         Description=Load rkt stage1 images
         Documentation=http://github.com/coreos/rkt
-        Requires=network-online.target
-        After=network-online.target
+        Wants=network-online.target
         Before=rkt-api.service
         [Service]
         Type=oneshot
         RemainAfterExit=yes
+        ExecStartPre=systemctl is-active network-online.target
         ExecStart=/usr/bin/rkt fetch /usr/lib/rkt/stage1-images/stage1-coreos.aci /usr/lib/rkt/stage1-images/stage1-fly.aci  --insecure-options=image
         [Install]
         RequiredBy=rkt-api.service
@@ -109,8 +113,7 @@ coreos:
       content: |
         [Unit]
         Description=Calico per-host agent
-        Requires=network-online.target
-        After=network-online.target
+        Wants=network-online.target
 
         [Service]
         Slice=machine.slice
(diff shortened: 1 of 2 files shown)


dghubble:
I think this takes the `systemctl is-active some.service` trick/hack too far. First, that only works if the service itself has restarts, otherwise some of these changes "hope" the dependency is satisfied and fail if not. `Requires=` or `Wants=` are still the right way to do this.

The is-active flanneld was done on clusters which required a remote etcd to come online before flannel starts and thus kubelet can start if I recall correctly. Starting etcd -> flannel -> kubelet should work here.

* https://github.com/systemd/systemd/issues/1312
* https://github.com/coreos/coreos-baremetal/issues/253

mumoshu:
@dghubble Thanks again for your feedback!

I guess things became a bit clearer for me now.

To sum up,

* [Modification to load-rkt-stage1.service](https://github.com/coreos/coreos-kubernetes/pull/697/files#diff-071e0ee4ae8c6adf4865f25d851f30c9R80) seems meaningless because it is a oneshot service and won't try to restart when ExecStartPre failed, by its nature
* Can `network-online.target` fail *transiently*? Maybe no. Then, modifications to [calico-node.service](https://github.com/coreos/coreos-kubernetes/pull/697/files#diff-071e0ee4ae8c6adf4865f25d851f30c9L89) seems to be meaningless in practice.
* Modification to kubelet.service would be useful according to coreos-baremetal#253
* Modification to flanneld.service would be useful according to https://github.com/coreos/coreos-kubernetes/issues/710
* Modification to docker.service would be useful according to https://github.com/coreos/coreos-kubernetes/issues/682

aaronlevy:
I'm going to close this PR, as the kube-aws work has moved to its own top-level repository here: https://github.com/coreos/kube-aws . If this is still something you would like to be merged, please open a PR under the new repo.

PR #703, opened 2016-10-03
Make etcd prefix configurable

gitoverflow:


aaronlevy:
Sorry for the super delayed response.

We try not to directly expose too many configuration options -- as there are many, so keeping these options to a minimum makes the maintenance of this tool much easier. I'm going to close this PR, as I don't believe this is a commonly needed configuration change (and is relatively simple to change the script if necessary for your own uses.

PR #715, opened 2016-10-06
Mark sections as optional in deploy-workers.md

stephen144:
k8s/deploy-workers.md: Add "(optional)" to headers of optional sections

This is mainly so that it matches the deploy-master document. Setting up Calico is optional. As a newb, I had to go back to the master instructions and make sure that these parts were optional for the workers as well.

### DIFF
--- a/Documentation/deploy-workers.md
+++ b/Documentation/deploy-workers.md
@@ -120,7 +120,7 @@ RestartSec=10
 WantedBy=multi-user.target
 ```
 
-### Set Up the CNI config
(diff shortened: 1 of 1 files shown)


pbx0:
LGTM as far as parity with the deploy-master documentation. Using the Calico network policy is definitely optional. 

However, these docs also need to be updated to reflect that we now use the CNI plugin to configure the network with flannel by default as this method works with both rkt and docker as a runtime. Will file a bug.

PR #717, opened 2016-10-07
Add name tags where applicable

c-knowles:
Not everything is tagged with a name which means the AWS console gets filled with empty named resources, this helps fix it up.


c-knowles:
Replaced by https://github.com/coreos/kube-aws/pull/36.

PR #722, opened 2016-10-12
[WIP] Fix dns addon docs

### LINKED ISSUE #720: CoreOS + Kubernetes Step By Step: DNS addon is broken
Step 5 of the CoreOS + Kubernetes Step By Step guide, is incomplete. The DNS addon created there cannot connect to the API server because it is missing TLS config. See [this](https://github.com/kubernetes/kubernetes/issues/32480) Kubernetes issue. I was able to get DNS working by following [this](https://github.com/kubernetes/kubernetes/issues/32480#issuecomment-249889480) comment in the linked issue. It seems that the kubeDNS pod should be using a `kubeconfig` file.

API server log shows a lot of:

> Unable to authenticate the request due to an error: crypto/rsa: verification error

While the kubeDNS log is filled with:

> reflector.go:136] Failed to list *api.Service: the server has asked for the client to provide credentials (get services)
> reflector.go:136] Failed to list *api.Endpoints: the server has asked for the client to provide credentials (get endpoints)

ElijahCaine:
Fixes #720 

Implemented a quick docs fix based on the [issue comment](https://github.com/kubernetes/kubernetes/issues/32480#issuecomment-249889480) referenced in the above issue.  I have not tested these changes yet, I'll get to that soon.

If you've got some expertise in this area any and all feedback is welcome.

### DIFF
--- a/Documentation/deploy-addons.md
+++ b/Documentation/deploy-addons.md
@@ -9,6 +9,7 @@ First create `dns-addon.yml` on your local machine and replace the variable. The
 [k8s-dns]: http://kubernetes.io/docs/admin/dns.html
 
 * Replace `${DNS_SERVICE_IP}`
+* Replace `${KUBERNETES_SERVICE_HOST}`
 
 **dns-addon.yml**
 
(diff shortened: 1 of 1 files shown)


pieterlange:
I'm not sure this fixes anything - or if it even fixes the root cause of #720.

#720 mentions kubernetes issue https://github.com/kubernetes/kubernetes/issues/32480 which is triggered by missing `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` variables inside the kube-dns pods. 

It seems to me #720 is caused by incorrect serviceaccount tokens (re-rolling a cluster against an existing etcd dataset would do that for example) but the issue is a little too light on details to make any conclusions.

ElijahCaine:
@pieterlange Thank you for the feedback.  I'll look into the issue more and see if we can figure out a fix for the root cause of #720.

aaronlevy:
Agreed - the dns addon should not need these changes (and should be able to use the service account injected into the pod). Closing this, but please let me know if this should be re-opened.

PR #725, opened 2016-10-12
contrib: fix bump-version script on macOS

felixbuenemann:
This fixes non-POSIX conformant usage of `expr` that is only supported
in the GNU/Linux variant and caused syntax errors on macOS 10.12.1.

### DIFF
--- a/contrib/bump-version.sh
+++ b/contrib/bump-version.sh
@@ -4,7 +4,7 @@
 # the CURRENT_VERSION to the TARGET_VERSION. This is meant as a helper - but
 # probably should still double-check the changes are correct
 
-if [ $# -ne 1 ] || [ `expr match $1 ".*_.*"` == 0 ]; then
+if [ $# -ne 1 ] || [ `expr $1 : ".*_.*"` == 0 ]; then
     echo "USAGE: $0 <target-version>"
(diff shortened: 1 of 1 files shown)


contrib/bump-version.sh
```diff
@@ -13,8 +13,8 @@ fi
 CURRENT_VERSION=${CURRENT_VERSION:-"v1.4.0_coreos.0"}
 TARGET_VERSION=${1}
 
-CURRENT_VERSION_BASE=${CURRENT_VERSION:0:`expr index $CURRENT_VERSION _` -1}
-TARGET_VERSION_BASE=${TARGET_VERSION:0:`expr index $TARGET_VERSION _` -1}
+CURRENT_VERSION_BASE=${CURRENT_VERSION%%_*}
+TARGET_VERSION_BASE=${TARGET_VERSION%%_*}
```
felixbuenemann: This deletes everything from the first underscore using bash variable substitutions.

felixbuenemann:
I have only tested the changes on macOS, so it would be great if someone could run this on linux before merging, although I'm pretty confident it'll work fine (I compared the linux manpage for expr).

pbx0:
Thanks for doing this, I tested it on a recent fedora and it works for me.

Issue #692, opened 2016-09-28
Conformance tests build fail


zihaoyu:
Ran on 45a0122f49d97e84ab262a1935e0328c3449c5d1 commit against a 1.4.0 Kubernetes cluster.

Output:

``````
Cloning into '/home/core/go/src/k8s.io/kubernetes'...
image: using image from file /usr/lib64/rkt/stage1-images/stage1-coreos.aci
image: remote fetching from URL "docker://golang:1.6.2"
Downloading sha256:c6072700a24 23.7 MB / 42.5 MB
Downloading sha256:5c90d4a2d1a 6.44 MB / 51.4 MB
Downloading sha256:ab30c63719b 9.76 MB / 18.5 MB
Downloading sha256:b0850b14d0d 1.35 KB / 1.35 KB
Downloading sha256:0ffc1204e0a 196 KB / 56.9 MB
Downloading sha256:d8a921df8ce 32.3 KB / 84.9 MB
Downloading sha256:d34bc98c277 0 B / 122 B
Downloading sha256:c6072700a24 23.7 MB / 42.5 MB
Downloading sha256:5c90d4a2d1a 6.44 MB / 51.4 MB
Downloading sha256:ab30c63719b 9.76 MB / 18.5 MB
Downloading sha256:b0850b14d0d 1.35 KB / 1.35 KB
Downloading sha256:0ffc1204e0a 196 KB / 56.9 MB
Downloading sha256:d8a921df8ce 32.3 KB / 84.9 MB
Downloading sha256:d34bc98c277 122 B / 122 B
Downloading sha256:c6072700a24 30.8 MB / 42.5 MB
Downloading sha256:5c90d4a2d1a 11.9 MB / 51.4 MB
Downloading sha256:ab30c63719b 16.1 MB / 18.5 MB
Downloading sha256:b0850b14d0d 1.35 KB / 1.35 KB
Downloading sha256:0ffc1204e0a 11.3 MB / 56.9 MB
Downloading sha256:d8a921df8ce 6.14 MB / 84.9 MB
Downloading sha256:d34bc98c277 122 B / 122 B
Downloading sha256:c6072700a24 33.9 MB / 42.5 MB
Downloading sha256:5c90d4a2d1a 14.5 MB / 51.4 MB
Downloading sha256:ab30c63719b 18.5 MB / 18.5 MB
Downloading sha256:b0850b14d0d 1.35 KB / 1.35 KB
[...]

PR #726, opened 2016-10-13
contrib: get go-bindata for conformance test dep

pbx0:
same as: https://github.com/kubernetes-incubator/bootkube/pull/153
fixes https://github.com/coreos/coreos-kubernetes/issues/692


aaronlevy:
lgtm

PR #730, opened 2016-10-15
add api_server_count param

rdtr:
It's not documented on kubernetes official site, but we need to add a `apiserver-count` flag when considering HA mode of msaters (a number of master nodes > 1).
https://github.com/kubernetes/kubernetes/issues/24327


aaronlevy:
This flag actually leads to some less than desirable results. For example, if one of your api-servers is down, half of your requests will fail. The proxy only does a simple round-robin, with no healthchecks of the endpoints, and the endpoint of a "dead" api-server will not automatically be removed.

Upstream there are some other options being discussed, but this probably isn't a direction we want to take.

Issue #620, opened 2016-08-15
Cannot pull private AWS ECR image from controller node


c-knowles:
I'm not sure if this is a kube-aws issue or a Kubernetes issue so hopefully someone can shed some light on it. I'm using `kube-aws` 0.8.0 and when launching a DaemonSet with a private AWS ECR image it seems that the controller node won't pull the image but all worker nodes will.

The error I get in the pod logs is:

```
container "private_name" in pod "private_name-5wafj" is waiting to start: image can't be pulled
```
#### Steps to reproduce
1. Launch a cluster using kube-aws, default configuration seems to have the same issue although we are using the multi AZ setup
2. Add a DaemonSet with a single container referencing a private AWS ECR image from the same AWS account

I will try with the latest kube-aws version soon just to check whether it's still an issue.
#### Current workaround

None that I know of that will get the private image running on the controller. Currently I added a label to the workers and then I used `nodeSelector` in the DaemonSet to avoid launching it on the controller at all. In my case the daemon is monitoring related so I need it to run on the controller as well.

whereisaaron:
Hi, I haven't seen the a problem with Daemonsets specifically, but when I did have issues with ~0.7 I found the journelctl logs on the nodes pretty informative. They normally said _why_ the image couldn't be pulled, e.g. DNS or authentication.

Does the image pull work if you manually specify your own ImagePullSecret?

c-knowles:
Good ideas, thanks.

I checked the controller journelctl logs and found this:
```
Aug 15 05:04:27 ip-10-0-0-50.eu-west-1.compute.internal kubelet-wrapper[1508]: I0815 05:04:27.961740    1508 reconciler.go:253] MountVolume operation started for volume "kubernetes.io/secret/default-token-icqo6" (spec.Name: "default-token-icqo6") to pod "792283c9-62a0-11e6-8d91-0a90a4438ea5" (UID: "792283c9-62a0-11e6-8d91-0a90a4438ea5"). Volume is already mounted to pod, but remount was requested.
[...]

PR #731, opened 2016-10-17
Add existing ECR policy to controller IAM role

c-knowles:
Goes some way to resolve https://github.com/coreos/coreos-kubernetes/issues/620


c-knowles:
Replaced by coreos/kube-aws/pull/35

PR #733, opened 2016-10-18
all: kubelet.service deletes pod on stop

pbx0:
Currently doesn't work stop command fails with:

```
unable to create process 1635 instance: open /proc/1635: no such file or directory
stop: error stopping "6b3f4d0b-b4c8-4ac9-a6b7-9ef92f236b09": exit status 1
stop: failed to stop 1 pod(s)
```

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-controller
+++ b/multi-node/aws/pkg/config/templates/cloud-config-controller
@@ -51,8 +51,11 @@ coreos:
         --volume stage,kind=host,source=/tmp \
         --mount volume=stage,target=/tmp \
         --volume var-log,kind=host,source=/var/log \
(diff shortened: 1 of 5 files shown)


lucab:
I reproduced with a single-node vagrant, and I see that the kubelet.service has a `KillMode=process`. This means that even if the main parent process is gone (as rkt stop shows), its children are still alive so the pod is still partly running and cannot be rm-ed.

pbx0:
@lucab Thanks for figuring that out. I updated the PR to include `KillMode=control-group` and things seem to work fine now -- I'm happy with this solution.

aaronlevy:
I believe this landed in: https://github.com/coreos/coreos-kubernetes/pull/759

PR #740, opened 2016-10-24
Delete all pods from the node before instance reboot, shutdown or terminate

camilb:
Maybe someone  find this useful. For me it works well with auto-scaling but also allow pods to be rescheduled faster on other nodes before reboots.

Behaviour 
On Stop: first set node "unschedulable=true", then delete all the pods from all namespaces
OnStart: set node "unschedulable=false"

### DIFF
--- a/multi-node/aws/pkg/config/templates/cloud-config-worker
+++ b/multi-node/aws/pkg/config/templates/cloud-config-worker
@@ -55,6 +55,23 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+    - name: clean-reboot.service
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Delete all pods from the node before instance reboot, shutdown or terminate
+        Requires=decrypt-tls-assets.service
+        After=multi-user.target
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/bin/sh -c "/opt/bin/node-status-change"
+        ExecStop=/bin/sh -c "/opt/bin/delete-pods-before-reboot"
+
+        [Install]
+        WantedBy=worker-reboot.target
+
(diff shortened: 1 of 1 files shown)


rochacon:
@camilb Does this work like an `kubectl drain`? If yes, Would it be possible to use the `kubectl drain` command for this?
I'm looking for a solution for this too. I was thinking in adding the `kubectl drain` in an separate SystemD unit hooked in the shutdown process (e.g. `Before=shutdown.target`), but I haven't tested this yet.

camilb:
@rochacon Yes, has the same functionality. I was also thinking to use `kubectl`first but found it simpler to use the API instead.

colhom:
Hello Kubernetes Community,

Future work on kube-aws will be moved to [a new dedicated repository](https://github.com/coreos/kube-aws). @mumoshu will be running point on maintaining that repository- please move all issues and PRs over there as soon as you can. We will be halting active development on the AWS portion of this repository in the near future. We will continue to maintain the vagrant single and multi-node distributions in this repository, along with our [hyperkube container image](https://quay.io/coreos/hyperkube) image.

A community announcement to end users will be made once the transition is complete. We at CoreOS ask that those reading this message avoid publicizing/blogging about the transition until the official annoucement has been made to the community in the next week.

The new dedicated kube-aws repository already has the following features merged in:
* Discrete etcd cluster
* HA control plane
* Cluster upgrades
* Node draining/cordoning

If anyone in the Kubernetes community would like to be involved with maintaining this new repository, find @ chom and/or @ mumoshu on the Kubernetes slack in the #sig-aws channel or via direct message.

~CoreOS Infra Team

camilb:
Closing this, since there is a similar feature in `kube-aws` 0.9 using `kubectl drain`.

PR #745, opened 2016-10-27
docs: link to shell scripts on GH

robszumski:
> Hi guys. Broken link on https://coreos.com/kubernetes/docs/latest/kubernetes-on-generic-platforms.html to download worker-install.sh (https://coreos.com/kubernetes/docs/multi-node/generic/worker-install.sh)

This PR will reference these scripts on Github to 1) fix these links and 2) its probably better than serving shell scripts off of our domain.


aaronlevy:
lgtm

PR #748, opened 2016-11-02
fix check for min CoreOS version required when using rkt

dimpavloff:
There is a minimum version of CoreOS required when using rkt and not
using the alpha channel. This commit fixes the semver comparison

### DIFF
--- a/multi-node/aws/pkg/config/config.go
+++ b/multi-node/aws/pkg/config/config.go
@@ -206,9 +206,8 @@ func (c Cluster) Config() (*Config, error) {
 	return &config, nil
 }
 
-// isMinImageVersion will return true if the supplied version is greater then
-// or equal to the current CoreOS release indicated by the given release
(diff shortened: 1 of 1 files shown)


colhom:
@dimpavloff could you move this PR over to https://github.com/coreos/kube-aws?

mumoshu:
@dimpavloff Sorry to bother you but If your fix is intended the same with https://github.com/coreos/kube-aws/commit/46054e97104071586669d1b69f68e028993aed49, it is already merged there.

AFYI, rkt+stable is supported since several days ago
https://github.com/coreos/kube-aws/issues/21

Thanks for your contribution anyway!

dimpavloff:
Hi guys, thanks for looking! Yes, https://github.com/coreos/kube-aws/commit/46054e97104071586669d1b69f68e028993aed49 fixes the same problem

PR #750, opened 2016-11-04
docs: Remove extra instance deployment

joshrosso:
If you follow this guide, it will only provision a worker node and
controller node. In order to deploy 2 workers, you'd need to modify
workerCount in the cluster.yaml file before rendering the
CloudFormation template.

This commit corrects the output to what the user sees. Another question
is whether this guide should have users increase that value so their
sample deployment features > 1 worker.


robszumski:
LGTM. I think the idea is that this gets you the most minimal cluster, since these machines cost money.

Note that these docs and all AWS functionality will be moving to https://github.com/coreos/kube-aws soon

aaronlevy:
lgtm - but also, the aws code is moving to its own top-level repository: github.com/coreos/kube-aws -- @joshrosso I might suggest making the same change there.

Issue #749, opened 2016-11-04
Only 1 worker is launched in a multi-node vagrant setup


F21:
I am using the multi-node vagrant setup and have configured it to create 3 worker nodes. I can see that the nodes are launched and can ssh into all of them. However, kubernetes only sees 1 worker:

```
$ kubectl get nodes
NAME           STATUS                     AGE
172.17.4.101   Ready,SchedulingDisabled   7m  <--- controller
172.17.4.201   Ready                      7m               <--- worker
```

This is also the case when I check the web dashboard.

I am using `core-kuberntes` master.

This is my `config.rb`:
```ruby
$update_channel="alpha"

$controller_count=1
$controller_vm_memory=512

$worker_count=3
$worker_vm_memory=2048

$etcd_count=3
$etcd_vm_memory=512
```

I also set `USE_CALICO=true` and `K8S_VER=v1.4.5_coreos.0` in `worker-install.sh` and `controller-install.sh`. 

I am running this in VirtualBox 5.1.8 on Windows 10 64-bit. To get vagrant up working on Windows, I applied this commit to master: https://github.com/ah45/coreos-kubernetes/commit/3ca05ebbf21610401781dc2410293636e20d6161

F21:
It seems to be pretty random whether a worker will be initialized correctly or not. I just destroyed the cluster and ran `vagrant up` again. This time, I can see 2 workers downloading the images and setting up, but the last worker did nothing. Is there anyway to check to see why the last worker did not install?

Using `journalctl` just showed `no entries`:
```
core@w3 ~ $ journalctl -u kubelet
-- No entries --
```

PR #756, opened 2016-11-09
Add a 20 minute timeout for flanneld, so that it doesn't fail on slow connections.

F21:
While troubleshooting #749, I noticed that flanneld fails to start because it didn't finish downloading the flannel image before the timeout. Because flanneld fails, the VM wasn't provisioned.

After adding a longer timeout, the VM provisions correctly.


zbwright:

PR #757, opened 2016-11-10
Change kubelet download url according to k8s faq

pizzarabe:
According to https://github.com/kubernetes/kubernetes/wiki/User-FAQ#where-should-i-get-kubernetes we should use the github release page to get our binaries instead of storage.googleapis.com/kubernetes.

This will increase the download size, because we will get the whole k8s release but `We recommend you do not use binaries from http://storage.googleapis.com/kubernetes, as these are neither official releases nor maintained.`

### DIFF
--- a/Documentation/configure-kubectl.md
+++ b/Documentation/configure-kubectl.md
@@ -10,18 +10,18 @@ To quickly launch a cluster, follow these guides for [AWS][kube-aws], [Vagrant][
 
 ## Download the kubectl Executable
 
-Download `kubectl` from the Kubernetes release artifact site with the `curl` tool.
+Download `kubectl` from the Kubernetes github release site with the `curl` tool.
 
 The linux `kubectl` binary can be fetched with a command like:
 
 ```sh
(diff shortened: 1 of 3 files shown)


aaronlevy:
Given that the release tarball is over 1GB, I don't think it's reasonable to expect users to download the entire thing just to get the kubectl.

I've been using the binaries we have linked for over a year, and haven't run into issues -- but that note is a little worrying. It might mean that we need to carry our own releases of kubectl if we don't trust that endpoint.

euank:
The upstream docs also reference the kubectl url we use as part of the [minikube docs](http://kubernetes.io/docs/getting-started-guides/minikube/#installation) and as part of the [kubectl install docs](http://kubernetes.io/docs/user-guide/prereqs/).

I think we're safe for now.

Here's the upstream issue that will, once resolved, allow us to use a "more official" url (more official than the one the official docs use): https://github.com/kubernetes/kubernetes/issues/28631

zbwright:

PR #764, opened 2016-11-18
Allow exposing of docker and kubectl installation

martinhoefling:
This makes docker accessible for development and enables the usage
of kubectl from the inside of the vm.


robszumski:
Question for the group: should we enable this for the multi-machine vagrant? Seems like yes?

pbx0:
@robszumski Sure, any change to single-node should generally be in multinode as well. 

@martinhoefling Could you elaborate on what you mean by exposing docker and the motivation behind it? Installing kubectl by default is certainly useful.

martinhoefling:
@robszumski - if it is considered useful - can port these options to the multi node setup as well. 
@pbx - The additional socket file exposes the docker control socket via tcp so that it can be used from the host system. For me, this is useful since with that setup, I can use docker directly from my mac.

In conjunction with the guest install of `kubectl`, I can do the following:

Set up a docker registry for the cluster - expose it via `kubectl` on 'localhost' on the coreos machine.
Setting `DOCKER_HOST` to the vm ip allows me to use docker and push / pull to the 'localhost' registry and at the same time use those images from kubernetes.

atamon:
Being able to interact with the docker engine running inside the node is great for development indeed! Great initiative!

I'm trying to run your fork @martinhoefling. Not being that familiar with Vagrant I'm struggling to figure out how to set EXPOSE_DOCKER=true in a way that it propagates to the `user-data` script. Any tips?

Thanks in advance

martinhoefling:
I did not find a way to achieve this as well. So I just modified the file. Keeping everything in git gives you an overview what is modified.

atamon:
Thanks for the quick reply!

I ended up doing that as well. But it would be great if there was an option one could give to `vagrant up` that would override the configuration. Keeping your local modifications in sync with the origin becomes cumbersome in the long run imo :)

Great work though, I can now connect to the coreos docker daemon by setting
```sh
export DOCKER_HOST="tcp://172.17.4.99:2375"
```

shaundomingo:
👍 This would be very handy to have in coreos:master.

My use case is also local dev - minikube isn't cutting it for me and being able to connect directly into the docker daemon from my host + run kubectl on the node are the missing pieces!

zbwright:

Issue #719, opened 2016-10-07
docs: reflect new way of configuring flannel with CNI


pbx0:
Explained in this comment: https://github.com/coreos/coreos-kubernetes/pull/715#issuecomment-252355541

ElijahCaine:
@pbx0 I'd be up for resolving this if it's still an issue.  Do you think you can elaborate on the problem & point me in the direction of any relevant stuff I should read to resolve this?

pbx0:
@ElijahCaine Sure, that would be great if you'd like to do it. Let me try to provide the context, but, bear with me because explaining the CNI jargon, can get a little confusing.

Mainly I filed the issue because prior to https://github.com/coreos/coreos-kubernetes/commit/7ad134744b75d0a8a75af165afeb107ea245f87a we did not use a kubernetes network plugin at all. We directly configured docker to use the flannel network by default for all pods. The kubernetes CNI plugin was optional and only used to setup Calico, which allows more complex control of the overlay network. 

However, since https://github.com/coreos/coreoskubernetes/commit/7ad134744b75d0a8a75af165afeb107ea245f87a we always use the CNI kubernetes network plugin by passing it a default a configuration that just uses plain flannel. This is nice because it works for both rkt and docker without having to change anything. Calico is still optional and uses a different cni plugin.

Thus, when the docs from #715 say `### Set Up the CNI config (optional)` that isn't quite true because CNI is not optional anymore and the docs should reflect the new way in which we setup flannel (via CNI). 

I'm not sure how far down this path you want to go as our setup scripts continue to diverge overtime from the manual docs and its difficult to keep them in sync as the setup scripts support more features and configurations by default. When rkt support was added to the setup scripts for example, it included quite a few sections of `if RUNTIME=rkt setup this thingy` which I think would make the manual docs more confusing and error prone then they already are. The question I have is: Should the manual docs should remain a first class k8s setup method or if it should just be an educational introduction to kubernetes that will likely help with future debugging efforts?

ElijahCaine:
Whoa. Well that is a whole bunch of complicated I wasn't expecting.

I'll dig around this issue and see if it's worth it to update the manual doc to more accurately reflect the new world order in the scripts.  If not I'll talk with some people about maybe migrating this doc toward an educational focus in favor of being an actual instruction manual, since that seems like it might be more sane as we develop things going forward.

robszumski:
[...]

PR #767, opened 2016-11-26
Docs: Sync Kubernetes step-by-step with generic scripts

alekssaul:
There seem to be some updates to the scripts that are not covered in the Kubernetes Step by Step documentation. Updates documentation to add the changes that are applied in generic scripts folder.

Namely:
- flannel drop-in uses environment file /etc/kubernetes/cni/docker_opts_cni.env
- changes to the kubelet.service , which also require /opt/bin/host-rkt creation and /etc/kubernetes/cni folder artifacts
- kube-proxy annotation and dbus mount
- api-server livenessprobe

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -109,27 +109,39 @@ Note that the kubelet running on a master node may log repeated attempts to post
 
 ```yaml
 [Service]
+Environment=KUBELET_VERSION=${K8S_VER}
+Environment=KUBELET_ACI=quay.io/coreos/hyperkube
+Environment="RKT_OPTS=--uuid-file-save=/var/run/kubelet-pod.uuid \
+  --volume dns,kind=host,source=/etc/resolv.conf \
+  --mount volume=dns,target=/etc/resolv.conf \
+  --volume rkt,kind=host,source=/opt/bin/host-rkt \
+  --mount volume=rkt,target=/usr/bin/rkt \
+  --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
+  --mount volume=var-lib-rkt,target=/var/lib/rkt \
+  --volume stage,kind=host,source=/tmp \
+  --mount volume=stage,target=/tmp \
+  --volume var-log,kind=host,source=/var/log \
+  --mount volume=var-log,target=/var/log"
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
 ExecStartPre=/usr/bin/mkdir -p /var/log/containers
-
-Environment=KUBELET_VERSION=${K8S_VER}
-Environment="RKT_OPTS=--volume var-log,kind=host,source=/var/log \
-  --mount volume=var-log,target=/var/log \
-  --volume dns,kind=host,source=/etc/resolv.conf \
-  --mount volume=dns,target=/etc/resolv.conf"
-
+ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/run/kubelet-pod.uuid
 ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --api-servers=http://127.0.0.1:8080 \
-  --network-plugin-dir=/etc/kubernetes/cni/net.d \
-  --network-plugin=${NETWORK_PLUGIN} \
   --register-schedulable=false \
+  --cni-conf-dir=/etc/kubernetes/cni/net.d \
+  --network-plugin=cni \
+  --container-runtime=docker \
+  --rkt-path=/usr/bin/rkt \
+  --rkt-stage1-image=coreos.com/rkt/stage1-coreos \
   --allow-privileged=true \
   --pod-manifest-path=/etc/kubernetes/manifests \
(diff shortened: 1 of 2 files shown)


Documentation/deploy-master.md
```diff
@@ -81,6 +81,34 @@ Again, we will use a [systemd drop-in][dropins]:
...
+**/etc/kubernetes/cni/docker_opts_cni.env**
+
+```yaml
+DOCKER_OPT_BIP=""
+DOCKER_OPT_IPMASQ=""
+```
+
+If using Flannel for networking. 
+
+*Note:* Do not use below if you intend to use Callico for networking
```
robszumski: Calico

Documentation/deploy-master.md
```diff
@@ -81,6 +81,34 @@ Again, we will use a [systemd drop-in][dropins]:
...
+Create the Docker CNI Options file:
+
+**/etc/kubernetes/cni/docker_opts_cni.env**
+
+```yaml
+DOCKER_OPT_BIP=""
+DOCKER_OPT_IPMASQ=""
+```
+
+If using Flannel for networking.
```
robszumski: what does this refer to? above or below?

Documentation/deploy-master.md
robszumski: I try not to use "Note:" type stuff. Is there a way to rephrase, maybe combining with the "If using flannel" comment?

Documentation/deploy-workers.md
```diff
@@ -71,6 +71,33 @@ Create `/etc/systemd/system/docker.service.d/40-flannel.conf`
...
+**/etc/kubernetes/cni/docker_opts_cni.env**
+
+```yaml
+DOCKER_OPT_BIP=""
+DOCKER_OPT_IPMASQ=""
+```
+
+If using Flannel for networking. 
+
+*Note:* Do not use below if you intend to use Callico for networking
```
robszumski: Same comments from the other file are valid here

Documentation/deploy-workers.md
```diff
@@ -179,30 +240,36 @@ spec:
...
+      path: "/etc/kubernetes/worker-kubeconfig.yaml"
+  - name: "etc-kube-ssl"
+    hostPath:
+      path: "/etc/kubernetes/ssl"
+  - hostPath:
+      path: /var/run/dbus
+    name: dbus
 ```
 
+
```
robszumski: stray line

alekssaul:
Thanks @robszumski , made the changes per your comments. Please note that I've only covered changes with flannel as the transport and haven't tested end-to-end for Calico pieces.

pbx0:
@alekssaul Thanks for working on this! Will solve: https://github.com/coreos/coreos-kubernetes/issues/719

One important thing to note here is that you've brought in a lot of changes that are only necessary if the kubelet's `--containter-runtime` flag is set to `rkt`. Since you've already done most of the work of making this guide rktnetes compatible its probably worth adding a section explaining that the kubelet's runtime can be switched or removing the rkt specific changes/hacks until we are sure we are ready to support it in this manual guides. Mainly, all the extra mount flags and host-rkt stuff aren't needed if we are just supporting docker as a runtime.

alekssaul:
Thanks @pbx0, I mostly wanted to sync in the script changes into the documentation. 

In order to move forward, it sounds like it makes sense to add a section such as "Expiremental rkt as the runtime" to highlight flag changes to setup rktnetes.

> all the extra mount flags and host-rkt stuff aren't needed if we are just supporting docker as a runtime.

Would it be ok to keep the kubernetes manifests as is? I believe the kubernetes scripts sets up host-rkt and it's required volumes regardless of the runtime. I am hoping we can apply the same to the documentation, and not maintain a seprate kubelet.service for rkt vs docker.

pbx0:
> Would it be ok to keep the kubernetes manifests as is? I believe the kubernetes scripts sets up host-rkt and it's required volumes regardless of the runtime. I am hoping we can apply the same to the documentation, and not maintain a seprate kubelet.service for rkt vs docker.

@alekssaul 
The extra mount-flags and host-rkt hack (this won't always be needed) aren't necessary for docker based setups but won't hurt docker installations. It may be confusing to readers of the guide. I think we can avoid confusion simply by mentioning that the guide is intended to support both docker and rkt as the runtime and includes configuration for both runtimes despite the fact that only one will be used by kubernetes. All one needs to do is to change the `--container-runtime` flag.

So if you want to support rkt, the example scripts become a little more confusing to users because of the hacks they contain, but we support both runtimes which is great. We are missing a few rkt specific requirements here such as the `rkt-api.sevice` and the `load-rkt-stage1.service`. These services are normally gated on the `CONTAINER_RUNTIME` env in the scripts but won't hurt docker installations either. 

Otherwise, getting the CNI stuff in is a great start and you could remove the rkt specific aspects for now and include rkt support in a separate PR. The choice is yours! I'm happy to help review, but I won't have time to test the guides manually.

alekssaul:
@pbx0 : removed the rkt specific mounts/hacks. Tested this with docker + flannel (no calico components)

PR #768, opened 2016-12-01
Updating Calico to hosted install

heschlie:
In order to get the hosted intall to work we had to make some changes to the deployment, the most disruptive being that we need to write to the hyperkubes `/opt/cni/bin` directory, so when Calico is enabled we mount that directory to the host.  Our calico/cni:v1.5.2 image deploys flannel, host-local, and loopback CNI binaries alongside the calico binaries.

Some other minor changes were:
- adding the pod-cidr to the proxy pods to get around some issues with Vagrant networking
- We use docker to run hyperkube with kubectl to deploy the calico.yaml manifest
- Removed all systemd Calico configuration
- Using one CNI conf for all nodes
- Vagrant files needed some touching as the VMs needed a bit more CPU/memory

We will want to update the remaining docs to use the same changes, but I wanted this reviewed before going through that process.

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -62,7 +62,7 @@ Next create a [systemd drop-in][dropins], which is a method for appending or ove
 ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
 ```
 
-[calico-docs]: https://github.com/projectcalico/calico-containers/tree/v0.19.0/docs/cni/kubernetes
+[calico-docs]: http://docs.projectcalico.org/v2.0/getting-started/kubernetes/
 [flannel-docs]: https://coreos.com/flannel/docs/latest/
 [pod-overview]: https://coreos.com/kubernetes/docs/latest/pods.html
 [service-overview]: https://coreos.com/kubernetes/docs/latest/services.html
@@ -124,7 +124,14 @@ Note that the kubelet running on a master node may log repeated attempts to post
 * Replace `${ADVERTISE_IP}` with this node's publicly routable IP.
 * Replace `${DNS_SERVICE_IP}`
 * Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release, e.g. `v1.4.6_coreos.0`.
-* Replace `${NETWORK_PLUGIN}` with `cni` if using Calico. Otherwise just leave it blank.
+* If using Calico for network policy
+  - Replace `${NETWORK_PLUGIN}` with `cni`
+  - Add the following to `RKT_OPS=`
+    ```
+    --volume cni-bin,kind=host,source=/opt/cni/bin \
+    --mount volume=cni-bin,target=/opt/cni/bin
+    ```
+  - Add `ExecStartPre=/usr/bin/mkdir -p /opt/cni/bin`
 * Decide if you will use [additional features][rkt-opts-examples] such as:
   - [mounting ephemeral disks][mount-disks]
   - [allow pods to mount RDB][rdb] or [iSCSI volumes][iscsi]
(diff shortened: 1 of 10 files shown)


single-node/user-data
```diff
@@ -149,7 +159,7 @@ EOF
         cat << EOF > $TEMPLATE
 [Unit]
 Requires=network-online.target
-After=network-online.target
+After=network-online.targetETCDENDPOINTS
```
caseydavenport: typo?

single-node/calico.yaml
```diff
@@ -0,0 +1,185 @@
+# This ConfigMap is used to configure a self-hosted Calico installation.
```
heschlie: I was thinking the same, but my concern is does it matter that the file would be in the multi-node install?  I'm thinking no?

single-node/user-data
heschlie: Indeed...

single-node/calico.yaml
heschlie: Pointed the single node to the same file and removed the extra, spun up the single node to make sure it went ok, looks good.

single-node/user-data
```diff
@@ -36,6 +36,14 @@ export USE_CALICO=false
 # Determines the container runtime for kubernetes to use. Accepts 'docker' or 'rkt'.
 export CONTAINER_RUNTIME=docker
 
+# We need to overwrite this for a hosted Calico install
+if [ "${USE_CALICO}" = "true" ]; then
+    export CALICO_OPTS="--volume cni-bin,kind=host,source=/opt/cni/bin \
+                        --mount volume=cni-bin,target=/opt/cni/bin"
```
philips: kube-flannel uses /etc/cni, I think we should do the same thing here https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml#L67

philips: ack, right.

multi-node/vagrant/Vagrantfile
```diff
@@ -91,13 +92,13 @@ Vagrant.configure("2") do |config|
...
     config.vm.provider vmware do |v|
-      v.vmx['numvcpus'] = 1
+      v.vmx['numvcpus'] = 2
       v.gui = false
     end
   end
 
   config.vm.provider :virtualbox do |vb|
-    vb.cpus = 1
+    vb.cpus = 2
```
aaronlevy: Is this necessary / safe assumption?

multi-node/vagrant/Vagrantfile
```diff
@@ -161,6 +162,9 @@ Vagrant.configure("2") do |config|
 
       controller.vm.provision :file, :source => CONTROLLER_CLOUD_CONFIG_PATH, :destination => "/tmp/vagrantfile-user-data"
       controller.vm.provision :shell, :inline => "mv /tmp/vagrantfile-user-data /var/lib/coreos-vagrant/", :privileged => true
+
+      controller.vm.provision :file, :source => CALICO_MANIFEST, :destination => "/tmp/calico.yaml"
+      controller.vm.provision :shell, :inline => "mkdir -p /srv/kubernetes/manifests && mv /tmp/calico.yaml /srv/kubernetes/manifests/", :privileged => true
```
aaronlevy: Ideally we're not relying on vagrant for anything beyond machine provisioning (this isn't completely true even now, but hopefully not adding more external configuration). What do you think about in-lining the calico manifest in the script (gated on USE_CALICO`)?

multi-node/generic/controller-install.sh
```diff
@@ -865,15 +771,20 @@ function start_addons {
...
+    until curl --silent "http://127.0.0.1:8080/version/"
     do
         sleep 5
     done
     echo
-    echo "K8S: Calico Policy"
-    curl --silent -H "Content-Type: application/json" -XPOST -d"$(cat /srv/kubernetes/manifests/calico-system.json)" "http://127.0.0.1:8080/api/v1/namespaces/" > /dev/null
+
+    # Deploy Calico
+    docker run --rm --net=host -v /srv/kubernetes/manifests:/host/manifests $HYPERKUBE_IMAGE_REPO:$K8S_VER /hyperkube kubectl apply -f /host/manifests/calico.yaml
```
aaronlevy: This looks like it will get run on every boot - the kubectl create should fail "already exists", which is fine - but if the exit code propagates to the docker run, then this could cause the script to exit (I'm just not sure about the actual behavior -- ignore if this has been tested / is safe).

aaronlevy:
Couple minor comments/questions

multi-node/vagrant/Vagrantfile
heschlie: I had run into some stability issues with one, but that may have been while I was trying to run the policy controller on the master, I'll retest and and revert back if possible.

multi-node/vagrant/Vagrantfile
heschlie: I was initially trying to avoid putting it into the script, but if that would be preferred I don't see a problem with that.

multi-node/generic/controller-install.sh
heschlie: The appears to fail on reboot long before it gets to this point.  Specifically it fails here:

https://github.com/coreos/coreos-kubernetes/blob/6bb3bb5b2267c716da8790e885b78d5e3f5c2e42/multi-node/generic/controller-install.sh#L53

As ETCD_ENDPOINTS seems to be unset on reboot.

heschlie: But on further review it should not report a failure if it did get to this point

heschlie:
I believe I hit all the necessary places.  The kube-aws docs aren't pulled from here anymore correct?

multi-node/generic/controller-install.sh
aaronlevy: Opened https://github.com/coreos/coreos-kubernetes/issues/775 to track this

single-node/Vagrantfile
```diff
@@ -12,6 +12,7 @@ $update_channel = "alpha"
 
 CLUSTER_IP="10.3.0.1"
 NODE_IP = "172.17.4.99"
+NODE_VCPUS = 2
```
aaronlevy: Did this turn out to be a requirement? It's probably generally fine to do (wouldn't work on single core machine -- not sure how common that would be). If we are keeping this, can we make this configurable in the multi-node installation too.

single-node/user-data
```diff
@@ -856,15 +951,14 @@ function start_addons {
...
+function enable_calico {
     echo "Waiting for Kubernetes API..."
     until curl --silent "http://127.0.0.1:8080/version"
     do
         sleep 5
     done
     echo
-    echo "K8S: Calico Policy"
-    curl --silent -H "Content-Type: application/json" -XPOST -d"$(cat /srv/kubernetes/manifests/calico-system.json)" "http://127.0.0.1:8080/api/v1/namespaces/" > /dev/null
+    docker run --rm --net=host -v /srv/kubernetes/manifests:/host/manifests $HYPERKUBE_IMAGE_REPO:$K8S_VER /hyperkube kubectl apply -f /host/manifests/calico.yaml
```
aaronlevy: If possible, can we switch this to rkt? We require rkt already due to running the kubelet via kubelet-wrapper, but if you want to use rkt as your runtime, we do not require the use of docker.

It should be mostly equivalent, but to implement `--rm` takes a little more coordination. 

When executing `rkt run`, add :
`--uuid-file-save=/var/run/calico-install.uuid`

Then after execution
`/usr/bin/rkt rm --uuid-file=/var/run/calico-install.uuid`

aaronlevy:
Couple more minor comments.

/cc @robszumski and @joshix for docs changes

single-node/user-data
heschlie: I'm getting `device or resource busy` when running rm against that pod, the full set of commands:

```sh
/usr/bin/rkt run --uuid-file-save=/var/run/calico-install.uuid \
                        --net=host \
                        --volume manifests,kind=host,source=/etc/kubernetes/manifests \
                        --mount volume=manifests,target=/host/manifests \
                        $HYPERKUBE_IMAGE_REPO:$K8S_VER --exec=/hyperkube -- kubectl apply -f /host/manifests/calico.yaml
/usr/bin/rkt rm --uuid-file=/var/run/calico-install.uuid
```

Any suggestions?  Should I just leave the pod in an exited state?

euank: The one other trick that might be worth trying is `rkt status --wait $uuid` between the `run` and `rm`

If a `wait` doesn't help, I'd be happy to dig a bit deeper

aaronlevy:
Based on @euank feedback (https://github.com/coreos/coreos-kubernetes/pull/768#discussion_r91819866) -- sorry to ask this, but prob can just switch back to using docker for now and add a TODO for switching to rkt once the issue is resolved

heschlie:
No problem, thanks for the fast feedback :)

aaronlevy:
@heschlie can you squash the cleanup commits / where appropriate and we can get this merged?

Documentation/deploy-master.md
```diff
@@ -98,7 +98,14 @@ Note that the kubelet running on a master node may log repeated attempts to post
...
 * Replace `${K8S_VER}` This will map to: `quay.io/coreos/hyperkube:${K8S_VER}` release, e.g. `v1.4.6_coreos.0`.
-* Replace `${NETWORK_PLUGIN}` with `cni` if using Calico. Otherwise just leave it blank.
+* If using Calico for network policy
+  - Replace `${NETWORK_PLUGIN}` with `cni`
+  - Add the following to `RKT_OPS=`
+    ```
+    --volume cni-bin,kind=host,source=/opt/cni/bin \
+    --mount volume=cni-bin,target=/opt/cni/bin
+    ```
+  - Add `ExecStartPre=/usr/bin/mkdir -p /opt/cni/bin`
```
robszumski: I wish there was a cleaner way to describe all of this, but I'm at a loss

Documentation/deploy-master.md
```diff
@@ -321,122 +328,213 @@ spec:
...
-The Calico node container runs on all hosts, including the master node. It performs two functions:
+Several things happen here.  First the `ConfigMap` will create some configuration needed for Calico and the kubelet.
+
+Second the `DaemonSet` runs on all hosts, including the master node. It performs several functions:
+* The `install-cni` container drops in the necessary CNI binaries to `/opt/cni/bin/` that was setup in the kubelet step
 * Connects containers to the flannel overlay network, which enables the "one IP per pod" concept.
 * Enforces network policy created through the Kubernetes policy API, ensuring pods talk to authorized resources only.
 
-When creating `/etc/systemd/system/calico-node.service`:
+Finally policy agent is the last major piece of the calico.yaml. It monitors the API for changes related to network policy and configures Calico to implement that policy.
```
robszumski: I think you can drop "finally" here.

> The policy agent is the last major piece of the Calico system.

caseydavenport: It should also read "The policy _controller_"

heschlie:
commits squashed!

aaronlevy:
@heschlie looks like there are some conflicts now. Can you rebase your changes

heschlie:
@aaronlevy and done!

aaronlevy:
Thanks. One last minor comment to address - then squash and we should be good.

single-node/user-data
```diff
@@ -31,11 +31,19 @@ export K8S_SERVICE_IP=10.3.0.1
 export DNS_SERVICE_IP=10.3.0.10
 
 # Whether to use Calico for Kubernetes network policy.
-export USE_CALICO=false
+export USE_CALICO=true
```
heschlie: yep sorry about that it should have been set to false, must have missed it, amended the last commit.

aaronlevy:
Perfect. Can you squash that last commit

PR #783, opened 2016-12-14
fix cn bug when generating tls assets using OpenSSL

dixudx:
These hostnames should be explicitly shown to avoid potential misunderstandings.


zbwright:

PR #787, opened 2016-12-17
allow creating devices to be mounted by pods

travisn:
Enable creation of storage devices in the vagrant environment so storage can be mounted by pods running in the cluster. By default no storage devices will be created. To create devices, the value of disks_per_node must be set to non-zero. For example, this will create two per node. By default the size of each is 5GB.
$disks_per_node=2
$disk_size = 5

### DIFF
--- a/lib/.gitignore
+++ b/lib/.gitignore
@@ -1,0 +1,1 @@
+.virtualbox/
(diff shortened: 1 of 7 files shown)


zbwright:

PR #789, opened 2016-12-20
Update deploy-master.md

jpap:
Added missing line continuation backslash in `/etc/systemd/system/kubelet.service` file listing.


robszumski:
Thanks for catching this!

PR #791, opened 2016-12-20
reference to master: https missing

stephanlindauer:


stephanlindauer:
because: 
```
MASTER_HOST=no default

The address of the master node. In most cases this will be the publicly routable IP of the node.
```
on `https://coreos.com/kubernetes/docs/latest/getting-started.html`

PR #792, opened 2016-12-21
caveat

stephanlindauer:


Documentation/deploy-workers.md
```diff
@@ -250,6 +250,7 @@ Start the kubelet, which will start the proxy.
 $ sudo systemctl start flanneld
 $ sudo systemctl start kubelet
 ```
+*Note:* It's vital, that at this point the API server is already up and running, which might not always be the case if Services are started via automated provisioning.
```
robszumski: Is there a way we can say this without the "note:" syntax?

What information can a user use to know if they are hitting this? What failures did you see that caused this note to be needed?

stephanlindauer:
after this has been idle for a while, i assume, this pr won't get in. 
closing the issue.

PR #799, opened 2017-01-05
Update Calico TAGS to v2 release and fix documentation

ArchiFleKs:
In the docs the kubectl apply part to deploy calico is missing.
Also in the doc /etc/kubernetes is used for calico manifest but
/srv/kubernetes is used in script. The doc is fixed to match the
script. Also, Calico tags are updated to v2 release
(http://docs.projectcalico.org/v2.0/releases/)


caseydavenport:
The Calico changes here look good to me.  @ArchiFleKs would you mind making the same version bumps to the equivalent single-node manifest?

https://github.com/coreos/coreos-kubernetes/blob/master/single-node/user-data#L852

Documentation/deploy-master.md
```diff
@@ -668,6 +668,14 @@ kube-controller-$node
 kube-proxy-$node
 ```
 
+### Configure Calico
+
+To start Calico services :
+
+```
+docker run --rm --net=host -v /srv/kubernetes/manifests:/host/manifests quay.io/coreos/hyperkube:v1.5.1_coreos.0 /hyperkube kubectl apply -f /host/manifests/calico.yaml
```
robszumski: Using a container stands out to me, I don't think this has been done anywhere else in all of the CoreOS docs. Is there another way?

caseydavenport: Hmm, other options I can think of would be:

- Running kubectl _not_ in a container by downloading the binary.
- Running the Kubernetes add-on manager.

robszumski: > Running kubectl not in a container by downloading the binary.

This would mean either duplicating the kubectl set up instructions, or moving them up from the 3rd step. Does it make sense to add this Calico part onto this page: https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-addons.md

It's kind of like an addon, so I can see that working from a "deploy cluster then addons" mental flow.

ArchiFleKs: @robszumski It is done i the script for multinode

ArchiFleKs: @caseydavenport : I'm using the addons manager on CoreOS and it is working fine, I can propose a PR for that

ArchiFleKs:
I updated the docs and generic multi and single node script to use kube-addon-manager v6.1.

I also added heapster and updated addons manifest in documentation for the "Deploy addons section"

Documentation/deploy-master.md
caseydavenport: I'm actually a little bit hesitant to go with either of the above approaches.

It's really important that if a user wants to install Calico, that Calico is installed before other addons.  

For example, if the DNS add-on is installed first, then it won't get networked with Calico and thus won't have Calico policy applied to it.  Ideally it is done as part of installing the master, or immediately after installing the nodes and before installing any other add-on.

In the "Calico only" case, the add-on manager would be fine since no other pods could start until Calico was installed, but since users are also installing Flannel, it will now be possible for pods to start prior to Calico getting installed, which would be a problem.

robszumski: Hmm ok, maybe the container method is the cleanest. The only other thing I can think of would be to use curl or downloading the binary.

ArchiFleKs: @caseydavenport Is starting order really important ? If Calico restart or is upgraded after addon deployment, network policies will still apply no ?

caseydavenport: @ArchiFleKs the problem is that if a pod is started using the flannel without the Calico CNI plugin, then Calico won't know about the pod and thus won't be able to enforce policy on it, even if you restart Calico.

You'd have to restart the pod itself in order for it to be recreated with the Calico CNI plugin in order to make it work.

ArchiFleKs:
I'd thought they're would be some kind of convergence/magic, since calico
is looking into etcd ^^.

Maybe we can start kubelet with calico and addon manager, wait for calico
and then drop the addons template?

Le lun. 9 janv. 2017 à 04:39, Casey Davenport <notifications@github.com> a
écrit :

ArchiFleKs:
@caseydavenport any update on this ?

ArchiFleKs:
@robszumski @caseydavenport : Hi Guys, any update on this ? :)

caseydavenport:
@ArchiFleKs Sorry for the delay on this, I've been totally swamped!

>  I'd thought they're would be some kind of convergence/magic, since calico
is looking into etcd

Calico just looks at it's own part of etcd (/calico) so it's not aware of anything that flannel might have written, so unfortunately no magic here.

I'm still leaning towards just using a `kubectl apply`to install Calico in the step-by-step documentation, provided @robszumski is OK with that.  We could either do that here, or in the deploy-addons stage, but my preference would be here given the ordering concerns with other add-ons.

ArchiFleKs:
@caseydavenport : maybe we could write the calico.yaml template with addons manager and wait for it to be ready (maybe with docker and `kubectl`) before writing the other addons ? 

This way we get to keep addons-manager and calico without manual cURLing ?

robszumski:
@caseydavenport Yeah, we might as well move forward with that

caseydavenport:
>  maybe we could write the calico.yaml template with addons manager and wait for it to be ready (maybe with docker and kubectl) before writing the other addons ?

My concern with that is that in the future people can unknowingly add additional add-ons via the add-on manager.  AFAIK the add-on manager doesn't have a concept of "wait for this to be running before deploying any other add-ons", so it's totally possible that other add-ons would be deployed before Calico.

ArchiFleKs:
@caseydavenport @robszumski Yes but if calico is the first addon we write, we can may be check the state of the calico pods before writing other addons with the script. Therefore we're sure that calico is the first addons to be deploy

ArchiFleKs:
@caseydavenport @robszumski I'm going to drop this, it seems to already been updated in docs

PR #804, opened 2017-01-09
change calico.yaml path from /etc to /srv

nimbusscale:
single-node/user-data: The calico.yaml needs to be created at /srv/kubernetes/manifests/calico.yaml instead of /etc/kubernetes/manifests/calico.yaml.

This is needed for the command `docker run --rm --net=host -v /srv/kubernetes/manifests:/host/manifests $HYPERKUBE_IMAGE_REPO:$K8S_VER /hyperkube kubectl apply -f /host/manifests/calico.yaml` to work. Without this change the above command fails and calico is not setup properly, and networking doesn't work at all.

### DIFF
--- a/single-node/user-data
+++ b/single-node/user-data
@@ -808,7 +808,7 @@ EOF
 }
 EOF
     fi
-    local TEMPLATE=/etc/kubernetes/manifests/calico.yaml
+    local TEMPLATE=/srv/kubernetes/manifests/calico.yaml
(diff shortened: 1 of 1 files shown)


heschlie:
Good catch, LGTM, @aaronlevy this is a pretty small change, could we get this merged if it looks good to you?

aaronlevy:
lgtm

PR #815, opened 2017-01-23
Use var instead of predefined value.

fbarbeira:
In the docs are specified to use the var ${NETWORK_PLUGIN} if you use calico. I guess that if you use flanneld you must leave that field blank instead of use the default value "cni".

"* If using Calico for network policy
  - Replace `${NETWORK_PLUGIN}` with `cni`"


aaronlevy:
This should be closed by: https://github.com/coreos/coreos-kubernetes/pull/813

PR #816, opened 2017-01-24
missing ports for master

stephanlindauer:
those ports have to be open on the master as well. otherwise things like kube-proxy+dashboard wont work.

### DIFF
--- a/Documentation/kubernetes-networking.md
+++ b/Documentation/kubernetes-networking.md
@@ -26,6 +26,8 @@ Master Node Inbound
 | Protocol | Port Range | Source                                    | Purpose                |
 -----------|------------|-------------------------------------------|------------------------|
 | TCP      | 443        | Worker Nodes, API Requests, and End-Users | Kubernetes API server. |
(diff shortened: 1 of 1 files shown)


aaronlevy:
@stephanlindauer we should probably document allowing traffic between both masters and workers on the pod network. So maybe updating the ingress rules from both master/worker to state source could be from master & worker?

stephanlindauer:
not 100% sure what you mean. the ports are already mentioned in the `Worker Node Inbound` section. or do you want to add a new table for inboud rules that should be applied to both master and worker nodes?

aaronlevy:
Sorry for the vagueness. We likely also need to allow master->worker pod traffic (and master to master pod traffic). Looks like we're only allowing pod traffic to be sourced from workers.

We mention opening all TCP traffic between worker/master, but not in the case for these UDP ports.

So I think it would just be a change of `source=master & worker` for the flannel rules in both the ingress tables.

stephanlindauer:
ok understood. makes sense. ;) 
i changed the pr accordingly.

aaronlevy:
LGTM, thanks!

aaronlevy:
Oh actually, would you mind squashing into a single commit?

stephanlindauer:
yup. done.

PR #820, opened 2017-01-31
Rename coreos-baremetal to matchbox

jstoja:
Since the evrsion v0.5.0, the coreos-baremetal project has ben renamed to matchbox.


jstoja:
Sorry it was a duplicate.

PR #822, opened 2017-02-01
CoreOS automatically imports flannel config into docker, so it's not …

cjyar:
…necessary to do that manually.


zbwright:
Thanks for your work. This file has recently been rewritten. Closing this PR.

PR #823, opened 2017-02-01
Set --apiserver-count if more than one master node.

cjyar:


aaronlevy:
The `--apiserver-count` flag behavior is a little less than desirable (see: kubernetes/kubernetes#22609)

Essentially you're putting a fixed number of endpoints into the kubernetes service, and if those endpoints happen to be down, a certain percentage of requests just fail (because the endpoints are not cleaned up).

Multiple api servers will continue to work without this flag - its just that they will fight over the single endpoint entry in the kubernetes service (constantly updating it). There will eventually be better upstream support for this, but in the interim the best option is to front your apiservers with a loadbalancer, or DNS entry which points to the correct endpoints. Then setting your apiserver `--advertise-address` as the LB.

PR #828, opened 2017-02-08
Remove link from kubelet.service to "CoreOS auto-update settings"

urzds:
This seems completely unrelated and the site being linked to does not contain a single bit about Kubernetes...


aaronlevy:
I believe this is still an important part of operating a CoreOS cluster (including a kubernetes cluster running on CoreOS), and doesn't hurt to have the one line mentioned.

PR #837, opened 2017-02-16
Retain IP address assignments on reboot or kubelet restarts

emptywee:
This will prevent from assigning the same IP addresses to new pods after node reboot or kubelet restart.

See https://github.com/coreos/kube-aws/pull/64/files and https://github.com/coreos/kube-aws/issues/60 for more details.


aaronlevy:
Good catch @emptywee .

Would you mind squashing into a single commit?

PR #840, opened 2017-02-22
"hostNetwork: true mentioned twice"

arehmandev:
"hostNetwork: true" mentioned twice in Kube controller manager static pod manifest. (duplicate key)


aaronlevy:
LGTM @arehmandev thanks!

PR #842, opened 2017-02-27
kubelet: document that 10250 shouldn't be exposed

### LINKED ISSUE #656: Document the fact that you absolutely should not expose port 10250 on any of your nodes to the internet.
This can live well in `Documentation/kubernetes-networking` but also should be very clearly called out.

The upstream issue is https://github.com/kubernetes/kubernetes/issues/11816

The possible solutions are:
1. nodes not exposed to public internet except api-server (e.g. AWS VPC) and you accept that all traffic in your vpc is trusted
2. ssh tunnels + localhost (http://kubernetes.io/docs/admin/master-node-communication/#master---cluster)
3. iptables based firewalling on every node

```
for iptablecmd in iptables ip6tables; do
  $iptablecmd -I INPUT -p tcp -s $masterIP --dport 10250 -j ACCEPT
  $iptablecmd -I INPUT -p tcp --dport 10250 -j DROP
done
```

euank:
Fixes #656

I'll file a followup to document the details of actually enabling kubelet authn/authz... but knowing is half the battle so we should have the note there until we have better docs.

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -138,6 +138,8 @@ Note that the kubelet running on a master node may log repeated attempts to post
   - [allowing access to insecure container registries][insecure-registry]
   - [changing your CoreOS auto-update settings][update]
 
+**Note**: Anyone with access to port 10250 on a node can execute arbitrary code in a pod on the node. Information, including logs and metadata, is also disclosed on port 10255. See [securing the Kubelet API][securing-kubelet-api] for more information.
+
 **/etc/systemd/system/kubelet.service**
 
 ```yaml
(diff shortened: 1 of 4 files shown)


Documentation/deploy-master.md
```diff
@@ -138,6 +138,8 @@ Note that the kubelet running on a master node may log repeated attempts to post
   - [allowing access to insecure container registries][insecure-registry]
   - [changing your CoreOS auto-update settings][update]
 
+**Note**: Anyone with access to port 10250 on a node can execute arbitrary code in a pod on the node. Information, including logs and metdata, is also disclosed on port 10255. See [securing the Kubelet API][securing-kubelet-api] for more information.
```
crawford: Needs a reference link.

Documentation/kubelet-wrapper.md
```diff
@@ -87,6 +87,17 @@ Environment="RKT_OPTS=--volume modprobe,kind=host,source=/usr/sbin/modprobe \
 
 Note that the kubelet also requires access to the userspace `rbd` tool that is included only in hyperkube images tagged `v1.3.6_coreos.0` or later.
 
+### Securing the Kubelet API
+
+By default, the Kubelet allows unathenticated access to its [API ports][kubernetes-ports].
```
crawford: unauthenticated

crawford:
Can you run marker over this while you are at it?

euank:
@crawford Addressed your comments, thanks!

Marker spotted other stuff which I'll get in a new PR

PR #847, opened 2017-03-08
Retain IP addresses of pods on node reboot or kubelet restart

emptywee:
This will prevent from assigning the same IP addresses to new pods after node reboot or kubelet restart.

See https://github.com/coreos/kube-aws/pull/64/files and https://github.com/coreos/kube-aws/issues/60 for more details.

### DIFF
--- a/Documentation/deploy-master.md
+++ b/Documentation/deploy-master.md
@@ -128,6 +128,8 @@ Note that the kubelet running on a master node may log repeated attempts to post
   - Replace `${NETWORK_PLUGIN}` with `cni`
   - Add the following to `RKT_OPS=`
     ```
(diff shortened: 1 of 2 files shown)


aaronlevy:
Looks like it might be missing the `ExecStartPre=/usr/bin/mkdir -p /var/lib/cni` from https://github.com/coreos/kube-aws/pull/64/files. Likely necessary because if you're using CNI plugins that directory may not exist when you first launch the kubelet.

emptywee:
It is missing. I was asked to squash it into single commit, when I did it, I lost that part. Anyway, I had to learn it the hard way, decided to share and possibly add it to the docs. I don't mind if someone just adds this to the docs.

zbwright:

PR #865, opened 2017-04-14
Docs fix cluster_{dns,domain} to cluster-{dns,domain}

5tefan:
Per the kubelet documentation [1], the options are

--cluster-dns stringSlice      Comma-separated list of DNS server IP address.
 This value is used for containers DNS server in case of Pods with
 "dnsPolicy=ClusterFirst". Note: all DNS servers appearing in the list MUST
 serve the same set of records otherwise name resolution within the cluster may
 not work correctly. There is no guarantee as to which DNS server may be
 contacted for name resolution.

--cluster-domain string        Domain for this cluster.  If set, kubelet will
 configure all containers to search this domain in addition to the host's search
 domains

[1] https://kubernetes.io/docs/admin/kubelet/


5tefan:
I'm somewhat doubting myself on this since it seems to be everywhere and I figured someone would have caught it if it wasn't correct....

There are a couple more occurrences to fix in case this is actually a misconfiguration.

aknuds1:
This looks right to me, since your options correspond to [the official documentation](https://kubernetes.io/docs/admin/kubelet/)!

zbwright:
Thanks for your update. These pages have recently been rewritten. Closing this PR.

zbwright:
Thanks for your update. These pages have recently been rewritten. Closing this PR.

PR #880, opened 2017-05-22
Fixing for Windows environment. Bash shell env is different. 

mikesmullin:
The commands need tweaking and escaping.

Eventually got it working for myself.

I also added some stuff to generate a self-signed CA and issuer that gets added to `/etc/ssl/certs` for use by HTTPS services like the Docker Registry container which is one of the first ones I wanted to run, and which I wanted the kubelet workers etc to use to download images from. That part isn't included in this PR but probably should be a feature, if others agree.

### DIFF
--- a/lib/init-ssl-ca
+++ b/lib/init-ssl-ca
@@ -30,6 +30,6 @@ fi
 
 # establish cluster CA and self-sign a cert
 openssl genrsa -out "$OUTDIR/ca-key.pem" 2048
-openssl req -x509 -new -nodes -key "$OUTDIR/ca-key.pem" -days 10000 -out "$OUTFILE" -subj "/CN=kube-ca"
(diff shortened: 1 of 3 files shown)


lib/init-ssl-ca
```diff
@@ -30,6 +30,6 @@ fi
 
 # establish cluster CA and self-sign a cert
 openssl genrsa -out "$OUTDIR/ca-key.pem" 2048
-openssl req -x509 -new -nodes -key "$OUTDIR/ca-key.pem" -days 10000 -out "$OUTFILE" -subj "/CN=kube-ca"
+openssl req -x509 -new -nodes -key "$OUTDIR/ca-key.pem" -days 10000 -out "$OUTFILE" -subj "//CN=kube-ca"
```
mikesmullin: Not sure what to do about these `//` they are double-escaping but it needs to be conditionally added by OS

zbwright:

PR #890, opened 2017-06-20
Fix: Pods' IP is now in POD_NETWORK range

### LINKED ISSUE #872: [Docker] Pods' IP not in $POD_NETWORK range
When deploying a cluster, everything works well but when I run a few docker's containers on it, pods do not have the appropriate network configuration. 

After some troubleshooting, it seems that the range set in the Docker's network is not well configured. The Flannel interface was different from the Docker one. 

Finally, the docker.service do not have the **--bip** option set. It seems that this file (on workers) might be the problem: **/etc/systemd/system/docker.service.d/40-flannel.conf**. I changed the systemd drop-in in my worker from: **EnvironmentFile=/etc/kubernetes/cni/docker_opts_cni.env** to: **EnvironmentFile=/run/flannel/flannel_docker_opts.env**. 

After reloading, my pods get the configured sub network specified in etcd.  

Am I doing something wrong or is it a problem in the documentation? 

I can run a few more tests if required.

### LINKED ISSUE #884: Docker not using the flannel IPs
In the [documentation](https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/deploy-master.md#docker-configuration) it says to create a drop-in containing the following:

    [Service]
    EnvironmentFile=/etc/kubernetes/cni/docker_opts_cni.env

And a `/etc/kubernetes/cni/docker_opts_cni.env` file containing the following:

    DOCKER_OPT_BIP=""
    DOCKER_OPT_IPMASQ=""

But it seems that these environment variables only seems nullify the environment variables in `/var/run/flannel/flannel_docker_opts.env` and cause docker to use its default IPs (ie. pods come up with a `172.17.0.x` address. Commenting out the `EnvironmentFile` line in the drop-in and restarting docker causes the pods to then pick up a `10.2.x.x` IP.

Is the documentation incorrect? or am I missing something?

JWimsingues:
Change the drop-in 40-flannel.conf. Pods are now in range of the env var POD_NETWORK.

Fix #872 
Fix #884

### DIFF
--- a/Documentation/deploy-workers.md
+++ b/Documentation/deploy-workers.md
@@ -72,7 +72,7 @@ Create `/etc/systemd/system/docker.service.d/40-flannel.conf`
 Requires=flanneld.service
 After=flanneld.service
 [Service]
-EnvironmentFile=/etc/kubernetes/cni/docker_opts_cni.env
+EnvironmentFile=/run/flannel/flannel_docker_opts.env
 ```
 
 Create the Docker CNI Options file:


zbwright:
Thanks for your work. This page has recently been rewritten. Closing this PR.

PR #893, opened 2017-06-27
Fixed a bug of apiserver's service account key file.

laohyx:
According to https://kubernetes.io/docs/admin/service-accounts-admin/, the apiserver accepts `service-account-key-file` with public key, not private key file.


zbwright:

PR #905, opened 2017-09-07
[WIP] docs: adding watch recovery page.

zbwright:


zbwright:
I cannot add reviewers in this repo - passing to Thai.

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,81 @@
...
+
+Open a terminal and a browser window with Tectonic Console. Keep them open side by side for this exercise. Kubernetes will rebuild the killed containers quickly.
+
+First, use Tectonic Console to determine which nodes are running the API server.
+Go to *Workloads > Pods*, and enter ‘api’ in the search field.
+
+Then, select one of the nodes listed, and SSH to its IP address.
+
+```
+Sssh core@ip
```
athai: Remove the extra 's' from the command

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+Remember that the kubelet runs on **reconciliation loops** which are constantly work to keep the system in a stable state. When we execute the command to kill this API server, the kubelet will notice immediately.
```
athai: The grammar in the first sentence is weird.

Should it be:
* "...runs on **reconciliation loops** which are constantly work*ing...*" (work --> working) or 
* "...run on **reconciliation loop** which ~~are~~ constantly work" (remove "are" from the sentence)?

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,81 @@
...
+
+Within a few seconds the kubelet will notice that the API server Pod is no longer running, and will relaunch under a new ID:
+
+```
+$ docker ps -f 'name=.*kube-apiserver.*' --format "{{.ID}}"
+f4a262619d1d
+d964fa94c69b
+```
+
+In the console, watch the list  of API servers to see that one is disabled, and a new one is launched. Note that the Tectonic Console session remained working because the API is configured with High Availability. This is the same Kubernetes features that enable your apps to remain up and running without constant developer monitoring and interaction.
```
athai: * There's an extra space after "list" in the first sentence
* Second sentence is somewhat awkwardly worded. I think "Note that the Tectonic Console session continues to work because the API..." sounds a little bit better?
* Third sentence: features --> feature, enable --> enables

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,81 @@
...
+$ systemctl stop dockerk
+$ systemctl stop kubelet
+$ ps aux | grep kube
+```
+
+Use Tectonic Console to inspect the node. Notice that the node is marked ‘unhealthy’, and nothing is rescheduled.
+
+Check your app. Even though the node is unhealthy, and the rescheduling process has not yet begun, the app is still up and running as before. Because the cluster runs in High Availability, the app does not go down, and your clients experience no interruption in service. Kubernetes maintains persistent availability automatically.
+
+Wait 5 mins, then reinspect Console. Notice that new nodes have been scheduled to repair the interruption.
```
athai: mins --> minutes

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,81 @@
...
+
+Tectonic configures the kubelet to wait 5 minutes after a node is determined to be unhealthy before workloads are moved off the node. Because nodes are run in High Availability, redundancy allows the system this time to regroup.
```
athai: "...redundancy allows the system this time to regroup." <-- Why say "this time"? Does this not consistently work?

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,81 @@
+# Break system locally and watch recovery
+
+First, make sure you are running a sample app. Follow the [Tectonic Sandbox tutorials][first-app] to launch a simple app, or use one of your own. Once an app is running, break the cluster and watch how it recovers.
+
+One of the core features of Kubernetes is that it’s designed to maintain the desired state defined by operators and app admins. Now that we have an app running, let’s see what happen when we intentionally kill a copy of the API server, which is the brains of the cluster. The API server is run as a Deployment with multiple Pods, to ensure High Availability. Each one of these runs on a master node.
```
athai: Kind of a nit, but it seems like a transition sentence is needed before the third sentence. Alternatively, I think restructuring the paragraph (so that the explanation of the API server and how it acts as the brains to the cluster comes first) may read better. E.g., something like:
> The API server acts as the brains of the cluster and is run as a Deployment with multiple Pods to ensure High Availability. Each one of these runs on a master node. Now that we have an app running, let's see what happens when we intentionally kill a copy of the API server.

I think if we end with "Let's do the thing", it more naturally leads into the next section.

But if you prefer it as is, that's cool too. In the second sentence, "happen" should be changed to "happens" though.

zbwright:
passing this to you, @robszumski

Documentation/watch-recovery.md
```diff
@@ -0,0 +1,104 @@
...
+
+```
+$ iptables -A OUTPUT -p tcp --dport 443  -j ACCEPT
+$ systemctl start dockerk
+$ systemctl enable kubelet
+$ ps aux | grep kube
+
+```
+
+## More testing opportunities
```
zbwright: @robszumski Without links to relevant doc, it doesn't make sense to add this section. I cannot find Tectonic doc to fulfill these three bullets.

could we go:
* performance and load testing (?)
* use Deployments to scale an app (https://coreos.com/tectonic/docs/latest/tutorials/sandbox/scale-app.html)
* review resource limits and namespace quotas (https://coreos.com/tectonic/docs/latest/tutorials/sandbox/check-logs.html ?)

zbwright:
also - the only pages we have to link to are Pods, Services, and Replication Controllers. until we build out others, doesn't make sense to link out from this one.

Documentation/watch-recovery.md
robszumski: Link to upstream? My thinking was that it's a hint that there is much more to do that just this guide. If the mindset of someone landing here is to learn, here are more things you can learn about.

https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
https://kubernetes.io/docs/tasks/administer-cluster/memory-default-namespace/
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

zbwright:
These pages have been moved to the tectonic-docs repo. Closing this PR.

PR #921, opened 2018-02-22
Update deploy-master.md

gavvvr:
Replace removed `--api-servers` with `--kubeconfig`


gavvvr:
Hmm. Just realised that `--kubeconfig` should point to a file. So my PR is wrong. 
I tried to install k8s 1.9.3 and faced an issue with `--api-servers`, it does not exists. So docs need a proper update anyways.

